PR #144, opened 2019-01-03
Inflation, Coinbase & Block Changes

AHBruns:


src/main/scala/bifrost/BifrostNodeViewHolder.scala
```diff
@@ -44,10 +44,11 @@ class BifrostNodeViewHolder(settings: ForgingSettings)
     */
   override def restoreState(): Option[NodeView] = {
     if (BWallet.exists(settings)) {
+      val x = BifrostHistory.readOrGenerate(settings)
       Some(
         (
-          BifrostHistory.readOrGenerate(settings),
-          BifrostState.readOrGenerate(settings),
+          x,
+          BifrostState.readOrGenerate(settings, true, x),
```
AHBruns: BifrostState now takes BifrostHistory as an arg to allow for CB TX's state validation method which asserts the inflation created by the CB TX is equal to the inflation value specified in the block the CB TX is included in.

src/main/scala/bifrost/BifrostNodeViewHolder.scala
```diff
@@ -123,12 +124,12 @@ object BifrostNodeViewHolder extends ScorexLogging {
     assert(Base58.encode(genesisTxs.head.id) == "5dJRukdd7sw7cmc8vwSnwbVggWLPV4VHYsZt7AQcFW3B", Base58.encode(genesisTxs.head.id))
 
     val genesisBox = ArbitBox(genesisAccountPriv.publicImage, 0, GenesisBalance)
-    val genesisBlock = BifrostBlock.create(settings.GenesisParentId, 0L, genesisTxs, genesisBox, genesisAccountPriv)
+    val genesisBlock = BifrostBlock.create(settings.GenesisParentId, 0L, genesisTxs, genesisBox, genesisAccountPriv, 10L) // arbitrary inflation for first block of 10 Arbits
```
AHBruns: Test updated to work with new BifrostBlock implementation. Has no effect on the outcome of the test as the block is never supplied with a CB TX.

src/main/scala/bifrost/BifrostNodeViewHolder.scala
```diff
@@ -123,12 +124,12 @@ object BifrostNodeViewHolder extends ScorexLogging {
...
 
     var history = BifrostHistory.readOrGenerate(settings)
     history = history.append(genesisBlock).get._1
 
-    val gs = BifrostState.genesisState(settings, Seq(genesisBlock))
+    val gs = BifrostState.genesisState(settings, Seq(genesisBlock), history)
```
AHBruns: BifrostState now take BifrostHistory as an arg.

src/main/scala/bifrost/blocks/BifrostBlock.scala
```diff
@@ -24,7 +24,8 @@ case class BifrostBlock(override val parentId: BlockId,
                         override val timestamp: Block.Timestamp,
                         forgerBox: ArbitBox,
                         signature: Signature25519,
-                        txs: Seq[BifrostTransaction])
+                        txs: Seq[BifrostTransaction],
+                        inflation: Long)
```
AHBruns: Test updated to work with new BifrostBlock implementation. Has no effect on the outcome of the test as the block is never supplied with a CB TX.

src/main/scala/bifrost/inflation/InflationQuery.scala
```diff
@@ -0,0 +1,52 @@
...
+        print("returning infUpdateVal : " + infUpdateVal + "\n")
+        sender() ! infUpdateVal
+      } else {
+        print("returning infVal : " + infVal + "\n")
+        sender() ! infVal
+      }
+      getUpdatedVals()
+  }
+
+}
```
AHBruns: This is the actor that queries the inflation server.

src/main/scala/bifrost/state/BifrostState.scala
```diff
@@ -688,6 +691,21 @@ case class BifrostState(storage: LSMStore, override val version: VersionTag, tim
...
+      assert(cb.newBoxes.size == 1) // one one new box
+      assert(cb.newBoxes.head.isInstanceOf[ArbitBox]) // the new box is an arbit box
+      Try {
+        assert(cb.newBoxes.head.asInstanceOf[ArbitBox].value
+          == history.modifierById(history.chainBack(history.bestBlock, helper).get.last._2).get.inflation) // inflation amount is correct
+      }
+    }
+    validConstruction
+  }
+
```
AHBruns: does some basic checks as well as checking that the value of the CB TX's arbit box is equal to the inflation value of the block it was included in.

src/main/scala/bifrost/state/BifrostState.scala
```diff
@@ -798,7 +817,7 @@ object BifrostState {
     }
   }
 
-  def readOrGenerate(settings: Settings, callFromGenesis: Boolean = false): BifrostState = {
+  def readOrGenerate(settings: ForgingSettings, callFromGenesis: Boolean = false, history: BifrostHistory): BifrostState = {
```
AHBruns: Just casting settings and adding BifrostHistory as an arg.

src/main/scala/bifrost/transaction/BifrostTransaction.scala
```diff
@@ -75,10 +75,88 @@ object BifrostTransaction {
   def nonceFromDigest(digest: Array[Byte]): Nonce = Longs.fromByteArray(digest.take(Longs.BYTES))
 }
 
+case class CoinbaseTransaction (to: IndexedSeq[(PublicKey25519Proposition, Long)],
+                                signatures: IndexedSeq[Signature25519],
+                                override val timestamp: Long,
+                                blockID: Array[Byte]) extends BifrostTransaction {
+  override type M = CoinbaseTransaction
```
AHBruns: moving classes around caused a conflict. The AssetCreation class didn't change.

src/main/scala/bifrost/transaction/BifrostTransactionCompanion.scala
```diff
@@ -772,6 +774,69 @@ object TokenExchangeTransactionCompanion extends Serializer[TokenExchangeTransac
...
+        bytesWithoutType.slice(numReadBytes + (i + 1) * elementLength - Longs.BYTES, numReadBytes + (i + 1) * elementLength)
+      )
+      (PublicKey25519Proposition(pk), v)
+    }
+
+
+    CoinbaseTransaction(to, signatures, timestamp, blockID)
+  }
+}
+
```
AHBruns: standard serialization stuff.

src/test/scala/bifrost/BifrostGenerators.scala
```diff
@@ -629,7 +629,7 @@ trait BifrostGenerators extends CoreGenerators {
     signature <- signatureGen
     txs <- bifrostTransactionSeqGen
   } yield {
-    BifrostBlock(parentId, timestamp, generatorBox, signature, txs)
+    BifrostBlock(parentId, timestamp, generatorBox, signature, txs, 10L)
```
AHBruns: updated tests to gen blocks with an inflation value. Doesn't change functionality.

src/test/scala/bifrost/BifrostGenerators.scala
```diff
@@ -660,7 +660,7 @@ trait BifrostGenerators extends CoreGenerators {
       1478164225796L,
       Seq(),
       ArbitBox(keyPair._2, 0L, 0L),
-      keyPair._1)
+      keyPair._1, 10L)  // genesis block has 10 Arbits of inflation
```
AHBruns: This is arbitrary.

src/test/scala/bifrost/ValidGenerators.scala
```diff
@@ -386,6 +387,24 @@ trait ValidGenerators extends BifrostGenerators {
...
+    val messageToSign = CoinbaseTransaction(
+      to,
+      fakeSigs,
+      timestamp,
+      id).messageToSign
+    // sign with own key because coinbase is literally giving yourself money
+    val signatures = IndexedSeq(PrivateKey25519Companion.sign(toKeyPairs._1, messageToSign))
+    CoinbaseTransaction(to, signatures, timestamp, id)
+  }
+
```
AHBruns: CB TX generator for testing. Can't be used for state validation checks since this doesn't add the fake blocks to a prepared state. Just for semantic validation testing.

src/test/scala/bifrost/transaction/CoinbaseTransactionSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+    preparedState.changes(parentBlock)
+    BifrostStateSpec.history.append(parentBlock)
+    val CB = CoinbaseTransaction.createAndApply(BifrostStateSpec.gw, IndexedSeq((test25519.publicImage, 666L)), parentBlock.id).get
+    val testBlock = BifrostBlock.create(parentBlock.id, 10001L, Seq(CB), forgerBox, test25519, 666)
+    preparedState.changes(testBlock)
+    BifrostStateSpec.history.append(testBlock)
+    preparedState.validate(CB).isSuccess shouldBe true
+
+  }
+}
```
AHBruns: CB tests gens random CB TXs and runs semantic validation on them. Then is generates a preparedState and simulates creating a block with CB TX in it. This is where the state validation method is checked. Future optimization would be to make this a generator to improve assurances.

src/main/scala/bifrost/BifrostNodeViewHolder.scala
scasplte2: why the extra variable x here?

I'd like to discuss how History is used (might be answered below)

src/main/scala/bifrost/BifrostNodeViewHolder.scala
scasplte2: Is it common to hardcode single numbers like this. You commented so I'm fine leavnig it (since we know what it is) but several places throughout Bifrost have hard coded 0's

src/main/scala/bifrost/forging/Forger.scala
```diff
@@ -32,6 +39,9 @@ class Forger(forgerSettings: ForgingSettings, viewHolderRef: ActorRef) extends A
   //should be a part of consensus, but for our app is okay
   val TransactionsInBlock = 100
 
+  // inflation query actor
+  private val infQ = ActorSystem("infChannel").actorOf(Props[InflationQuery], "infQ")
+
```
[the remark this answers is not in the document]
scasplte2: Is there usually more than one forger instance or is that instance started and stopped at any time?

src/main/scala/bifrost/forging/Forger.scala
```diff
@@ -41,18 +51,27 @@ class Forger(forgerSettings: ForgingSettings, viewHolderRef: ActorRef) extends A
     if (forging) context.system.scheduler.scheduleOnce(1.second)(self ! StartForging)
   }
 
-  def pickTransactions(memPool: BifrostMemPool, state: BifrostState): Seq[BifrostTransaction] =
-    memPool.take(TransactionsInBlock).foldLeft(Seq[BifrostTransaction]()) { case (txSoFar, tx) =>
+  def pickTransactions(memPool: BifrostMemPool, state: BifrostState, parent: BifrostBlock): Seq[BifrostTransaction] = {
+    implicit val timeout: Timeout = 10 seconds
+    val view = (viewHolderRef ? GetCurrentView).mapTo[CurrentView[BifrostHistory, BifrostState, BWallet, BifrostMemPool]]
+    infQ ? "getUpdatedVals"
+    val res = Await.result(view, Duration.Inf)
```
scasplte2: Could this be a problem? Shouldn't we avoid using awaits serially? I'll need to understand what the inflationQuery actor is doing to really see if its a problem though.

src/main/scala/bifrost/forging/Forger.scala
```diff
@@ -41,18 +51,27 @@ class Forger(forgerSettings: ForgingSettings, viewHolderRef: ActorRef) extends A
...
+    lazy val to: PublicKey25519Proposition = PublicKey25519Proposition(res.vault.secrets.head.publicImage.pubKeyBytes)
+    val infVal = Await.result(infQ ? res.history.height, Duration.Inf).asInstanceOf[Long]
+    print("infVal being used in forger: " + infVal + "\n")
```
scasplte2: is there a reason other places use log.debug or log.warn vs just print?

src/main/scala/bifrost/inflation/InflationQuery.scala
```diff
@@ -0,0 +1,52 @@
...
+  var infUpdateVal = 1L
+  var infUpdateBlock = 1L
+
+  @throws(classOf[java.io.IOException])
+  private def get(url: String)= scala.io.Source.fromURL(url).mkString
+
+  private def getUpdatedVals() = {
+    try {
+      //print("starting update call\n")
+      val resp = get("http://167.99.135.150/") // returns a string "infVal,###,infUpdateVal,###,infUpdateBlock,###"
```
scasplte2: Should we hard code the server address this deep?

src/main/scala/bifrost/inflation/InflationQuery.scala
```diff
@@ -0,0 +1,52 @@
...
+      //print("update call complete: infVal " + infVal + ", infUpdateVal " + infUpdateVal + ", infUpdateBlock " + infUpdateBlock + "\n")
+    } catch {
+      case e: Exception => print(e + "\n")
+    }
+  }
+
+  override def receive: Receive = {
+    case "getUpdatedVals" => // prompting val update
+      getUpdatedVals()
+    case x: Long =>
```
scasplte2: Can we reformat this to be like the other actors? (I'll explain what that means)

src/main/scala/bifrost/transaction/BifrostTransaction.scala
```diff
@@ -1221,7 +1299,7 @@ case class TokenExchangeTransaction(buyOrder: BuySellOrder,
       "")
   }
 
-  lazy val token2Tx: PolyTransfer = {
+  lazy val token2Tx: PolyTransfer = { /** STANDUP | Is this supposed to be here? */
```
scasplte2: we should talk about this

src/test/scala/bifrost/transaction/CoinbaseTransactionSpec.scala
scasplte2: could you write a todo to that affect so we can remember this in the future?

src/main/scala/bifrost/BifrostNodeViewHolder.scala
Tuxman: In this case yes, for the box the nonce is the first nonce ever created so it is a good time to use a value of 0. For the block time stamp of 0, well it is time 0. Yes, Intellij will yell at you for using "magic" numbers in other cases, but this at least is a reasonable use case.

src/main/scala/bifrost/inflation/InflationQuery.scala
scasplte2: will move

src/main/scala/bifrost/inflation/InflationQuery.scala
scasplte2: will reformat

src/main/scala/bifrost/state/BifrostState.scala
scasplte2: good enough for now will move in a big bright future with consensus

PR #152, opened 2019-02-05
Ne 2.7 changes

Tuxman:
Changed directory location of data in config files from tmp to home folder, and add in new chainFromRange API endpoint.


src/main/resources/hel.json
```diff
@@ -13,9 +13,9 @@
   },
   "walletSeed": "genesisoo",
   "keyFileDir": "keyfiles/hel",
-  "walletDir": "/tmp/bifrost/hel/wallet",
-  "dataDir": "/tmp/bifrost/hel/data",
-  "logDir": "/tmp/bifrost/hel/log",
+  "walletDir": ".bifrost/hel/wallet",
+  "dataDir": ".bifrost/hel/data",
+  "logDir": ".bifrost/hel/log",
```
scasplte2: do we want this so variable? Should we just put this in a bin folder somewhere?

src/main/scala/bifrost/api/http/DebugApiRoute.scala
```diff
@@ -115,6 +118,32 @@ case class DebugApiRoute(override val settings: Settings, nodeViewHolderRef: Act
...
+              val blocks = view.history.chainBack(startingBlock, view.history.isGenesis, count)
+              SuccessApiResponse(Map(
+                "history" -> blocks.get.map(_._2).map(Base58.encode).mkString(",").asJson
+              ).asJson)
+          }
+        }
+      }
+    }
+  }
+
```
scasplte2: do we know if this works with Yamir's changes?

src/main/resources/hel.json
scasplte2: nvm, I see how you handled it. What you did will be much easier to move forward in the future

PR #158, opened 2019-04-10
Standardized transaction jsons

yamir1995:


src/main/resources/testSettings.json
```diff
@@ -26,5 +26,6 @@
   "agent": "2-Hop",
   "testnet": true,
   "version": [0, 0, 1],
-  "verboseAPI": true
+  "verboseAPI": true,
+  "apiKeyHash": "66akULzvGr1wsoXGzrTTaWrfkafHBukAaTJpM5c2Jtpo"
```
scasplte2: what is the key to produce this hash? Can we make a comment in the JSON file?

src/main/scala/bifrost/BifrostApp.scala
```diff
@@ -33,8 +33,8 @@ import scala.reflect.runtime.universe._
 
 class BifrostApp(val settingsFilename: String) extends GenericApplication with Runnable {
   // use for debug only
-  //  val path: Path = Path ("/tmp")
-  //  Try(path.deleteRecursively())
+  // val path: Path = Path ("/tmp")
```
scasplte2: I don't think we should leave these uncommented in the push.

src/test/scala/bifrost/api/AssetRPCSpec.scala
```diff
@@ -48,7 +49,7 @@ class AssetRPCSpec extends WordSpec
       HttpMethods.POST,
       uri = "/asset/",
       entity = HttpEntity(MediaTypes.`application/json`, jsonRequest)
-    )
+    ).withHeaders(RawHeader("api_key", "test_key"))
```
scasplte2: "test_key" I assume is the key resulting in the hash from above? Is there a tool or something for generating these hashes?

src/main/scala/bifrost/BifrostApp.scala
scasplte2: oh , so sorry. I misread. Bah, this is alright

PR #277, opened 2020-03-10
Reject txs that have matching boxes being created and consumed

Tuxman:
LSMStore catches an edge case where a transfer transaction both consumes and creates a box with a matching Id while still being considered valid at a state validation level. This fixes the issue by enforcing that the transaction cannot even be considered valid and therefore rejected from the mempool and never allowed to be broadcasted.


jingsting:
Noice

Issue #332, opened 2020-04-02
[3pt] Creating monitoring Docker container


scasplte2:

scasplte2:
this has been merged and can be closed

PR #370, opened 2020-05-04
Jaa mon tools #332

scasplte2:
Added provisioning for grafana dashboard and docker images that can be launched locally to monitor a instance of bifrost


scasplte2:
forgot that I need to update all settings file with the changed named for local testnet. Will commit and we should go through a review on this.

PR #526, opened 2020-08-08
Route inv response from Peer Sync through NetwrokController

scasplte2:
I'd like to maintain a consistent approach that all actor route messages that must be sent to remote peers through the network controller.


scasplte2:
also included a couple of other minor changes

PR #544, opened 2020-08-21
refactor state to separate out stateChanges

scasplte2:
Created a dedicated state changes class file instead of returning state changes from a method in an instance of state. Also reorganized State slightly.


Tuxman (approved):
:+1:

PR #558, opened 2020-09-07
Parse in handlers

scasplte2:
Created a trait for "Synchronizers" which are actors that will act as handlers for message from remote peers. In doing this, the NetworkController now just forwards messages to their handler without parsing them (in line with our discussion with @aaronschutza on Friday). This trait requires that partial function handlers be defined to process the message types that each synchronizer actor is registered to handle.

Overall functionality remains largely unchanged.

Documentation has not been updated to reflect these structure changes so DO NOT MERGE these commits yet but please do look over them @Tuxman and @jingsting and ask any questions you might have.


scasplte2:
oh, also got rid of the `DataFromPeer` class as we no longer require a different representation of messages to be passed between the NetworkController and the Synchronizers

Tuxman (approved):
:+1:

PR #571, opened 2020-09-14
Modified delivery tracker 

scasplte2:
Added comments during documentation and updated delivery tracker to limit the number of requests of modifiers


Tuxman (approved):
:+1:

PR #610, opened 2020-09-26
Version update #607 #608 #652

jingsting:
Updated packages; Fixed errors caused by deprecation; iodb version update is ignored since we still need LSMStore for now


src/main/scala/bifrost/network/NetworkController.scala
```diff
@@ -200,17 +200,19 @@ class NetworkController ( settings      : NetworkSettings,
   /**
    * Schedule a periodic connection to a random known peer
    */
-  private def scheduleConnectionToPeer ( ): Unit = {
-    context.system.scheduler.schedule(5.seconds, 5.seconds) {
-      if ( connections.size < settings.maxConnections ) {
-        val randomPeerF = peerManagerRef ? RandomPeerExcluding(
-          connections.values.flatMap(_.peerInfo).toSeq
+  private def scheduleConnectionToPeer (): Unit = {
+    context.system.scheduler.scheduleWithFixedDelay(5.seconds, 5.seconds) (new Runnable() {
```
scasplte2, line 204: Where did this come from and why add the Runnable? It doesn't seem like a terrible idea, just wondering why

scasplte2 (approved):
Overall looks good @Tuxman, we just need to figure out the best way to integrate these changes with the package refactor (maybe a rebase?) then I am happy to import in everything.

Issue #529, opened 2020-08-11
write a test to ensure creation of a peer connection handler


scasplte2:
Ran into a bug where duplicate message codes stopped the creation of a PCH. This should have and should be tested for in the future since the application will run fine when not communicating via P2P if there is a bug with duplicate message codes.

PR #641, opened 2020-10-02
PeerConnectionHandler test #529

Tuxman:


scasplte2 (changes requested):
Should remove `PrivateMethodTester` but otherwise looks good to go 👍

scasplte2 (approved):
gotta keep those peer connections in check 👍

PR #645, opened 2020-10-03
dropping inactive peer connections

scasplte2:
Implementing a method to drop inactive connections after 15 minutes as realized in the latest Scorex PR https://github.com/ScorexFoundation/Scorex/pull/373


scasplte2:
this has been superseded by #688

PR #654, opened 2020-10-06
Ignore confirmed tx

Tuxman:


scasplte2 (approved):
everything appears in working order 👍

PR #660, opened 2020-10-08
Gjallarhorn

Tuxman:
This update is for inclusion of the Gjallarhorn key management and wallet application. The main proof-of-concept components included in this update are:

- Public/private key file creation
- Signing and broadcasting transactions
- Communication with local and remote Bifrost nodes
- Balance query for specified keys from Bifrost


scasplte2 (approved):
Going to update key files to use new scorex tagged types and update CircleCI to run Gjallarhorn tests

PR #675, opened 2020-10-12
Version update; Fixed hostname/port in BifrostApp and docker.scala; Commented out dbOperations

jingsting:


scasplte2 (approved):
keeping up with the Joneses 👍

PR #684, opened 2020-10-16
Refactor genesis

scasplte2:
This PR addresses
- how genesis is initialized
- creates a new trait `GenesisProvider` to define what is required for genesis
- creates an object for dynamically generating new genesis blocks for local testnets
- how keys are generated and stored on disk
- flattens the GNVH and NodeViewHolder


scasplte2:
NOTE: The tests are currently not compiling because we should refactor and centralize the definition of generating a blank state to be shared among the couple of places where genesis state is required.

scasplte2:
@Tuxman one of the TBR tests is failing on Circle but passing for me locally. Any idea what might be different about the circle env to cause the difference?

Tuxman (approved):
👍

PR #706, opened 2020-10-22
Active fork transition

scasplte2:
Provides a method for us to maintain backwards compatibility by designating protocol settings that must be agreed upon to maintain consensus. This mechanism only allows for us to decide which version of the protocol to use. We do not yet have the ability to read in older data.


Tuxman (approved):
Everything is pretty straightforward, no questions. Looks good.

PR #708, opened 2020-10-22
schedule dropping connections

scasplte2:
previously forgot to schedule the dropping dead peer connections in Scorex update


Tuxman (approved):
👍

PR #711, opened 2020-10-24
Change PublicKey25519 to use address

scasplte2:
PublicKey25519Props are now targeted through their address which incorporates an address version and checksum value.


Tuxman (approved):
👍

PR #713, opened 2020-10-28
Add startup seed to command line options

scasplte2:
Added several startup options to the command line arguments that can control application behavior at runtime. Also fixed some minor bugs that @mhalloran21 found in her testing. You should now be able to run bifrost with the following command line arguments.

```
--seed (-s) [String]: applicable to only private or local testnets, this will deterministically generate a set of keys based on the seed. This will also result in the creation of a deterministic genesis block (since the genesis block is based off of the keys included in the block.

--forge (-f): applicable to all networks, this will enable the forger to attempt to forge as soon asthe node initialization has completed
```

As an example usage, we may run Bifrost with
```
sbt "runMain co.topl.BifrostApp --local --forge --seed genesis
```
This set of runtime args would indicate that the node look use a configuration file `local.conf` (as in a local network), try forging as soon as possible, and generate a set of keys (and block) based on the seed `genesis`.


Tuxman (approved):
👍

PR #749, opened 2020-10-29
Transaction by id and invalid boxes

### LINKED ISSUE #735: validBoxes in pickTransactions logic error
The logic was switched to excluding transactions that create new boxes that do not already exist in state. This needs to be switched back and unit tested.

### LINKED ISSUE #736: txById creates incorrectly sized ModifierId
Block Id's are stored in the database with their type Id's prepended to the Id, this function is not removing this Id before creating a ModifierId so it is instantiated as 33 bytes instead of 32.

Tuxman:
Closes #735 
Closes #736 

- The transactionById method in History now returns a proper block ModifierId with the type id chopped off
- Switched the logic to invalidate and remove a transaction if any of the box id's already exist in the state UTXO set


src/main/scala/co/topl/consensus/Forger.scala
```diff
@@ -267,9 +267,10 @@ class Forger (settings: AppSettings, appContext: AppContext )
 
     memPool.take(numTxInBlock(chainHeight)).foldLeft(Seq[Transaction]()) { case (txAcc, tx) =>
       val txNotIncluded = tx.boxIdsToOpen.forall(id => !txAcc.flatMap(_.boxIdsToOpen).contains(id))
-      val validBoxes = tx.newBoxes.forall(b ⇒ state.getBox(b.id).isEmpty)
+      // if any newly created box matches a box already in the UTXO set, remove the transaction
+      val idExists = tx.newBoxes.exists(b => state.getBox(b.id).isDefined)
 
-      if ( validBoxes ) memPool.remove(tx)
+      if (idExists) memPool.remove(tx)
```
scasplte2, line 273: @Tuxman did you make a ticket to move this checking to occur when we place transactions in the mempool? We don't need to do it now, just don't want to forget.

scasplte2 (approved):
:octocat:

src/main/scala/co/topl/consensus/Forger.scala
Tuxman: I believe it is in my notes, I will make a ticket for it though

Issue #750, opened 2020-10-29
WCH only sends Gjallarhorn relevant blocks


mhalloran21:
- when a new block is created, WCH should only send blocks that include keys from gjallarhorn

Issue #659, opened 2020-10-08
Bifrost should send new blocks to Gjallarhorn when app is running


mhalloran21:
- when gjallarhorn runs --> bifrost should know gjallarhorn is running and should now send every new block to gjallarhorn
- when gjallarhorn stops running -> bifrost should remove the gjallarhorn remote actor and no longer try to send new blocks

Issue #632, opened 2020-10-01
Setup method to maintain wallet state


scasplte2:
Have Gajllarhorn "watching" some set key(s), then externally send funds to that account (through Postman), We would want to have Gjallarhorn update it's wallet state automatically through the "watching" integration of Gjallarhorn and Bifrost.

Issue #631, opened 2020-10-01
Diagram out communication patterns between Bifrost and Gjallarhorn


scasplte2:

scasplte2:
Diagram available at https://github.com/Topl/Documentation/blob/master/engineering-diagrams/gjallarhorn/Bifrost_Gjallarhorn_Communication.drawio

PR #768, opened 2020-11-04
Communication from Bifrost to Gjal through Actor system

mhalloran21:
Issues: #630, #750, #632, #631, #659, #674
- set up WCH which allows Gjallarhorn to communicate with Bifrost through akka actor system
- WCH saves the public keys that Gjallarhorn is keeping track of.
- WCH only sends relevant blocks (blocks with the public keys from remote wallet)


Tuxman (approved):
👍

PR #791, opened 2020-11-15
Version update 20201114

jingsting:


Tuxman (approved):
Removed protobuf so we no longer need to support updates for it.

Issue #747, opened 2020-10-29
Generalize attestation process


scasplte2:
Currently or attestation process for proving ownership of a UTXO is very rigid and only allows for a single type of proposition and proof to be used to claim ownership. This process should be generalized to allow for different types of propositions, secrets, proofs, and network specific addresses. This is important to achieve
1. Network specific addresses and transactions
  - a user should receive an error if they try to create a transaction targeting the wrong network
2. Greater extensibility of proof schemes allowed in Bifrost.
  - This extensibility should allow for transactions that claim ownership of UTXOs through multiple proofs including threshold signatures and public key signatures (to name two examples). Additionally, we should allow for the introduction of additional elliptic curve signing routines (this will be needed to transition to Ouroboros).

3. Address standardization
  - This is needed to allow for multiple types of propositions since the proposition itself may be quite large. Adopting a standard approach for how to compute the "evidence" of a proposition and construct an address from that evidence will allow us to maintain a succint and common representation of various propositions. Additionally, this will make our chain more resilient to potential quantum computing attacks in the spirit of the Bitcoin Pay-to-PublicKeyHash type transaction.

PR #807, opened 2020-11-23
#747 - Attestation process feature upgrade

scasplte2:
Generalizing the process for transaction attestation. This involves greatly refactoring transaction creation and validation (syntactic and semantic) as well as the serializers, encoders, decoders.

Changes
* Changed restApi to rpcApi in settings file
* CHanged API endpoints to all be available at the / route
  Use bytes to signify the type of Transaction, Box, Proposition, and Proof expected
* Chagned transaction, box, proposition, and proof to define a generic serializer for sub-types of that class
* Move functionality as high up as it could go
* Migrated toward network specific transactions and addresses
* Included implicit networkType parameter for generating an address from a proposition (prop -> evidence -> address)
* Migrated to a generalized transaction attestation scheme 
* Created an EvidenceProdcuer type class for propositions that defines how evidence is generated from a given proposition
* Transfer transaction now designate fees to be paid in Polys and generate a Poly output box at the head of the newBoxes property


scasplte2:
closed to stop CircleCI from running the tests

PR #852, opened 2020-12-01
4. Add type to modifier

scasplte2:
* Added a moderifier type id to modifier id's
* updated history, storage, and full block processor to use difficulty and height from block
* added block validation rule to check local calculation of difficulty and height
* added filtering of API routes in settings file
* added sorting to Threshold proposition keys for consistent hashing
* modify forger transaction selection strategy to avoid zero value fee transactions
* forger now informs nodeViewHolder of transactions to eliminate so that this update is propogated to other actores


Tuxman (approved):
Went over with James and Jing, everything looks good.

PR #853, opened 2020-12-01
3. Split block into components

scasplte2:
* Added bloom filter and merkle tree to block
* split block into header and body segments
* refactored key ring to work with multiple types of keys


jingsting (approved):
🔢

PR #859, opened 2020-12-02
1. Generalizing tx to address

scasplte2:
* Changed restApi to rpcApi in settings file
* CHanged API endpoints to all be available at the / route
* Use bytes to signify the type of Transaction, Box, Proposition, and Proof expected
* Chagned transaction, box, proposition, and proof to define a generic serializer for sub-types of that class
* Move functionality as high up as it could go
* Migrated toward network specific transactions and addresses
* Included implicit networkType parameter for generating an address from a proposition (prop -> evidence -> address)
* Migrated to a generalized transaction attestation scheme 
* Created an EvidenceProdcuer type class for propositions that defines how evidence is generated from a given proposition
* Transfer transaction now designate fees to be paid in Polys and generate a Poly output box at the head of the newBoxes property


jingsting (approved):
🐶 🚯

scasplte2:
ready to merge this into `dev` @Tuxman ?

PR #867, opened 2020-12-04
added HasName type class #866

### LINKED ISSUE #866: Bug in raw transaction creation

scasplte2:
fixes #866 
This fixes a bug in the transaction method when creating raw transactions. Currently the `getPropTypeString` method tries to pull proposition type of the transaction by looking at the head of the attestation map. However, when creating a raw transaction, this map is empty so an error occurs since there is no head to get. I replaced this functionality with a type class that assumes a function exists that can return the name of the proposition. This allows the type argument P (defined in the Transaction) to pull the HasName type class and implicity query the name of the proposition. 

* Added HasName typeclass to allow abstract type arguments to target concrete values to return

### DIFF
--- a/src/main/scala/co/topl/modifier/transaction/TransferTransaction.scala
+++ b/src/main/scala/co/topl/modifier/transaction/TransferTransaction.scala
@@ -27,7 +27,7 @@ abstract class TransferTransaction[
     val minting: Boolean
   ) extends Transaction[TokenBox.Value, P] {
 
-  lazy val bloomTopics: IndexedSeq[BloomTopic] = to.map(BloomTopic @@ _._1.bytes)
+  lazy val bloomTopics: IndexedSeq[BloomTopic] = to.map(x => BloomTopic.apply(x._1.bytes))
 
   lazy val boxIdsToOpen: IndexedSeq[BoxId] = from.map { case (addr, nonce) =>
     BoxId.idFromEviNonce(addr.evidence, nonce)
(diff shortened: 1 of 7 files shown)


src/main/scala/co/topl/modifier/transaction/TransferTransaction.scala
```diff
@@ -79,90 +76,124 @@ object TransferTransaction {
...
+
+            case _ => false
+          }
           .map {
-            case bx: PolyBox  => ("Poly", s, bx) // always get polys because this is how fees are paid
-            case bx: ArbitBox if txType == "ArbitTransfer" => ("Arbit", s, bx)
-            case bx: AssetBox if (txType == "AssetTransfer" &&
-              bx.assetCode == assetArgs.getOrElse(throw new Error("Undefined assetCode parameter"))._2 &&
-              bx.issuer == assetArgs.getOrElse(throw new Error("Undefined asset issuer parameter"))._1) => ("Asset", s, bx)
+            case bx: PolyBox  => ("Poly", s, bx)
```
Tuxman, line 106: I do not immediately see why the guards can be removed here, but I will make sure to test this situation

scasplte2:
they can't be removed entirely but I was mistakenly only trying to traverse the sequence once. When the `if` is there and you get an ArbitBox but it is a PolyTransfer then a match error was being thrown. As far as I know a map operation can't change the size of a collection so I moved the if guards to a filter instead.

Tuxman (approved):
👍

PR #884, opened 2020-12-11
added TokenValueHolder class and updated boxes

scasplte2:
The point of this PR was to introduce a value class in Token boxes that would allow us put the details of assets into the value field of a box. Now all token boxes have Evidence, Nonce, and Value. From a user point of view, transactions can now be created that distribute assets with unique metadata attached to each asset (up to 128 bytes). 

* Added in TokenvalueHolder class
* Added JSON encoder/decoder and byte serializers/deserializers
* Added consolidation address to raw transaction creation
* Added an AssetCode class for holding asset codes that are composed of an issuer + 8 character string
* Added Util API endpoint `util_assetcode` to aide in the generation of assetcodes given an address and shortname
* Enforce transaction validity on receipt of a persistent modifier (How does Ergo do this?)
* Enforce non-zero fees for all user generated transactions (to enforce consumption of state)
* Enforce the inclusion of the parent block id in the transaction data for Arbit and Poly rewards transactions (to ensure uniqueness of their outputs that is not biasable, previously only the timestamp enforced this)
* Made the `data` field in a transaction optional


Tuxman (approved):
👍

PR #927, opened 2021-01-11
added route to check addresses #925

scasplte2:
* enforce correct network prefix when decoding Addresses from Json
* added an API route for testing whether addresses are valid for an optionally specified network


jingsting (approved):
👍

PR #930, opened 2021-01-12
Api key command line arg closes #928

### LINKED ISSUE #928: Add startup option for API key
Currently the only way to manipulate the API key is through the settings file. Since this is such a common procedure we'd like to add a startup option to avoid users having to create a whole settings file just to manipulate their API key

scasplte2:
* refactored startup ops to a slightly more scalable solution to overwrite application settings
* added `apiKeyHash` as cmd line argument to provide an API key at runtime


jingsting (approved):
🎃

PR #987, opened 2021-02-01
Shutdown when JVMCI option is not found #508

### LINKED ISSUE #508: Gracefully handle error when Bifrost is run in JVM without compiler interface

Tuxman:
closes #508


jingsting (approved):
👍

scasplte2 (approved):
👍

PR #996, opened 2021-02-01
Compare false positive count to a fixed number since the generated addresses are the same #991

jingsting:
…every time


scasplte2 (approved):
👍

PR #1002, opened 2021-02-02
Gjallarhorn can interact with front-end, run offline, and create raw transactions

mhalloran21:
Lots of change to gjallarhorn:
- now interacts with front-end
- can run offline (without bifrost running)
- walletManager actor still runs when offline and keeps track of balances
- can create raw transactions just on gjallarhorn side


Tuxman (approved):
👍

PR #1006, opened 2021-02-03
added type params to StateReader

scasplte2:
fixing a few things to better separate definitions that need to be shared people projects
- refactored the type params out of state reader
- moved box out of state into modifier (these are useful to have outside of Bifrost)
- refactored InvData string function


jingsting (approved):
👍 Looked at this with Nick and James

PR #1009, opened 2021-02-04
added functions and fixed test to resolve #1005

### LINKED ISSUE #1005: Add signing functions to transactions

scasplte2:
* added diskops object to keyring to separate operations that use the disk
* added function to import a keyfile into the keyring
* moved keyring to crypto
* refactored remove keyfile function from api


jingsting (approved):
👍 Reviewed with James and Nick

PR #1013, opened 2021-02-09
fixes #1012

### LINKED ISSUE #1012: Evidence producer includes an extra byte
The evidence producer for PublicKeyCurve25519Propositiona nd Threshold... are including an extra byte due to use of the `.bytes` method

scasplte2:
There was an issue with an extra byte being added to the evidence of a proposition due to using the `.bytes` method. This has been resolved by adding `.tail` to pop off the head byte.


scasplte2:
Merged in with #1028

PR #1014, opened 2021-02-09
added rewards addresses mgmt

scasplte2:
- added routes to manage rewards addresses in forger from the API
- migrated all admin namespace methods to the `.get[A](...)` construction


scasplte2:
Merged in with #1028

PR #1021, opened 2021-02-10
added 128 bit class

scasplte2:
## Summary
This PR implements a 128 bit number as the value for tokens in Bifrost. Currently we are representing this values as Longs which may present an issue of running out of room in the numbers we can represent since the 64 bit. These Long numbers are only able to represent 18 orders of magnitude (in decimal) so we would've been limited to only being able to represent numbers up to ~9.2 billion (with the lower 9 orders used for fractions of a token). While that might have been enough, it doesn't really leave a lot of room for growth if we start off by issuing 200 million coins (we only had about a factor of 20 in head room). The custom 128 bit number now lets us represent numbers up to 10^38, so with fractional tokens the maximum number is now 170 octillion (10^27). 

## Changes
- added custom fixed length 128 bit integer type
  - arthmetic for this type is implemenbted via BigInt but otherwise this number acts like Long, Int, etc.
- replaced the quantity type in AssetValue and SimpleValue to be represented by Int128
- replaced the fee typoe in Transactions to be represented by Int128
- replaced instances of `.sum` with `.reduce(_ + _)` due since sum is a trait of Numeric[T]
- refactored `.time()` to `.time` to avoid unnecessary parentheses
- updated serializers for all modified types (ArbitTransfer, PolyTransfer, AssetTransfer)
- updated JSON encoder/decoders for modified types
  - note that JSON number inputs are now assumed to be Strings in JSON that are then interpreted as Int128
- modified BlockValidator and Forger classes to correctly implement the Int128 classes
- Updated Private genesis block creation
- updated `Balance` API request to return String encoded Int128
- updated some val names in BloomFilter (shouldn't effect anything)

## Testing
- Bifrost passes all unit tests
- Able to spin up a node on a private network and query using Postman

### DIFF
--- a/src/main/scala/co/topl/consensus/BlockValidator.scala
+++ b/src/main/scala/co/topl/consensus/BlockValidator.scala
@@ -52,7 +52,7 @@ class DifficultyBlockValidator(storage: Storage, blockProcessor: BlockProcessor)
     // calculate the adjusted difficulty the forger would have used to determine eligibility
     val timestamp = block.timestamp
     val target = calcAdjustedTarget(parent, parent.height, parent.difficulty, timestamp)
-    val valueTarget = (target * BigDecimal(block.generatorBox.value.quantity)).toBigInt
+    val valueTarget = (target * BigDecimal(block.generatorBox.value.quantity.doubleValue())).toBigInt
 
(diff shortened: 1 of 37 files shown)


scasplte2:
Merged in with #1028

PR #1026, opened 2021-02-11
Private (Local) forging can now recreate keys using seed

scasplte2:
- refactored the Forger startup to allow for keys to attempt to be regenerated if private or local forging
- fixed bug in Int128 sum when seq is empty


scasplte2:
Merged in with #1028

PR #1028, opened 2021-02-11
abstract box reader from state reader to separate defintions

scasplte2:
Separated the stateReader functionality to a more generic trait that I placed under modifier so that I can remove the `state` package in BramblSc. This cleans up the interface and should help us separate out the `topl-definitions` more readily


Tuxman (approved):
:+1

PR #1089, opened 2021-03-09
add string extension method to enforce latin 1 valid strings

scasplte2:
## Summary
Want to enforce Latin-1 encoding on strings and return an error to the user if they provide non-latin-1 characters

## Changes
* Moved Extensions object to root of utils package and refactored affected dependencies
* added StringOps extension object with a function to get the byte array assuming latin-1 encoding
* updated usages of latin-1 encoding to use the new extension method


jingsting (approved):
🥏

PR #1093, opened 2021-03-09
Filter out reward transactions from mempool

### LINKED ISSUE #1062: Filter Arbit and Poly mint transactions when rolling back
Currently, when there is a diverging chain, the coinbase transactions get rebroadcast. This is not appropriate. We should filter these types of transactions out before placing the re-orged transactions back into the mempool.

Tuxman:
closes #1062 

This update will drop reward transactions before they are put into the mempool whenever blocks are rolled back.


jingsting (approved):
🛸

PR #1154, opened 2021-03-29
Add info and idsFromHeight endpoints

### LINKED ISSUE #1043: add back in debug API function to return a given number of block ids from a certain height

### LINKED ISSUE #1060: [API] Add route to get node info
Would be useful to be able to query your node and get info about the version, the network address, etc.

Tuxman:
These are two new endpoints to query block ids from a given height until a given block limit, and an info endpoint to return basic node and network properties.

closes #1060 
closes #1043


src/main/scala/co/topl/http/api/endpoints/DebugApiEndpoint.scala
```diff
@@ -121,4 +122,35 @@ case class DebugApiEndpoint(
         .map(d => d._1.address -> d._2)
         .asJson
     }
+
+  /** #### Summary
+    * Return all block ids from a given height and down to a given limit
```
b-vennes, line 127: If I provided a height of 10 and a limit of 3, would it return 8, 9, and 10? Or 10, 11, and 12?

jingsting (approved):
🌵 nice one!

src/main/scala/co/topl/http/api/endpoints/DebugApiEndpoint.scala
Tuxman: It would return 8, 9, 10

PR #1167, opened 2021-04-01
Persist Consensus State

### LINKED ISSUE #1112: Persist maxStake, inflation, height, and difficulty In Consensus Package Object
The `maxStake`, `inflation`, `height`, and `difficulty` fields in the `consensus` package object should be persisted between restarts.

There also needs to be a way to rollback these fields with reference to a particular block ID.

b-vennes:
## Purpose

- Persist the consensus variables `totalStake`, `inflation`, `difficulty`, and `height` after Bifrost restarts.

## Approach

- Create a consensus storage class that contains an LSM store reference with each variable stored in a key.
  - Allows rollback implementation in the future and persisted global information across process restarts.
- Instantiate the storage class in the Forger actor's Pre-Start method so that it is available for use during forging.

## Tickets

- closes #1112


src/main/scala/co/topl/consensus/ProtocolVersioner.scala
```diff
@@ -19,7 +20,7 @@ class ProtocolVersioner(appVersion: Version, protocolVersions: SortedSet[Protoco
    * @param blockHeight height of the block being considered
    * @return
    */
-  def current(blockHeight: Long): Option[ProtocolSettings] = applicable.find(blockHeight >= _.startBlock)
+  def current(blockHeight: Int128): Option[ProtocolSettings] = applicable.find(blockHeight >= _.startBlock)
```
scasplte2, line 23: @b-vennes @jingsting lets chat about this Int128 for height. I just want to discuss my considerations for what this type should based on it's usage and whether we need to change it everywhere or not.

src/main/scala/co/topl/consensus/package.scala
```diff
@@ -13,41 +13,30 @@ import scala.math.{max, min}
...
   def nxtBlockNum: Int = 3
 
   /** Find the rule set for the given app version and block height */
-  def getProtocolRules(blockHeight: Long): ProtocolSettings =
+  def getProtocolRules(blockHeight: Int128): ProtocolSettings =
     protocolMngr.current(blockHeight)
     .getOrElse(throw new Error("Unable to find applicable protocol rules"))
 
-  def targetBlockTime(blockHeight: Long): FiniteDuration =
+  def targetBlockTime(blockHeight: Int128): FiniteDuration =
```
scasplte2, line 33: this definitely doesn't need to be an Int128

src/main/scala/co/topl/consensus/package.scala
```diff
@@ -13,41 +13,30 @@ import scala.math.{max, min}
...
     getProtocolRules(blockHeight).targetBlockTime.get
 
-  def numTxInBlock(blockHeight: Long): Int =
+  def numTxInBlock(blockHeight: Int128): Int =
```
scasplte2, line 36: nor this, we could decrease this to an Int really

scasplte2 (changes requested):
overall it looks good, lets chat about the types tomorrow at some point 👍

scasplte2 (approved):
👍

PR #1189, opened 2021-04-06
Forger refactoring

b-vennes:
- Moved KeyRing out of Forger and into new actor
- Moved rewards initialization out of Forger
- Replaced several instances of history, state, and mempool with readers
- Added persistent storage for consensus parameters


scasplte2 (approved):
🔥

Tuxman (approved):
:octocat:

PR #1212, opened 2021-04-15
Lift Leader Election Out Of The Forger

### LINKED ISSUE #1198: Lift Leader Election Out Of Forger And Into Object
## Description

As part of the Prosomo upgrade process, leader election can be lifted out of the forger and into a pure function in an external object.

This will make it easier to swap leader election strategies to the one used by Prosomo.

It will also make it easier to unit test leader election separately from the forger, but this is not the primary reason for this ticket.

## Acceptance Criteria

Leader election has been lifted from the Forger and into a Leader Election object.

Leader election is unit tested.

b-vennes:
## Purpose

In preparation for working on Tetra, the leader election process needs to be de-coupled from the forging process.

## Approach

- Created an object containing a static `getEligibleBox` function that determines if the given arbit box is a valid leader for the next block.
- Replaced usages of option and throw with in `attemptForging` an either type with specific failure cases.
- Simplified the calculate target function to only use Int128 and Long with a single division.

## Testing

- Leader Election `getEligibleBox` test that addrsses exist to stake with
- Leader Election `getEligibleBox` test that addresses contain arbit boxes to stake with

## Tickets

- closes #1198


scasplte2 (approved):
This is a good improvement and I like the use of Either's. I'd like to eventually get rid of the error throwing in `forgeBloxkWithBox` but I don't think we need to focus on that at the moment.

jingsting (approved):
🛸 nice job!

PR #1213, opened 2021-04-16
Enforce correct handling of Poly fee

scasplte2:
## Summary
While trying to move genesis funds on a network, I found that the `TokenBoxRegistry` was failing to update due to poorly defined logic for formatting updates to the LSMStore. During my investigation of the initial TBR issue, I found that several edge cases of combinations of TransferTransaction inputs were not correctly handled. This resulted in invalid output states for the transactions. The following work addresses those mistakes and preps some preliminary ideas on how to update the `modifier` package in the future.

## Changes
- Simplified TBR update logic 
  - by attempting to construct a new state for an address by updating it's previous state. If the new state would be empty, that address is marked for deletion.
- Clarified `TransferTransaction` limiting cases
  - all 13 unique cases are given in the more information section below
- Temporary placeholders to prep for `modifier` refactor
- Updated semantic checking to enforce zero-sum for each partial input. 

## Testing
- Able to run all combinations of transaction shown in the figure below, example output is available from James. 
- All unit tests are passing and the AssetTransferSpec has been updating to test for a case where there are no Polys present in a transaction causing semantic validation to fail as expected.

image


jingsting (approved):
👍 Reviewed with everyone

PR #1215, opened 2021-04-16
Rpc Library

SeanCheatham:
## Purpose
- Bifrost's previous RPC implementation lacked type safety, especially around error conditions.  JSON-RPC defines a specific set of error conditions and codes

## Approach
- Created akka-http-rpc module which abstracts the JSON-RPC specification on top of akka-http.
- Created topl-rpc module which defines Topl's RPCs, the parameter types, the response types, and the client and server codecs

## Testing
- Implemented basic unit tests for akka-http-rpc


scasplte2 (approved):
Gone over with the entire Core team and documented in the wiki https://github.com/Topl/Bifrost/wiki/BramblSc-examples

PR #1220, opened 2021-04-19
Chain Program SBT Module

SeanCheatham:
## Purpose
- The Chain Program Engine code is not currently used, but it relies on GraalVM polyglot functionality
- The polyglot functionality causes issues when mixing java8 with java11, so it interferes with the main `node` module

## Approach
- Create new SBT module `chain-program`
- Move Program code to `chain-program` module.  Move tests and related generators to new module.

## Testing
- `sbt compile`, `sbt test:compile`, `sbt it:compile` all work
- `sbt test` passes


scasplte2 (approved):
:octocat:

PR #1221, opened 2021-04-19
Syntactic Validate and Address Check ValidatedNec

SeanCheatham:
## Purpose
- The upcoming RPC refactor leverages a more Cats-y pattern for data validation
- This PR is in preparation for the RPC refactor
- Syntactic Validation and Address Verification are performed by the RPC/API

## Approach
- Refactor syntacticValidate method to use a ValidatedNec pattern instead of throwing exceptions
- Refactor Address validation to return an Either instead of throwing exceptions

## Testing
- `sbt test` passes locally


scasplte2 (approved):
🔥

PR #1230, opened 2021-04-21
Added Brambl package

scasplte2:
## Summary
Adding example usage of the Topl Rpc package and simplified logic for accessing the KeyRing and using keyfiles.

## Changes
- Added Brambl object for providing accessor functionality to import and generate keyfiles. A helper function for signing is also stubbed
- Defined a Provider abstract class to abstract away the definition of the needed request modifier for using the Akka-Http-Rpc package
- Added an extensive set of 'rpcExample' objects that demo usage of the ToplRpc package
- updated the KeyRing class
  - no longer requires a directory to be provided at instantiation time, it is now an option, with error cases thrown if None is available when trying to access DiskOps
  - added an overload to the `generateAttestation` method to sign transaction using multiple keys in a keyring
  - changed the primary keyring constructor to use an `empty` method instad of the `apply` method 
- Split out functionality related to the KeyfileCompanion for PrivateKeyCurve25519
- Removed default value of the `minting` property for ArbitTransfer and PolyTransfer
- Changed the `consolidationAddress` type to be required in the ToplRpc codecs for a raw arbit and raw asset request

## Testing
`sbt compile` and `sbt test:compile` work
`sbt test` passes


Tuxman (approved):
LGTM :octocat:

PR #1265, opened 2021-04-26
Replace Argyle with Mainargs

### LINKED ISSUE #1130: Empty String apiKeyHash Not Interpreted as None
The HttpService class uses an optional `apiKeyHash` to determine if requests should use API key authorization.  The default value in the configuration is `""` (presumably intended to be the equivalent of None or null).  But since this value can technically be Base58-decoded, HttpService enables authorization but no requests will be properly authorized.

Bifrost should instead assume that authorization is always enabled, unless a (new) flag indicates that it should explicitly be disabled.  This means that an `apiKeyHash` value must either be provided, or a (new) flag `authEnabled` must be set to `false`.  The base application.conf should remove or comment out the line `apiKeyHash = ""` to make it `null`.  If it is never overridden in a custom configuration and if a command line option is not provided, then the node should crash/terminate at startup.

Acceptance Criteria:
- Bifrost authorization is enabled by default
- Bifrost authorization can be disabled when desired (like local environments)
- When authorization is enabled, an `apiKeyHash` is also required.  A default should _not_ be provided in configuration.

### LINKED ISSUE #1139: Replace Argyle with Mainargs
The library used by Bifrost to parse command-line arguments, Argyle, is no longer maintained.  In particular, there are no releases for Scala 2.13, which prevents Bifrost from upgrading.

Instead, we can use [mainargs](https://github.com/com-lihaoyi/mainargs) which is actively maintained in both 2.12 and 2.13.  The interface is different though, so we will need to modify how we capture startup args.

Acceptance Criteria:
- Argyle is no longer a dependency
- Usage of Argyle is replaced by `mainargs`

Tuxman:
## Purpose
Argyle is no longer being actively supported or maintained, and blocks us from upgrading to 2.13.

## Approach
- 1-to-1 swap of Argyle definitions to MainArgs annotations
- keep very close to the same behavior at the user level

## Testing
- Testing the flags was done manually to ensure compatibility with running the same CLI option flags.

## Tickets
- closes #1139 
- closes #1130


node/src/main/scala/co/topl/http/HttpService.scala
```diff
@@ -95,10 +95,12 @@ final case class HttpService(apiServices: Seq[ApiEndpoint], settings: RPCApiSett
...
-    optionalHeaderValueByName("x-api-key") { keyOpt =>
-      if (isValid(keyOpt)) route
-      else complete(HttpEntity(ContentTypes.`application/json`, "Provided API key is not correct"))
-    }
+    if (settings.authEnabled) {
+      optionalHeaderValueByName("x-api-key") { keyOpt =>
+        if (isValid(keyOpt)) route
+        else complete(HttpEntity(ContentTypes.`application/json`, "Provided API key is not correct"))
+      }
+    } else complete(HttpEntity(ContentTypes.`application/json`, "Requests are not authorized for this node."))
```
SeanCheatham, line 103: I think this `else` statement should just return `route` instead of completing the request

node/src/main/scala/co/topl/settings/StartupOpts.scala
```diff
@@ -33,9 +31,15 @@ object StartupOpts {
...
   * @param apiKeyHash hash of API key
   */
+@main
 final case class RuntimeOpts(
+  @arg(name = "seed", short = 's', doc = "string to deterministically generate keys on private and local networks")
   seed:           Option[String] = None,
-  forgeOnStartup: Boolean = false,
+  @arg(name = "forge", short = 'f', doc = "enable forging as soon as the node starts")
+  forgeOnStartup: Flag = Flag(),
+  @arg(name = "authEnabled", short = 'a', doc = "Allow the node to receive API requests")
```
SeanCheatham, line 40: Should this flag be flipped to `disableAuth` instead?  I think James mentioned the ideal behavior is to enable auth by default, and only disable it if intentionally specified by the user.  If I understand the `mainargs` docs correctly, a Flag will default to `false` unless `--authEnabled` is provided as an argument.

Tuxman: In #1130 @scasplte2 left a comment saying he would like the API turned off by default.

Tuxman: Never mind I misunderstood, yes you are correct.

node/src/main/scala/co/topl/http/HttpService.scala
Tuxman: I reversed the logic and return route when `disableAuth` is true, else move to the API key check

scasplte2 (approved):
:octocat:

PR #1272, opened 2021-04-27
Balances RPC Response Boxes Type

### LINKED ISSUE #1271: topl-rpc boxDecoder Always Decodes to ArbitBox
The implicit `Decoder[Box]` always decodes to an ArbitBox.  The ArbitBox decoder always succeeds on anything that can be decoded to a `SimpleValue`, and as such, the current implementation of the `Box` decoder always results in an `ArbitBox` even if a `PolyBox` or `AssetBox` should be returned.

`co.topl.rpc.SharedCodecs#tokenBoxDecoder` should be updated to use a discriminator on the `type` field, and it should return the proper Box sub-type using the proper decoder.

Acceptance Criteria:
- Boxes decode into the proper Box type for the `topl_balances` RPC

SeanCheatham:
## Purpose
- A bug in the topl-rpc Box decoder causes all returned boxes in Balances.Response.{}.Boxes to be an `ArbitBox`
- Balances.Response.{}.Boxes was a generic Map but could be made into a type-specific case class instead

## Approach
- Added `Balances.EntryBoxes` case class
- Implemented corresponding codecs

## Testing
- Implemented unit tests for EntryBoxes codecs

## Tickets
- closes #1271

### DIFF
--- a/topl-rpc/src/main/scala/co/topl/rpc/ToplRpc.scala
+++ b/topl-rpc/src/main/scala/co/topl/rpc/ToplRpc.scala
@@ -152,8 +152,9 @@ object ToplRpc {
         */
       case class Params(addresses: List[Address])
       type Response = Map[Address, Entry]
-      case class Entry(Balances: EntryBalances, Boxes: Map[String, List[TokenBox[_]]])
+      case class Entry(Balances: EntryBalances, Boxes: EntryBoxes)
       case class EntryBalances(Polys: Int128, Arbits: Int128)
+      case class EntryBoxes(PolyBox: List[PolyBox], ArbitBox: List[ArbitBox], AssetBox: List[AssetBox])
     }
 
     object TransactionById {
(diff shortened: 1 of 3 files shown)


topl-rpc/src/main/scala/co/topl/rpc/ToplRpc.scala
```diff
@@ -152,8 +152,9 @@ object ToplRpc {
         */
       case class Params(addresses: List[Address])
       type Response = Map[Address, Entry]
-      case class Entry(Balances: EntryBalances, Boxes: Map[String, List[TokenBox[_]]])
+      case class Entry(Balances: EntryBalances, Boxes: EntryBoxes)
```
SeanCheatham, line 155: The NodeView API Endpoint class constructs the Map by "grouping" by each TokenBox's `typeString` value (`PolyBox`, `ArbitBox`, and `AssetBox`).  These are the only three values that would ever be included in the Map, so they can be represented as a product ADT instead of a generic map.

b-vennes (approved):
🔥

scasplte2 (approved):
looks good 👍

PR #1278, opened 2021-04-29
Semantic Validation Refactor

### LINKED ISSUE #1273: Semantic Validation "Either"
Semantic Validation for transactions is performed in a method `semanticValidation`.  This method currently returns a `Try[Unit]`, but it should instead return an `Either[SemanticValidationFailure, TX]`, where SemanticValidationFailure is a new ADT which defines the subset of failure conditions for semantic validation.  This should return an Either instead of a ValidatedNec because this validation should fail as eagerly as possible.

Acceptance Criteria:
- New `SemanticValidationFailure` ADT exists for the semantic validation failure conditions
- `semanticValidation` returns an `Either[SemanticValidationFailure, TX]`.

SeanCheatham:
## Purpose
- Syntactic Validation was recently refactored to adopt a cats.Validated pattern for error handling
- This PR performs similar functionality for Semantic Validation

## Approach
- Refactored Syntactic Validation into a "typeclass", with Ops generation using the Simulacrum library
- Re-implemented Semantic Validation using the cats.Validated pattern
- Change Transaction and TransferTransaction type bounds usage to make the implicit evidence values of the class

## Testing
- Refactored all transaction validation tests into a single file
- Implemented a few new tests for some of the error conditions

## Tickets
- closes #1273


common/src/main/scala/co/topl/modifier/transaction/AssetTransfer.scala
```diff
@@ -62,17 +62,17 @@ object AssetTransfer {
...
-  ): Try[AssetTransfer[P]] = {
+    boxReader:                   BoxReader[ProgramId, Address],
+    toReceive:                   IndexedSeq[(Address, AssetValue)],
+    sender:                      IndexedSeq[Address],
+    changeAddress:               Address,
+    consolidationAddress:        Address,
+    fee:                         Int128,
+    data:                        Option[String],
+    minting:                     Boolean
+  )(implicit evidenceProducerEv: EvidenceProducer[P], identifiableEv: Identifiable[P]): Try[AssetTransfer[P]] = {
```
scasplte2, line 75: what motivated you to go away from the context bounds and use the implicit arguments instead?

node/src/test/scala/co/topl/modifier/transaction/AssetTransferSpec.scala
```diff
@@ -1,9 +1,7 @@
 package co.topl.modifier.transaction
```
scasplte2, line 2: Do we need these specs anymore?

scasplte2 (approved):
🔥 solid work

PR #1329, opened 2021-05-13
Migrate Tests to Modules

### LINKED ISSUE #1321: Refactor 'common' unit tests out of 'node'
The `node` SBT module currently contains unit tests related to code that is defined in the `commons` module.  Where applicable, these tests should be moved into the `commons` module instead.  This will likely require moving several of the generators as well.

Acceptance Criteria:
- Unit tests specific to functionality defined in the `common` module should be located in the `common` module

SeanCheatham:
## Purpose
- The unit tests for the `node` module run tests on code implemented in `common`.  The tests and the implementations should be coupled

## Approach
- Split up the test Generators and moved generators related to `common` to the common module.  `node` now depends on `common` in the test scope
- Move unit tests related to common over to the common module
- Refactored several tests to leverage beforeAll for test setup

## Testing
- Unit tests pass locally

## Tickets
- closes #1321


node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
```diff
@@ -10,12 +10,20 @@ import org.scalatest.wordspec.AnyWordSpec
 
 class NodeViewRPCSpec extends AnyWordSpec with Matchers with RPCMockState {
 
-  val txs: Seq[TX] = bifrostTransactionSeqGen.sample.get
-  val txId: String = txs.head.id.toString
-  val block: Block = blockGen.sample.get.copy(transactions = txs)
+  var txs: Seq[TX] = _
```
scasplte2, line 13: what is happening here?

node/src/test/scala/co/topl/nodeView/state/MockState.scala
```diff
@@ -1,52 +1,31 @@
...
-  val genesisBlock: Block = PrivateGenesis(keyRing.addresses, settings).getGenesisBlock.get._1
-
-  val genesisBlockId: ModifierId = genesisBlock.id
-
-  def createState(genesisBlockWithVersion: Block = genesisBlock): State = {
-    val file = createTempFile
-    val tempSettings = settings.copy(
-      application = settings.application.copy(
-        dataDir = Some(file.getPath + "data")
+  def createState(genesisBlockWithVersion: Block = genesisBlock, inTempFile: Boolean = true): State = {
```
scasplte2, line 18: can we just do this all in memory? I've always disliked that this goes through disk.

scasplte2 (approved):
🔥

PR #1330, opened 2021-05-14
TransactionTest Fee Tests

### LINKED ISSUE #1281: Transaction Fee Integration Test
The TransactionTest suite runs basic Poly, Arbit, and Asset transactions with 0 fees.  Additional tests should be written which sends transactions with an associated fee.  The test should verify that the forging node receives the poly fees after forging the block for the transaction.

Acceptance Criteria:
- TransactionTest integration test includes new transaction tests which include fees
- The forging node should receive the poly fee

Note:
- A suggested strategy is to lock out all but one keyfile on the single forging node
- After the transaction, verify that the one keyfile's address has received the poly funds

SeanCheatham:
## Purpose
- Our integration tests need to verify that transaction fees are properly awarded to a particular forging/rewards address

## Approach
- Added Poly and Arbit tests with fees
- The balance for the rewards address is checked before and after the transaction

## Testing
- New tests pass locally
image

## Tickets
- closes #1281


node/src/it/scala/co/topl/it/util/NodeRpcApi.scala
```diff
@@ -62,6 +65,16 @@ case class NodeRpcApi(host: String, rpcPort: Int)(implicit system: ActorSystem,
...
+
+  def pollForTransaction(transactionId: ModifierId, timeout: FiniteDuration = 20.seconds)(implicit
+    scheduler:                          Scheduler
+  ): EitherT[Future, RpcClientFailure, TransactionById.Response] =
+    Retry(
+      () => ToplRpc.NodeView.TransactionById.rpc.call.apply(ToplRpc.NodeView.TransactionById.Params(transactionId)),
+      interval = 500.milli,
+      timeout = timeout,
+      attempts = Int.MaxValue
+    )
```
scasplte2, line 77: I know this is bordering on the event sourcing thing but is it possible (reasonable) to expose a function like this on the node API? I primarily worry about abuse but if that can be mitigated it would extremely useful, at least in the short term.

scasplte2 (approved):
👍

PR #1335, opened 2021-05-18
Refactor NodeViewSynchronizer initialization

scasplte2:
## Purpose
Currently, when restarting Valhalla, I observe the NetworkController receiving many unhandled messages from the NodeViewSynchronizer due to the setup of the NVS preStart function which schedules a peer connection attempt. This scheduling should be delayed until the NodeviewHolder is ready since the NetworkController can't process any messages until this point.

## Changes
- Moved the scheduling invocations to occur when the `NodeViewReady` message is received

## Testing
- Ran in local environment
- All tests passing

### DIFF
--- a/node/src/main/resources/application.conf
+++ b/node/src/main/resources/application.conf
@@ -2,7 +2,7 @@ bifrost {
 
   application {
     # application version of this software
-    version { value = 1.4.3 }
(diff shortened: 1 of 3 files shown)


Tuxman (approved):
Ran it successfully on my machine, everything checks out.

PR #1357, opened 2021-05-18
AppSettings lenses

### LINKED ISSUE #1125: Implement AppSettings Lens
Applying changes to mutable objects (potentially deeply-nested) is very easy.  Applying changes to immutable objects is a bit more difficult (especially for deeply-nested structures).  Scala case classes automatically generate `.copy` methods which make it easier to return updated copies of an object.  But they are still cumbersome to use when attempting to update a deeply nested item since they require multiple levels of `.copy` calls.

Lenses are a way of drilling down into these sorts of structures.  They provide a much easier way for modifying deeply nested pieces of data while still preserving the notion of immutability.  Lenses can be functionally composed, meaning nested Lenses play very well with each other.

Implement Lens functionality in the AppSettings case class and its nested case classes.  This should hopefully be as simple as using the compile time macros.  Use these Lenses in StartupOpts when applying command line arguments to AppSettings.  If applicable, do the same in other cases.

Acceptance Criteria:
- AppSettings and its related case classes have corresponding Lenses
- StartupOpts uses the AppSettings lenses when applying command line arguments

See:
- https://www.scala-exercises.org/monocle/lens
- https://www.optics.dev/Monocle/

Tuxman:
## Purpose
Copying and replacing every field in a large, nested case class leads to verbose code. This PR removes the boilerplate by introducing lenses, that can be generated for every field in a case class. Now whenever many changes need to be made to a single instance of a case class, in particular `AppSettings`, these much more concise lenses can be used instead.

## Approach
* Implement `Focus` macro from the Monocle library to generate lenses for each value being modified from `AppSettings` in `StartupOpts`
* Replicate each copy operation as a `modify` or `replace`

## Testing
* Manually tested all startup flags to make sure behavior is preserved for every option

## Tickets
* closes #1125

### DIFF
--- a/node/src/main/scala/co/topl/settings/StartupOpts.scala
+++ b/node/src/main/scala/co/topl/settings/StartupOpts.scala
@@ -2,6 +2,7 @@ package co.topl.settings
 
 import co.topl.utils.NetworkType
 import mainargs.{arg, main, Flag, ParserForClass, TokensReader}
+import monocle.syntax.all._
 
 /**
  * Parameters that are given at application startup. Only parameters that are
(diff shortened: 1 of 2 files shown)


scasplte2 (approved):
👍

PR #1360, opened 2021-05-20
Moved EDDSA files from Crypto repo

jingsting:


jingsting:
The EDDSA files are from Topl/Crypto  https://github.com/Topl/Crypto/tree/master/src/main/scala/crypto/primitives/eddsa

scasplte2 (approved):
👍

PR #1367, opened 2021-05-25
Resolve protocol conflict

scasplte2:
There are patch updates on main causing a conflict with the protocol version defined in application.conf for the latest update. This resolves the conflict to the latest protocol version (v1.6.3)


Tuxman (approved):
👍

PR #1396, opened 2021-06-02
crossbuild-patch

Tuxman:
This is a patch to fix some Scala version incompatibilities that use `LazyList` or `BuildFrom`, which are only in Scala 2.13.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -104,8 +104,9 @@ object Dependencies {
 
   lazy val common: Seq[ModuleID] = {
     Seq(
-      "com.typesafe.akka" %% "akka-actor" % akkaVersion,
-      "org.typelevel"     %% "simulacrum" % "1.0.0"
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -104,8 +104,9 @@ object Dependencies {
 
   lazy val common: Seq[ModuleID] = {
     Seq(
-      "com.typesafe.akka" %% "akka-actor" % akkaVersion,
-      "org.typelevel"     %% "simulacrum" % "1.0.0"
+      "com.typesafe.akka"      %% "akka-actor"              % akkaVersion,
+      "org.typelevel"          %% "simulacrum"              % "1.0.0",
+      "org.scala-lang.modules" %% "scala-collection-compat" % "2.4.4"
```
scasplte2, line 109: is there a reason you added this to separate modules instead of having a shared dependency fo scala-collection-compat?

PR #1399, opened 2021-06-07
String Encoded Types

### LINKED ISSUE #1061: Unit test for non-Latin 1 encoding characters
Test to make sure anywhere that Latin 1 encoding is used that non-Latin 1 characters will not be accepted.

### LINKED ISSUE #1308: Create String Encodings As New Types
## Description

Encoded strings should be their own value types around the string class with validation required.

The following encodings should have custom strings:

- Base58String
- Base16String
- UTF8String
- Latin1String

## Acceptance Criteria

Encoded string types can be created with validation.

String instances have been replaced with the type-safe versions.

Unit tests are written for string types and encoding methods.

b-vennes:
## Purpose

The Bifrost node uses many different types of encoded strings with different purposes in different places. At the moment, all of these values are represented by the default `String` type. This is all fine and good until we want to get the underlying byte representation (for the purpose of hashing, for example). In this case, we need to know what type of encoding was used so that we can decode it in the same way. Thus, having higher-level types for the encodings UTF-8, Latin-1, Base-16, and Base-58 would be useful.

## Approach

I've made use of the Newtype library for creating wrappers around existing types with additional validation. The new types are `Utf8String`, `Latin1String`, `Base16String`, and `Base58String`. Each type can only be instantiated using a `validated` or `unsafe` function. In the case of `validated`, the result will be a cats `Validated` instance. In the case of `unsafe`, an error will be thrown. Along with these methods, I've implemented the `Eq` and `Show` typeclasses for each new type. These will let us use the typesafe `===` and `.show` functions from cats.

## Testing

- added `StringTypesSpec` tests for testing the validation and typeclasses for all four string types

## Tickets

- closes #1308
- closes #1061


common/src/main/scala/co/topl/attestation/Proof.scala
```diff
@@ -85,19 +93,26 @@ object SignatureCurve25519 {
   lazy val genesis: SignatureCurve25519 =
     SignatureCurve25519(Signature(Array.fill(SignatureCurve25519.signatureSize)(1: Byte)))
 
-  def apply(str: String): SignatureCurve25519 =
+  def apply(data: Base58Data)(implicit d: DummyImplicit): SignatureCurve25519 =
```
scasplte2, line 96: Did you figure out if the DummyImplicit is still needed here?

common/src/main/scala/co/topl/attestation/Proposition.scala
```diff
@@ -126,9 +154,18 @@ object ThresholdPropositionCurve25519 {
...
-      case Success(prop: ThresholdPropositionCurve25519) => prop
-      case Success(_)                                    => throw new Error("Invalid proposition generation")
-      case Failure(ex)                                   => throw ex
+      case Right(prop: ThresholdPropositionCurve25519) => prop
+      case Right(_)                                    => throw new Error("Invalid proposition generation")
+      case Left(failure) =>
+        throw new Error(s"Failed to create Threshold Curve 25519 proposition from string: ${failure.show}")
+    }
+
+  def fromBase58(data: Base58Data): ThresholdPropositionCurve25519 =
```
scasplte2, line 163: Could we make this a function `def fromBase58[T](data: Base58Data): T` and put it in the Proposition singleton maybe? Just thinking to reduce the like code.

b-vennes: I made some modifications to proof so that it handles the result from parsing the proof type. Let me know if you like it or not.

common/src/main/scala/co/topl/attestation/Proof.scala
b-vennes: I removed the dummy implicits that were no longer necessary, but some are still required since `Signature` and `Base58Data` are both newtypes and have the same type of `Object` after type erasure.

common/src/main/scala/co/topl/attestation/Proposition.scala
scasplte2: Looks good 👍

scasplte2:
@b-vennes If y'all are ready, I think this can be merged. Didn't want to step on toes in case this would mess anyone up but if not please go ahead and merge it in.

scasplte2 (approved):
ready to merge 🔥

PR #1409, opened 2021-06-22
BIP-39 Mnemonic Seed

### LINKED ISSUE #1319: Add BIP 39 To Key Management
## Description

Add a BIP 39 implementation to Bifrost in the `common/keymanagement` package.

An implementation currently exists in the `ghallarhorn` project. It needs to be copied to `keymanagement` and cleaned up if necessary.

## Acceptance Criteria

- BIP 39 mnemonic phrases can be generated from a random UUID
- seeds can be generated from BIP 39 phrases of length 12, 15, 18, 21, and 24 with an optional password
- BIP 39 unit tests added and are passing

### LINKED ISSUE #1385: PBKDF2 Key Generator
## Description

PBKDF2 is a requirement of the [BIP-0039 spec] and the implementation is non-trivial.

The building blocks for PBKDF2 are available from bouncy castle, but it will need to be constructed and tested.

## Acceptance Criteria

- A key can be generated using a PBKDF2 function from a message, salt, iteration count, and key size
  - the only pseudo random function needed is HMAC-SHA512

b-vennes:
## Purpose

Addresses and keys are a pain to keep track of, especially when someone wants to manage funds in different places. BIP-39 mnemonic seeds allow users to manage all of their addresses and keys under one main secret code called a `mnemonic phrase`. This mnemonic phrase is a set of words that can be safely converted to a seed byte string using the BIP-39 protocol.

The seed can then be combined with an HD-wallet protocol like BIP-32 or SLIP-23 which can generate millions of addresses and keys from a starting seed. SLIP-23 will be developed in a separate PR.

## Approach

The mnemonic seed system was designed around type-safety and security of the mnemonic phrase. To generate a phrase, one must provide a valid phrase, the expected length type, and the expected language. With this info provided, the mnemonic seed is guaranteed to be valid.

The `Mnemonic` type returned from `Mnemonic.fromPhrase` is an alias for a function that takes in an option string password and returns a byte array. This representation of the mnemonic phrase was used so that the secret phrase can't be leaked by an accidental logging of the mnemonic value.

## Testing

Tests were added which use test vectors from https://github.com/trezor/python-mnemonic/blob/master/vectors.json.

## Tickets

- closes #1319
- closes #1385


common/src/main/scala/co/topl/attestation/keyManagement/mnemonicSeed/Language.scala
```diff
@@ -0,0 +1,55 @@
...
+ * @param hash the SHA-256 hash of the words for verification
+ */
+sealed abstract class Language(val filePath: String, val hash: String) {
+  val wordlistDirectory: String = "bip-0039"
+
+  /**
+   * Gets the valid set of words for the language.
+   * @return either a read failure or the word list
+   */
+  def getWords: Either[ReadWordListFailure, List[String]] =
```
SeanCheatham, line 23: I think this should be a `val` to avoid re-reading from disk each time this method is called

PR #1431, opened 2021-07-09
ThresholdProposition fix and tests

jingsting:
## Summary
Fixed the validation function in threshold proposition and added tests for it

## Changes
- Fixed and updated isValid in ThresholdSignatureCurve25519 so it checks if enough number of unique proposition and signature pair exists
- Moved testsGenesis file from main to test

## Testing
- Added valid threshold transfer generators
- Tested threshold transfers in RPC tests
- Added funded threshold addresses in testGenesis


scasplte2 (approved):
👍

PR #1466, opened 2021-07-27
Add apply method to HiccupBlock

b-vennes:
## Purpose

Creating a `HiccupBlock` to check against the existing list of hiccup blocks for semantic and block validation can be simplified by creating an `apply` method which converts a block into a `HiccupBlock`.

## Approach

Added an `apply` method inside of a new `HiccupBlock` companion object.


scasplte2:
should this target the hotfix branch or is this replacing that branch?

b-vennes:
> should this target the hotfix branch or is this replacing that branch?

This is meant to be additive to the leveldb branch that will become 1.7

scasplte2 (approved):
👍

PR #1443, opened 2021-07-19
Switch out IODB for LevelDB

Tuxman:
Purpose
	IODB is no longer maintained and has clear performance hits when the blockchain grows beyond a few thousand blocks. Replacing IODB with LevelDB will allow us to use a much more common and standard database for storing blocks and block data, and immediately benefit performance without any extra optimisations.
Approach
- Replace all instances of IODB storage with LevelDB
- Update rollback to `rollbackTo` api that uses separate database to store rollback data
- Update tests to reflect minor api changes
Testing
- All unit tests pass
- a node running LevelDB can successfully sync and stay synced with a node running IODB
- a node running LevelDB can sync to the Valhalla network and correctly read and store blocks

Tickets
- closes #439

PR #1286, opened 2021-05-03
Crypto Module

b-vennes:
## Purpose

- Remove dependency on scorex for cryptographic primitives
- Refactor forked scorex scrypto library to use typeclasses, validation, and eithers
- Improve crypto/key management code where-ever possible

## Approach

- Refactored hashing functions to implement a `Hash[H, D]` typeclass pattern
- Replace usages of supertagged with newtypes
- Implement a `Digest[T]` typeclass for working with digests of size 32 and 64
- Move encoding types into `co.topl.utils`
- Return an either type from hashing to account for the possibility of a digest failing to be valid.

## Testing

- Created tests for values and types pulled in from scrypto

## Tickets

- closes #1225 
- closes #1269
- closes #1234
- closes #1226
- closes #1222 
- closes #1363
- closes #1364

PR #1475, opened 2021-07-29
v1.6.3 -> 1.7.1

scasplte2:
## Summary
This release is largely focused on scalability and performance enhancements. The most notable of which is the transition from IODB -> LevelDB. Additionally, there have been several improvements in our usage of value classes in the String encoded types. Improvements in Crypto module and seed generation are targeting functionality required for Tetra address management. 

We also fixed a number of pressing bugs that had been introduced in the latest releases. As of 2021-07-29, priority hotfixes for v1.4.4 and v1.6.3 clients are to be releases shortly to address an overflow introduced in v1.4.4. This latest version includes the hotfix and several additional rules defining checkpointed blocks due to (i) invalid child block timestamp (ii) invalid proposition included in reward minting transaction. Finally, the collections changes between Scala 2.12 -> 2.13 required us more specifically define the ordering of keys in the attestation data of a transaction.

### Feature
- #1443 - [Feature] Switch out IODB for LevelDB
- #1409 - [Feature] Bip39 seed generation from mnemonic
- #1399 - [Feature] String encoded types
- #1286 - [Feature] crypto module

### Patch
- #1464 - [Patch] Signing function update
- #1437 - [Patch] Transaction data length fix and test
- #1461, #1466, #1453 - [Patch] Ignore block validation for Hiccup blocks impacted by overflow and proposition failure
- #1452 - [Patch] Fix bug in target overflow
- #1452 - [Patch] Transaction attestation maps transitioned to ListMap
- #1360 - [Patch] Move in EDDSA file from crypto repo
- #1418 - [Patch] Deterministically order genesis pre-mine addresses using ListMap
- #1396 - {Patch] Scala crossbuild bug fix
- #1480 - [Patch] V1.7.0 Integration Test updates

## Testing
- Unit tests
- Integration tests
- Demo bootstrap from Valhalla genesis


node/src/main/scala/co/topl/consensus/ConsensusStorage.scala
```diff
@@ -128,12 +122,13 @@ object ConsensusStorage {
     }
     val file = new File(s"$dataDir/consensus")
     file.mkdirs()
-    val storage = new LSMStore(file)
+
+    val versionedStore = new LDBVersionedStore(file, 100)
```
SeanCheatham, line 126: Was there a new decision on the use of `100` here?  100 rollbacks could potentially be too low if the nodes become hundreds of blocks out-of-sync.

common/src/main/scala/co/topl/utils/StringDataTypes.scala
```diff
@@ -0,0 +1,153 @@
...
+    implicit val showLatin1Data: Show[Latin1Data] =
+      latin1Data => new String(latin1Data.value, StandardCharsets.ISO_8859_1)
+  }
+
+  /**
+   * Byte data represented by Base-58 encoded text.
+   * @param value the underlying bytes data
+   */
+  @newtype
+  class Base58Data(val value: Array[Byte])
```
SeanCheatham, line 75: At the moment I can't find any cases where we improperly check equality on Base58Data, but it's something we may need to keep an eye on based on our findings with Latin1Data

node/src/main/scala/co/topl/consensus/ConsensusStorage.scala
SeanCheatham: We only have one forger right now, so rollbacks shouldn't happen.  We'll need to check in on this issue though.

SeanCheatham (approved):

Tuxman (approved):

PR #1480, opened 2021-08-02
V1.7.0 Integration Test Fixes

SeanCheatham:
# Purpose
- TimestampValidation checks the timestamp of a block's parent by searching in Storage (the canonical chain).
  - But TimestampValidation can be invoked for a block meant to go into BlockProcessor (the tine cache) instead of the canonical chain
  - This results in an exception and node will have difficulty syncing
- StartupOpts does not consider a "seed" from the config file
- Integration Test container enable debug port

# Approach
- TimestampValidation will search BlockProcessor and then Storage for a parent timestamp
- StartupOpts read "seed" from config file
- Enable port 5005 and command line arg for JMX debugging in integration tests

# Testing
- Ran integration tests locally, including ChainSelectionTest 
- Ran unit tests locally


scasplte2:
All integration tests now pass 👍

scasplte2 (approved):
💯

PR #1483, opened 2021-08-02
Fix None.get from .sample.get

### LINKED ISSUE #1473: Remove ".sample.get" Calls on Unit Test Generators
Scalatest generators have the ability to return a single generated value using `.sample`.  However, this returns an Option which may result in an unexpected NoSuchElementException when calling `.get`.  This will cause the test to fail even though the test never actually runs as desired.

We should replace these `.sample.get` calls with either `CommonGenerators#sampleUntilNonEmpty(generator)`, or do `generator.pureApply(Gen.Parameters.default, Seed.random())`.

Acceptance Criteria:
- `.sample.get` is not used on any Generators in the test cases

SeanCheatham:
# Purpose
- scalacheck generators generate new data most of the time, but sometimes a sample attempt may not yield a result.  In these cases, it should be tried again
- In many places in our unit tests, we only attempt a single generator sample which may result in a None.get which will fail the test

# Approach
- Implemented a GeneratorOps implicit class which adds a `.sampleFirst()` method to a `Gen`
- Updated references to `.sample.get` with `.sampleFirst()`

# Testing
- Unit tests pass locally.  Hopefully they continue to pass everywhere else!

# Tickets
- closes #1473

### DIFF
--- a/common/src/test/scala/co/topl/modifier/transaction/proposition/MultiSignatureCurve25519Spec.scala
+++ b/common/src/test/scala/co/topl/modifier/transaction/proposition/MultiSignatureCurve25519Spec.scala
@@ -7,6 +7,7 @@ import co.topl.attestation.{
   ThresholdSignatureCurve25519
 }
 import co.topl.utils.CommonGenerators
+import co.topl.utils.GeneratorOps.GeneratorOps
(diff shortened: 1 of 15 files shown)


scasplte2 (approved):
seems like this was a simple enough fix 👍

PR #1604, opened 2021-09-07
2.12 Compile Fixes

SeanCheatham:
# Purpose
- ArraySeq exists in Scala 2.13 but not in 2.12
- WrappedArray exists in 2.12 but is deprecated in 2.13
- Compile issues in Test
- CircleCI issues when running checkPR

# Approach
- Re-integrate 2.12 types
- Add checkPR step to verify 2.12 compiles
- Specify CircleCI resource_class.  Remove SBT settings.

# Testing
- `checkPR` passes locally and in CI

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,6 @@ val scala213 = "2.13.6"
 inThisBuild(List(
   organization := "co.topl",
   scalaVersion := scala213,
-  crossScalaVersions := Seq(scala212, scala213),
   versionScheme := Some("early-semver"),
   dynverSeparator := "-",
   version := dynverGitDescribeOutput.value.mkVersion(versionFmt, fallbackVersion(dynverCurrentDate.value)),
@@ -33,6 +32,7 @@ lazy val commonSettings = Seq(
       case _                       => sourceDir / "scala-2.12-"
     }
   },
+  crossScalaVersions := Seq(scala212, scala213),
   Test / testOptions ++= Seq(
     Tests.Argument("-oD", "-u", "target/test-reports"),
     Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "2"),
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -328,5 +326,5 @@ lazy val loadTesting = project
   )
   .dependsOn(common, brambl)
 
-addCommandAlias("checkPR", "; scalafixAll --check; scalafmtCheckAll; test")
-addCommandAlias("preparePR", "; scalafixAll; scalafmtAll; test")
+addCommandAlias("checkPR", s"; scalafixAll --check; scalafmtCheckAll; + test")
```
SeanCheatham, line 329: Prefixing the command with `+` will run the crossBuild configuration: https://www.scala-sbt.org/1.x/docs/Cross-Build.html#Cross+building+a+project+statefully

scasplte2: ⚡

crypto/src/main/scala/co/topl/crypto/accumulators/merkle/MerkleTree.scala
```diff
@@ -10,9 +10,10 @@ import scala.collection.mutable
 
 /* Forked from https://github.com/input-output-hk/scrypto */
 
+/* NOTE: Use of mutable.WrappedArray.ofByte for Scala 2.12 compatibility */
 class MerkleTree[H, D: Digest](
   topNode:           Option[Node[D]],
-  elementsHashIndex: Map[mutable.ArraySeq[Byte], Int]
+  elementsHashIndex: Map[mutable.WrappedArray.ofByte, Int]
```
scasplte2, line 16: would it be preferable to shadow the 2.13 with a 2.12 specific version instead of using a data type that we'd rather avoid?

SeanCheatham: I think it would be preferred, but the `crypto` module isn't currently setup for the 2.12/2.13 source roots yet.  I also figure it may change again if we move to scodec ByteVectors.

scasplte2 (approved):

Issue #1564, opened 2021-08-24
Newtypes for unsigned values of Short, Int, and Long


b-vennes:
## Description
The binary serializers accept values of `UShort`, `UInt`, and `ULong`, but perform the validation on these values at serialization time. This would be an excellent place to have three newtypes which do some validation on the input values when they are created. This way the serializers don't have to hardcode the actual ranges of these types.

## Acceptance Criteria
* `UShort`, `UInt`, and `ULong` are added as types with validation

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-765) by [Unito](https://www.unito.io)

Issue #1546, opened 2021-08-16
Change VLQ Reader Functions Into Pure Functions


b-vennes:
## Description
All `VLQReader` functions are stateful in that they modify some internal buffer state while parsing a data type. It would be more type-safe and defensive if they were pure functions which took in some existing buffer and output a result of the parsing.

## Acceptance Criteria
* Unit tests are added for reading byte data using the new functions
  * Includes integration tests for existing Bifrost types
* Serializers are backwards compatible with previous versions

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-777) by [Unito](https://www.unito.io)

Issue #1545, opened 2021-08-16
Change VLQ Writer Functions Into Pure Functions


b-vennes:
## Description
All `VLQWriter` functions are stateful in that they modify a builder as functions are called and the bytes are constructed at the end using a `.result()` method.

It would be more typesafe and defensive to use pure functions which take in some existing serialized bytes and returns the new result. It would also be best if a result of type `Either` or `Validated` was returned from this function instead of throwing an error when a failure occurs.

## Acceptance Criteria
* `VLQWriter` is a library of pure functions using immutable objects
* Unit tests are written for all serializer methods
  * Includes integration tests with particular types
* Serialization is backwards compatible for all types with previous serializers

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-778) by [Unito](https://www.unito.io)

PR #1630, opened 2021-09-10
Custom Scodec Codecs For Primitive Types

b-vennes:
## Purpose

- Implement binary serialization as functions that can be composed together to define serialization for complex types
- Perform serialization with less reliance on mutable objects

## Approach

- Implemented the Scodec `Encoder`, `Decoder`, and `Codec` algebras on the following types: `byte`, `byte-array`, `int`, `uint`, `short`, `ushort`, `intString`, `smallString`, `long`, `ulong`, `boolean`, and `option`

## Tests

- backwards compatibility tests between the new codecs and the existing `VLQByteStringReader` and `VLQByteStringWriter` classes
- encoding/decoding compatibility tests

## Closes

- #1545
- #1546
- #1564


common/src/main/scala/co/topl/utils/IdiomaticScalaTransition.scala
```diff
@@ -58,5 +60,22 @@ private[topl] object IdiomaticScalaTransition {
     }
   }
 
-  object implicits extends ToValidatedOps with ToEitherOps with ToTryOps
+  trait AttemptOps[R] {
```
scasplte2, line 63: I think there is a typo in this file for the toTryOps above this line. The implicit def is called toEitherOps and I think it should be toTryOps in keeping with all the others here.

scasplte2: should we also add a toOptionOps? Or are we just converting all options to Either with a specified leftMap?

common/src/main/scala/co/topl/utils/codecs/binary/BooleanCodec.scala
```diff
@@ -0,0 +1,40 @@
...
+object BooleanCodec {
+
+  private val trueByte: Byte = 0x01
+  private val trueByteBitVector: BitVector = BitVector(trueByte)
+
+  private val falseByte: Byte = 0x00
+  private val falseByteBitVector: BitVector = BitVector(falseByte)
+
+  def decode(from: BitVector): Attempt[DecodeResult[Boolean]] =
+    if (from.length < byteSize) Attempt.failure(Err.insufficientBits(byteSize, from.length))
```
scasplte2, line 15: a little confusing what scope `byteSize` is being pulled from

common/src/main/scala/co/topl/utils/codecs/binary/BooleanCodec.scala
```diff
@@ -0,0 +1,40 @@
...
+    else {
+      val (headByteBits, remainder) = from.splitAt(byteSize)
+      Attempt.successful(DecodeResult(headByteBits === trueByteBitVector, remainder))
```
scasplte2, line 18: I'd like this operator `===` addressed when we go over this

common/src/main/scala/co/topl/utils/codecs/binary/BooleanCodec.scala
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+  def encode(value: Boolean): Attempt[BitVector] =
+    if (value) Attempt.successful(trueByteBitVector)
+    else Attempt.successful(falseByteBitVector)
+
+  val codec: Codec[Boolean] = new Codec[Boolean] {
```
scasplte2, line 25: I'm sure we will but also please focus on this Codec[_] higher-kinded type

common/src/main/scala/co/topl/utils/codecs/binary/BooleanCodec.scala
scasplte2: realized that byteSize is just 8, not much we can do to make that more clear other than a comment in every file which is unfeasible.

common/src/main/scala/co/topl/utils/codecs/binary/BytesCodec.scala
```diff
@@ -0,0 +1,31 @@
+package co.topl.utils.codecs.binary
+
+import scodec.{Attempt, Codec, DecodeResult, Err, SizeBound}
+import scodec.bits.BitVector
+
+object BytesCodec {
+
+  def decode(from: BitVector, size: Int): Attempt[DecodeResult[Array[Byte]]] = {
+    val requiredBits = size * byteSize
+    if (from.length < size * byteSize) Attempt.failure(Err.insufficientBits(requiredBits, from.length))
```
scasplte2, line 10: you defined the variable then didn't use it

common/src/main/scala/co/topl/utils/codecs/binary/IntCodec.scala
```diff
@@ -0,0 +1,29 @@
...
+    ULongCodec.decode(from).map(result => result.map(uLongVal => decodeZigZagInt(uLongVal.toInt)))
+
+  def encode(value: Int): Attempt[BitVector] = ULongCodec.encode(encodeZigZagInt(value))
+
+  val codec: Codec[Int] = new Codec[UShort] {
+    override def decode(bits: BitVector): Attempt[DecodeResult[UShort]] = IntCodec.decode(bits)
+
+    override def encode(value: UShort): Attempt[BitVector] = IntCodec.encode(value)
+
+    override def sizeBound: SizeBound = SizeBound.atMost(64)
```
scasplte2, line 19: 64? I thought an Int was 4 bytes (32 bits)?

common/src/main/scala/co/topl/utils/codecs/binary/ULongCodec.scala
```diff
@@ -0,0 +1,69 @@
...
+    var result: Long = 0
+    var iteration = 0
+
+    while (iteration < 10) {
+      val bitPointer = iteration * byteSize
+
+      if (from.length < bitPointer + byteSize)
+        return Attempt.failure(Err.insufficientBits(bitPointer + byteSize, from.length))
+
+      val b = from.slice(bitPointer, bitPointer + byteSize)
```
scasplte2, line 24: I'd really like some more comments explaining the magic in these 15 or so lines. At first glance, why is the iteration limited to 10, what is the iteration * 7 for, what is the first conditional check doing (after inspection that one is easy but a comment would make it obvious)?

common/src/main/scala/co/topl/utils/codecs/binary/ULongCodec.scala
```diff
@@ -0,0 +1,69 @@
...
+    var output = BitVector.empty
+    var runningValue = value
+
+    while (true)
+      if ((runningValue & ~0x7fL) == 0) {
+        output = output ++ BitVector(runningValue.asInstanceOf[Byte])
+        return Attempt.successful(output)
+      } else {
+        output = output ++ BitVector(((runningValue.asInstanceOf[Int] & 0x7f) | 0x80).toByte)
+        runningValue >>>= 7
```
scasplte2, line 47: review should also discuss this operator `>>>=`

common/src/main/scala/co/topl/utils/codecs/binary/SmallStringCodec.scala
```diff
@@ -0,0 +1,59 @@
+package co.topl.utils.codecs.binary
+
+import cats.implicits._
+import scodec.bits.BitVector
+import scodec.{Attempt, Codec, DecodeResult, Decoder, Err, SizeBound}
+
+object SmallStringCodec {
```
scasplte2, line 7: I'm not really a fan of mixing SmallString and IntString. I agree that ByteString was a little confusing but we should either pick a different way to describe strings of varying length, or we should standardize to ByteString, ShortString, IntString, LongString, etc.

common/src/main/scala/co/topl/utils/codecs/binary/SmallStringCodec.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  val maxBytes: Int = 255
+
+  def decode(from: BitVector): Attempt[DecodeResult[SmallString]] =
+    Attempt.fromEither(
+      for {
+        // split input by size bits and remaining
+        sizeSplitTuple <-
+          Either.cond(from.length >= byteSize, from.splitAt(byteSize), Err.insufficientBits(byteSize, from.length))
+        stringSizeBits = sizeSplitTuple._1.toInt(signed = false) * byteSize
```
scasplte2, line 17: should a string have a positive length?

common/src/main/scala/co/topl/utils/IdiomaticScalaTransition.scala
b-vennes: I'll fix the toTryOps, but Option already has `.get` which is basically the same thing

common/src/main/scala/co/topl/utils/codecs/binary/ULongCodec.scala
b-vennes: Unfortunately, I don't have a good explanation for the magic. I just made sure that the same couple of steps are run as before.

common/src/main/scala/co/topl/utils/codecs/binary/SmallStringCodec.scala
b-vennes: Can you explain what you mean?

b-vennes:
Closing branch to continue work on serialization refactor in a separate branch than `dev`. This branch will become the new `serialization-refactor` source of work.

Issue #1467, opened 2021-07-27
MemPoolAuditor and CleanupWorker Refactor


SeanCheatham:
The MemPoolAuditor and CleanupWorker actors work with the NodeViewHolder to eliminate transactions that are no longer valid based on changes to the NodeView.  The coordination and delegation between MemPoolAuditor and CleanupWorker could be simplified into a single actor.  Additionally, there is a bug where transactions with an insufficient fee are not eliminated from the MemPool in a timely manner.  This may also be an opportune time to refactor the MemPoolAuditor into a typed actor.

Acceptance Criteria:

* Simplify/Improve the MemPoolAuditor and CleanupWorker
* Add test cases for MemPool Auditor
* Eliminate transactions which have insufficient fee
* Optional: Convert to typed actor

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-803) by [Unito](https://www.unito.io)

PR #1634, opened 2021-09-13
MemPoolAuditor typed

jingsting:
## Purpose
- `MemPoolAuditor` and `CleanupWorker` need a refactor, there are some procedures that can simplified. 
- Transactions in the mem pool that become invalid due to changes to the state need to be eliminated sooner.

## Approach
- Merged `CleanupWorker` into `MemPoolAuditor`
- Turned `MemPoolAuditor` into a typed actor
- Added a check to eliminate transactions with invalid input box during mem pool cleanup

## Tests
- Tested if transactions that stayed in the mem pool longer than the `mempoolTimeout` will be eliminated
- Tested if transactions with invalid input box will be eliminated during cleanup

## Closes
- #1627
- #1467


node/src/main/scala/co/topl/Heimdall.scala
```diff
@@ -321,10 +321,13 @@ object Heimdall {
       )
     }
 
-    val mempoolAuditor = context.actorOf(
-      MempoolAuditorRef.props(settings, appContext, state.nodeViewHolder, networkController),
-      MempoolAuditor.actorName
-    )
+    val mempoolAuditor = {
+      implicit val timeout: Timeout = Timeout(10.seconds)
```
jingsting, line 325: remove this

scasplte2:
is this ready to be converted @jingsting ?

node/src/main/scala/co/topl/nodeView/MemPoolAuditor.scala
```diff
@@ -1,152 +1,166 @@
...
+
+  /**
+   * Constant which shows on how many cleanup operations (called when a new block arrives) a transaction re-check
+   * happens. If transactions set is large and stable, then about (1/RevisionInterval)-th of the pool is checked
+   */
+  val RevisionInterval: Int = 4
+
+  def idle(validatedTx: TreeSet[Transaction.TX], iteration: Int): Behavior[ReceivableMessage] =
+    Behaviors.receiveMessagePartial[ReceivableMessage] { case ReceivableMessages.RunCleanup =>
+      context.pipeToSelf(withNodeView(splitIds(validatedTx, _))) {
```
scasplte2, line 87: @jingsting I don't think we need to change this now but in the future, when we see these sort of nested function calls, we should strive to make them a for comprehension (at least that is the first thing that comes to mind). Really, I want to avoid this nesting functions. To me that is a red flag that screams to be refactored because nested functions (1) make the code harder to read and reason about and (2) means that you have to have the error handling cases for the inner functions be captured by the outer functions. If you use something like a for-comprehension, or mapping, then the error handling can be embedded into the return type and handled by the calling function.

For this particular case, it doesn't seem much of an issue but I want you to double check that splitIds can't error out in some odd fashion. What are the failure conditions for all of the operations being performed in that function? Can any of them result in an exception? If so then we should change the output type to be an Either.

node/src/main/scala/co/topl/nodeView/MemPoolAuditor.scala
```diff
@@ -1,152 +1,166 @@
...
+        case Success(decision)  => decision
+        case Failure(exception) => ReceivableMessages.Fail(exception)
+      }
+      active(validatedTx, iteration)
+    }
 
-  val actorName = "mempoolAuditor"
+  private def active(validatedTx: TreeSet[Transaction.TX], iteration: Int): Behavior[ReceivableMessage] =
```
scasplte2, line 94: can we rename this to `awaitingDecision`? Since that is the state we are in.

node/src/main/scala/co/topl/nodeView/MemPoolAuditor.scala
```diff
@@ -1,152 +1,166 @@
...
+        val newIteration = if (iteration < Int.MaxValue) iteration + 1 else 0
+        if (newIteration % RevisionInterval == 0) {
+          // drop old index in order to check potentially outdated transactions again.
+          idle(TreeSet.empty ++ validatedTransactions, newIteration)
+        } else {
+          idle(validatedTx ++ validatedTransactions, newIteration)
+        }
+
+      case ReceivableMessages.Fail(throwable) =>
+        throw throwable
```
scasplte2, line 117: How does the typed actor system behave when you throw an error? Is this just logged to the console or does it crash the actor since no behavior is returned?

node/src/main/scala/co/topl/nodeView/MemPoolAuditor.scala
```diff
@@ -1,152 +1,166 @@
...
-////////////////////////////////////////////////////////////////////////////////////
-//////////////////////////////// ACTOR REF HELPER //////////////////////////////////
+    import scala.concurrent.duration._
+    implicit val timeout: Timeout = Timeout(10.seconds)
+    implicit val typedSystem: ActorSystem[_] = context.system
+    nodeViewHolderRef.askWithStatus[T](NodeViewHolder.ReceivableMessages.Read(f, _))
+  }
 
-object MempoolAuditorRef {
+  private def rebroadcastTransactions(transactions: Seq[Transaction.TX]): Unit = {
```
scasplte2, line 155: I think we should update this pattern eventually. Maybe put a todo above this (we can get to it when we revamp the networking) but I'd like to just make this a tell to the nodeview synchronizer what modifiers to broadcast. That way the Mempool Auditor doesn't need to know anything about how to construct a message, since that really seems outside the scope of what the MempoolAuditor should do.

scasplte2 (changes requested):
Please address the comments below and make the todo about the messaging bit. Otherwise, this should be good to finish up shortly.

scasplte2 (commented):
👍

scasplte2 (approved):
👍

PR #1638, opened 2021-09-15
KES MMM Verification

SeanCheatham:
# Purpose
- Creates valid MMM proofs when staking
- Verifies valid MMM proofs when verifying
- Basic key evolving
- Introduces Sized.strictUnsafe as a helper/shortcut
- Minimal Codec and Signable instances for block creation
- Organization and consistency of typeclasses and algebras

# Approach
- `Evolves` typeclass for KesCertificates
- `ContainsVerificationKey` typeclass
- `KeyInitializer` KES instance
- `KesCertifies` typeclass
- Move crypto typeclasses to crypto module
- Move algebra traits to algebras module

# Testing
- Updated header validation tests to create valid KES MMM proofs
- Updated "FullNode" to use new KES implementation


consensus/src/main/scala/co/topl/consensus/LeaderElection.scala
```diff
@@ -1,137 +1,132 @@
...
-  }
+  object Hit {
+
+    object Eval {
+
+      def make[F[_]: Monad](
+        secret:               PrivateKeys.Vrf,
+        thresholdInterpreter: LeaderElectionEligibilityAlgebra[F]
+      ): LeaderElectionHitAlgebra[F] =
+        (relativeStake: Ratio, slot: Slot, slotDiff: Epoch, eta: Eta) =>
```
scasplte2, line 28: slitDiff is an Epoch?

PR #1640, opened 2021-09-15
Transaction builder

### LINKED ISSUE #279: Standardize transfer transaction generation and structure
Release B

b-vennes:
## Purpose

- have a single entry-point for creating any type of transfer transaction
- provide users with the ability to strategize about which boxes should be chosen for a particular transfer transaction

## Approach

- adds `buildTransfer` function which takes in some general details about a transaction along with an implicit blueprint for creating a transaction using a particular value type and expected transfer result type
- `buildTransfer` handles choosing which boxes should be pulled from state and used as inputs to the transfer construction blueprint
- a `ValueTransferBlueprint` type-class determines how to validate and generate a transfer from a set of parameters and boxes
- a `BoxPickingStrategy` type-class determines which boxes that the sending addresses control should be used in a transaction

## Testing

- added a builder spec which tests that the expected boxes are chosen for a transaction

## Closes

- closes #279


common/src/main/scala/co/topl/modifier/transaction/ArbitTransfer.scala
```diff
@@ -56,14 +56,91 @@ object ArbitTransfer {
...
+
+    case object DuplicateArbitInputs extends InvalidArbitTransfer
+
+    case object NoRecipients extends InvalidArbitTransfer
+
+    case object NonUniqueRecipients extends InvalidArbitTransfer
+
+    type ValidationResult[T] = Either[InvalidArbitTransfer, T]
+
+    def validatePolyBoxes(polyBoxes: IndexedSeq[(Address, PolyBox)]): ValidationResult[IndexedSeq[(Address, PolyBox)]] =
```
scasplte2, line 78: I'd like to discuss a way to visualize or build enumerations of the possible states of the inputs? I was just looking at the error cases and thinking there may be a couple more. If you think the visualization would be cumbersome, then we can just discuss.

b-vennes:
This feature needs some more thought, so I'm closing this PR.

Issue #1538, opened 2021-08-13
Improve error response when key not present


scasplte2:
## Summary
Received the following error

    topl@toplnet-00:~$ ./initForging
    {"id":"1","jsonrpc":"2.0","error":{"code":-32098,"message":"Unknown failure","data":{"reason":"UnlockKeyFailureException(akka.pattern.AskTimeoutException: Ask timed out on [Actor[akka://bifrost/user/keyManager#343951459]] after [5000 ms]. Message of type [co.topl.consensus.KeyManager$ReceivableMessages$UnlockKey]. A typical reason for `AskTimeoutException` is that the recipient actor didn't send a reply.)"}}}when attempting to unlock a keyfile that is not in the directory. ## Acceptance
Provide a more informative error to the user when they attempt to unlock a keyfile that is not in their keyfileDir.

Note: I am not too worried about people using this to learn what keyfiles are in the directory since the `admin` namespace is disabled by default and users should be intentional about opening access to these routes when running a node.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-782) by [Unito](https://www.unito.io)

Issue #1532, opened 2021-08-10
Circe decoding failure returns a 500


scasplte2:
We have an issue on 1.7.1 when querying an invalid transactionIdd, the node returns a 500 error instead of a decoding error.

We should write a test to ensure that invalid transaction and block ids return the correct error  to the user

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-787) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://topl.atlassian.net/secure/attachment/10231/948d8b66-e33a-4162-ac36-1e12d17dfb29">948d8b66-e33a-4162-ac36-1e12d17dfb29</a> | <a href="https://topl.atlassian.net/secure/attachment/10232/image.png">image.png</a> | <a href="https://topl.atlassian.net/secure/attachment/10230/Screenshot+2021-09-16+123631.jpg">Screenshot 2021-09-16 123631.jpg</a>

Issue #1530, opened 2021-08-09
Remove LocalTestnet Network Type


SeanCheatham:
The LocalTestnet network type is redundant in our system as we generally just use PrivateTestnet in any circumstances where we need to run a local/independent node.  LocalTestnet can be removed from Bifrost.

### Acceptance Criteria:
* LocalTestnet Network Type no longer exists in Bifrost

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-789) by [Unito](https://www.unito.io)

PR #1643, opened 2021-09-16
Error response fixes and tests

jingsting:
## Purpose
- Make sure useful rpc error responses are returned for unlockKeyFile
- Check and test all the modifier query with invalid ids
- Remove local-testnet since we no longer use it

## Approach
- Added error response when there is no corresponding keyfile found for unlockKeyFile
- Added test for querying modifiers with an id containing non base58 characters
- Removed local-testnet

## Closes
- #1532 
- #1538 
- #1530


common/src/main/scala/co/topl/attestation/keyManagement/KeyRing.scala
```diff
@@ -215,18 +218,20 @@ class KeyRing[
...
-          require(listOfKeyfiles.size == 1, s"Cannot find a unique matching keyfile in $keyDirectory")
-          listOfKeyfiles.head
+    private def checkValid(address: Base58Data, password: Latin1Data): Try[KF] =
+      Try {
+        listKeyFiles()
+          .map {
+            _.filter {
+              _.address == address.decodeAddress.getOrThrow()
+            }
+          } match {
```
scasplte2, line 228: I think you could do a `.toTry` here right?

scasplte2 (changes requested):
looks good, lets go over with everyone and look at the .toTry question. Otherwise this would be good to go I think.

scasplte2:
These are all the changes that you wanted to incorporate right @jingsting ?

jingsting:
> These are all the changes that you wanted to incorporate right @jingsting ?

Yes, this has the changes that we wanted yesterday

scasplte2:
👍 got it, going to merge

scasplte2 (approved):
🔥

PR #1739, opened 2021-10-26
Mongo DB Codecs, Typeclasses, and Models

b-vennes:
## Purpose

- create a package for reading and writing transaction and block data from a Mongo DB collection
- centralize data models, encoding, and decoding methods

## Approach

- create a `DocumentEncoder` and `DocumentDecoder` typeclass which converts a type to and from a Mongo Bson Document
- create data models and sub-models representing Bifrost Blocks and Transactions
- implement document codec typeclasses for the data models


scasplte2 (commented):
I really like how this is organized and setup but could we denote that the DataModels are specific to Dion? I'm not sure how best to approach it but some things (namely blocks) will need a different data model in Tetra, so perhaps we can anticipate that now so we don't break an API in the future. If there is no obvious path (or the correct categorization requires more clarity) we can postpone this until it is needed.

b-vennes:
> I really like how this is organized and setup but could we denote that the DataModels are specific to Dion? I'm not sure how best to approach it but some things (namely blocks) will need a different data model in Tetra, so perhaps we can anticipate that now so we don't break an API in the future. If there is no obvious path (or the correct categorization requires more clarity) we can postpone any further changes to this PR until we actually need it.

This could be solved as simply as creating Dion and Tetra specific models packages, then when users go to create some data, they can choose which package to pull models from.

Such as `mongodb.Dion.BlockDataModel` and `mongodb.Tetra.BlockDataModel`.

PR #1748, opened 2021-11-02
SecureStore

SeanCheatham:
# Purpose
- To support a Key Evolving Scheme, there must me a mechanism to securely erase a private key once it has been used
- This mechanism can be abstracted into an implementation that securely erases any sort of data that has a corresponding binary codec

# Approach
- Added a `SecureStore` algebra that defines methods: `list`, `write`, `consume`, and `erase`
- Added an `AkkaSecureStore` implementation that persists data to some directory on disk and is backed by an actor to manage concurrency

# Testing
- Added unit tests for AkkaSecureStore

# Closes
- I'm not quite ready to close the corresponding ticket yet - there is still follow-up work to integrate the SecureStore with KES


demo/src/main/scala/co/topl/demo/AkkaSecureStore.scala
```diff
@@ -0,0 +1,136 @@
+package co.topl.demo
```
scasplte2, line 1: this is great!, I really like this pattern 🐙

scasplte2 (approved):
🔥 💾

PR #1753, opened 2021-11-03
DionAddress, Propositions Restructure, ByteVector, Key Init from Entropy

SeanCheatham:
# Purpose
- Renamed `Address` to `DionAddress`
- Added `DionAddressable` typeclass
- Restructure the Propositions object
- Use scodec's ByteVector
- Elliptic Curve key initialization from entropy+password

# Approach
- New DionAddressable typeclass, implemented for Curve and Ed
- Change EllipticCurveSignatureScheme#createKeyPair to accept entropy+password+implicit entropyToSeed
- EntropyToSeed helper, accepted as an implicit in the createKeyPair

# Testing
- Unit tests still pass locally
- Unit tests use a "fast" key generator
- Added basic test vectors for the "non-fast" key generators


byte-codecs/src/main/scala/co/topl/codecs/bytes/VLQWriter.scala
```diff
@@ -168,3 +152,26 @@ trait VLQWriter extends Writer {
     this
   }
 }
+
+object VLQWriter {
+
+  def uLongSerializer(x: Long): Array[Byte] = {
```
SeanCheatham, line 158: Moved this to a companion object because creating a "Proposition" of a Treshold signature requires us to preserve the _exact_ encoding scheme from Dion.

common/src/main/scala/co/topl/modifier/block/BloomFilter.scala
```diff
@@ -81,17 +81,17 @@ object BloomFilter extends BifrostSerializer[BloomFilter] {
   private val bitElemMask: Int = 63 /*   63 -> 0000 0000 0011 1111 - mask for finding the bit to modify */
 
   /** Create a new Bloom filter given a sequence of topics to be hashed */
-  def apply(topics: Set[BloomTopic]): BloomFilter = update(empty, topics)
+  def apply(topics: Iterable[BloomTopic]): BloomFilter = update(empty, topics)
```
SeanCheatham, line 84: `Set[BloomTopic]` doesn't result in the expected elements because a `BloomTopic` is a newtype around `Array[Byte]` which does not satisfy proper hashcode or equals expectations in Sets.

crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
```diff
@@ -244,24 +255,22 @@ object ExtendedEd25519 {
...
-    def clampBits(sizedSeed: Sized.Strict[Bytes, SecretKeys.ExtendedEd25519.Length]): SecretKeys.ExtendedEd25519 = {
-      val seed = sizedSeed.data.toArray
+  def fromEntropy(entropy: Entropy)(password: Password = ""): SecretKeys.ExtendedEd25519 =
+    clampBits(entropyToSeed(entropy)(password))
 
-      // turn seed into a valid ExtendedPrivateKeyEd25519 per the SLIP-0023 Icarus spec
-      seed(0) = (seed(0) & 0xf8.toByte).toByte
-      seed(31) = ((seed(31) & 0x1f.toByte) | 0x40.toByte).toByte
+  /** clamp bits to make a valid Bip32-Ed25519 private key */
+  def clampBits(sizedSeed: Sized.Strict[Bytes, SecretKeys.ExtendedEd25519.Length]): SecretKeys.ExtendedEd25519 = {
```
scasplte2, line 262: can we make this private?

crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
```diff
@@ -244,24 +255,22 @@ object ExtendedEd25519 {
...
 
-    clampBits(entropyToSeed(entropy)(password))
+    SecretKeys.ExtendedEd25519(
+      Sized.strictUnsafe(Bytes(seed.slice(0, 32))),
+      Sized.strictUnsafe(Bytes(seed.slice(32, 64))),
+      Sized.strictUnsafe(Bytes(seed.slice(64, 96)))
+    )
   }
 
   /** do a PBDKF2-HMAC-SHA512 per the SLIP2-0023 Icarus spec */
```
scasplte2, line 276: can we remove this from here?

scasplte2 (approved):
👍

PR #1758, opened 2021-11-04
ExtendedEd25519 Fix.  Incorporate ByteVector.

SeanCheatham:
# Purpose
- ExtendedEd25519 unit tests occasionally fail
  - This seems to be the result of an issue in the byte implementation in `ExtendedEd25519.fromEntropy`
- Switched the `Bytes`  type to use scodec's ByteVector

# Approach
- Remove the `.reverse` calls in the byte construction of ExtendedEd25519.fromEntropy
- Update the `MnemonicToExtendedEd25519` tests to use the raw vector test strings
- Include scodec dependency.  Change `type Bytes` to `type Bytes = ByteVector`.  Update any broken references.

# Testing
- Unit tests pass locally with updated test vector strings from the [spec](https://topl.atlassian.net/wiki/spaces/CORE/pages/304023492/HD+Wallet+Protocols+and+Test+Vectors)

### DIFF
--- a/crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
+++ b/crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
@@ -8,20 +8,23 @@ import co.topl.models.utility.Sized
 import org.bouncycastle.crypto.digests.SHA512Digest
 import org.bouncycastle.crypto.macs.HMac
 import org.bouncycastle.crypto.params.KeyParameter
+import scodec.bits.ByteVector
 
 import java.nio.charset.StandardCharsets
 import java.nio.{ByteBuffer, ByteOrder}
 
 class ExtendedEd25519
-    extends eddsa.Ed25519
-    with EllipticCurveSignatureScheme[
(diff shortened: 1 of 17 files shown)


crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
```diff
@@ -8,20 +8,23 @@ import co.topl.models.utility.Sized
...
+    extends EllipticCurveSignatureScheme[
       SecretKeys.ExtendedEd25519,
       VerificationKeys.ExtendedEd25519,
       Proofs.Signature.Ed25519
     ] {
-  override val SignatureLength: Int = SIGNATURE_SIZE
-  override val KeyLength: Int = SECRET_KEY_SIZE
-  val PublicKeyLength: Int = PUBLIC_KEY_SIZE
+
+  private val impl = new eddsa.Ed25519
```
scasplte2, line 23: is there a benefit to this construction over mixing it in?

scasplte2 (approved):

PR #1765, opened 2021-11-11
Common Codecs Package

### LINKED ISSUE #1624: Replace usages of Bifrost Serializers with serde functions
All the binary serde functionality in Bifrost is handled through the `BifrostSerializer` trait. This trait should be replaced or modified to use the binary serde functions that have been implemented as part of the Functional Bytes Serialization/Deserialization epic.

## Acceptance Criteria
* All serialization in a Bifrost node passes through the new serde functions.
* Backwards compatibility is tested between the new functions and the old ones
* The old writers/readers are marked as deprecated

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-945) by [Unito](https://www.unito.io)

b-vennes:
## Purpose

- move all JSON and binary codecs for all types in the common package into one simple-to-import package
- categorize our various usages of binary data in Bifrost
- replace current usages of the `BifrostSerializer` trait with implementations using Scodec's `Codec` typeclass

## Approach

- created a top-level package named `co.topl.codecs` with sub-packages `co.topl.codecs.json` and `co.topl.codecs.binary`
- moved all binary and json serializers into their respective packages
- moved all existing `BifrostSerializer` instances into a `legacy` package within `co.topl.codecs.binary`
- made all typeclass instances and extensions available under one import `co.topl.codecs._`
- adds three new categories of bytes represented as typeclasses: `Persistable`, `Transmittable`, and `BinaryShow`

## Testing

No new tests have been added.

## Closes

- closes #1624


scasplte2:
@b-vennes could I get an update on this? I am unsure how to proceed given that the test are failing here. Are tests passing on `binary_codecs`?

b-vennes:
> @b-vennes could I get an update on this? I am unsure how to proceed given that the test are failing here. Are tests passing on `binary_codecs`?

Oops, let me merge in the latest changes from `dev` into `binary-codecs` and this branch and then get the tests fixed.

scasplte2 (commented):
was walking through the ULong encoder to make sure I understand it and saw this in the Java protobuf example. Do you think it could be worth comparing to the original method?
https://github.com/protocolbuffers/protobuf/blob/a7252bf42df8f0841cf3a0c85fdbf1a5172adecb/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L1085

Just wondering if even a modest optimization to this function could improve the overall performance of the byte codecs since this function is one of the foundational ones for all the other codecs.

other comments so I don't forget
In ULong decode, why did you pick 10 iterations? Should be sufficient to limit it to 8 right? I think the Java version is limited to 8 iterations
I think I figured this one out. As far as I can tell the only reason you might want to allow for 9 bytes is if a negative value is directly encoded by ULong (skipping the Zig-Zag).
If this is correct, I think we should just add a comment about it in the source code and the documentation
In ULong encode, why not just use 0x80 instead of ~0x7F ?
The default string encoding is set to read UTF_8 characters. I don't see it being used anywhere (at least by CTRL-click) but can we talk about making this Latin-1 so that our byte arrays are fixed length?
For unsigned Int and unsigned short, why not just convert them directly to Longs and then encode them with ULong? If they are unsigned then they must be positive values so there is no need to go through Zig-Zag encoding for those
It seems a little confusing to use the typeclasses in different approaches. One instance I stumbled across was Line 196 of State where you explicitly invoke Persistable[Box[_]].persistedBytes instead of b.persistedBytes . Was just looking usages of Persistable and stumbled across so I think it might be worth circling through to try and maintain consistency (for the sake of future us)
Can we resolve the issue for Block and Transaction (with the messageToSign relying on the legacy mechanism) by creating an Unsigned version of them? I think Sean had to do this in Tetra
PeerSpecSerializer still using .bytes
Chat more about usages of .bytes
I really like these deriveEncoder functions in the json package
What are the Transformer instances used for?
JSON attestations commented out?

b-vennes:
> was walking through the ULong encoder to make sure I understand it and saw this in the Java protobuf example. Do you think it could be worth comparing to the original method? https://github.com/protocolbuffers/protobuf/blob/a7252bf42df8f0841cf3a0c85fdbf1a5172adecb/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L1085
> 
> Just wondering if even a modest optimization to this function could improve the overall performance of the byte codecs since this function is one of the foundational ones for all the other codecs.

I think its worth a shot to optimize it if we can. Just about every codec uses the ULong codec in some way or another.

Would it be best to add a ticket to optimize the codec, or should I do it as part of this PR?

b-vennes:
> In ULong decode, why did you pick 10 iterations? Should be sufficient to limit it to 8 right? I think the Java version is limited to 8 iterations
> 
>     * I think I figured this one out. As far as I can tell the only reason you might want to allow for 9 bytes is if a negative value is directly encoded by ULong (skipping the Zig-Zag).
> 
>     * If this is correct, I think we should just add a comment about it in the source code and the documentation

Thank you for figuring this one out, I was staring at it for a bit and I couldn't remember why. It turns out commenting code _can_ be helpful.

b-vennes:
> * In ULong encode, why not just use 0x80 instead of ~0x7F ?

I'm not sure why that would be. The encoder was mostly copied from [VLQWriter.putULong](https://github.com/Topl/Bifrost/blob/b868f946fdf8eb8541016e97ef5262b261357c2e/common/src/main/scala/co/topl/utils/serialization/VLQWriter.scala#L97) and it uses ~0x7f there? This could be part of investigating an optimization for the ulong codec.

b-vennes:
> * The default string encoding is set to read UTF_8 characters. I don't see it being used anywhere (at least by CTRL-click) but can we talk about making this Latin-1 so that our byte arrays are fixed length?

The `stringCharacterSet` constant is used for the `byteStringCodec`, `intStringCodec`, and `latin1DataCodec`. I think this means that any time we go to serialize a string, we will convert it to UTF8 first if it is not already. 

I don't know what impact it would have changing UTF_8 to Latin-1, but I could try it and see if there are any incompatibilities.

b-vennes:
> * For unsigned Int and unsigned short, why not just convert them directly to Longs and then encode them with ULong? If they are unsigned then they must be positive values so there is no need to go through Zig-Zag encoding for those

I think we are currently converting them to `ULong` values and encoding them with the `ULong` codec. With some additional checks to make sure they are within range of the underlying type.

```
  implicit val uIntCodec: Codec[UInt] =
    uLongCodec.exmapc[UInt](uLong =>
      if (uLong >= minUIntValue && uLong <= maxUIntValue) Attempt.successful(uLong)
      else Attempt.failure(Err("UInt value is outside of valid range."))
    )(uInt => Attempt.successful(uInt))

  implicit val uShortCodec: Codec[UShort] =
    uLongCodec.exmapc(uLong =>
      if (uLong >= minUShortValue && uLong <= maxUShortValue)
        Attempt.successful(uLong.toInt)
      else
        Attempt.failure(Err("UShort value is outside of valid range."))
    )(uShort => Attempt.successful(uShort))
```

b-vennes:
> * It seems a little confusing to use the typeclasses in different approaches. One instance I stumbled across was Line 196 of State where you explicitly invoke Persistable[Box[_]].persistedBytes instead of b.persistedBytes . Was just looking usages of Persistable and stumbled across so I think it might be worth circling through to try and maintain consistency (for the sake of future us)

Yeah the `Box[_]` type really gave me some trouble. Ideally we would be able to do `box.persistedBytes`. However, the Scala type checker does not find an instance of `Persistable` for type `Box[_]`, because technically the type of `Box[_]` used in `State` is different than the type of `Box[_]` I used for the instance `implicit val boxPersistable: Persistable[Box[_]] = Persistable.fromCodec`. This is because the `_` in Scala has some funky behaviors with typeclass traits.

b-vennes:
> * PeerSpecSerializer still using .bytes

All the messages in the `node` package are still using the legacy serializers for stuff. My next PR does all the work of fixing `node` to work with Scodec.

b-vennes:
> > was walking through the ULong encoder to make sure I understand it and saw this in the Java protobuf example. Do you think it could be worth comparing to the original method? https://github.com/protocolbuffers/protobuf/blob/a7252bf42df8f0841cf3a0c85fdbf1a5172adecb/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L1085
> > Just wondering if even a modest optimization to this function could improve the overall performance of the byte codecs since this function is one of the foundational ones for all the other codecs.
> 
> I think its worth a shot to optimize it if we can. Just about every codec uses the ULong codec in some way or another.
> 
> Would it be best to add a ticket to optimize the codec, or should I do it as part of this PR?

I see what you mean now about wanting to just pull in the Java code. I will do this as part of this PR.

b-vennes:
> * What are the Transformer instances used for?

The Transformer typeclass is especially helpful for converting codecs of type `List` to codecs of `Seq` or `Array`.

This lets us do stuff like `val someCodec = listCodec(someOtherCodec).as[Seq[SomeOtherType]]`.

scasplte2:
> > * JSON attestations commented out?
> 
> Which file is this?

sorry, in common/src/test/scala/co/topl/serialization/JsonTests.scala line 54-78

scasplte2:
ok, overall I think that satisfies most of my questions. The only concern I have is in the default string encoder. I think I'd like to change this to Latin-1 but we should check if the legacy tests pass with that config and also if we can sync both public networks from genesis.

b-vennes:
To explain some possible confusion with my latest commit: for attestation maps, we have to encode the propositions using the most generic Proposition encoder we have. And when we decode, we always know the specific type we are decoding to (from the proposition type field on a transaction), so we will always ignore the first byte of the signature proposition.

This can be confusing because if you try to encode a `ListMap[PublicKeyCurve25519Proposition, Proof[PublicKeyCurve25519Proposition]]`, then there is no proposition type byte on the proposition field because it uses the specific encoder for `PublicKeyCurve25519Proposition`, which breaks backwards compatibility.

scasplte2:
We should discuss how to approach the Dion and Tetra serialization schemes. I'd like to try and learn from these types of silly mistakes. Perhaps Tetra should be where we do so.

PR #1781, opened 2021-11-29
Health check RPC

### LINKED ISSUE #1423: Healthcheck Endpoint
Cloud services generally include a "healthcheck" HTTP endpoint that indicates the current state of the system (healthy vs unhealthy).  These are usually called by automated systems to ensure that a process is behaving normally.  Failing healthchecks can often be automatically handled via restart.

Healthchecks usually serve two endpoints: liveliness and readiness.  Liveliness indicates if the application is up and running.  Readiness indicates that the application is not only running but is also ready to serve live traffic.  A failing liveliness check indicates total service failure requiring a restart.  A failing readiness check indicates that the application is temporarily unable to serve requests (thus it should temporarily be removed from any load balancers).

Bifrost doesn't necessarily need both.  A single healthcheck may suffice for now

### Acceptance Criteria:
* The `/status` endpoint is updated to return a JSON response
  * The JSON should include a `forging: true|false` value indicating if the node is actively forging

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-812) by [Unito](https://www.unito.io)

jingsting:
## Purpose
We need a way to check the status of the node, whether it's forging or not

## Approach
- Added an RPC method `topl_status` that returns the forging status of the chain

## Tests
- Tested the responses when the forger is stopped and started

## Closes
#1423


topl-rpc/src/main/scala/co/topl/rpc/ToplRpc.scala
```diff
@@ -287,6 +301,17 @@ object ToplRpc {
...
+
+    object NodeStatus {
+
+      /**
+       * Retrieve information about this running node
+       */
+      val rpc: Rpc[Params, Response] = Rpc("topl_nodeStatus")
+
+      case class Params()
+      case class Response(forgingStatus: String)
```
scasplte2, line 313: this isn't the node status, this is the forging status. If we want it to be the node status then I think we need to retrieve more data about the node

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
```diff
@@ -146,7 +160,7 @@ class NodeViewRpcHandlerImpls(
     view:        HistoryReader[Block, BifrostSyncInfo],
     startHeight: Long,
     endHeight:   Long
-  ): List[Block] =
+  ): ToplRpc.NodeView.BlocksByIds.Response =
```
scasplte2, line 163: this should not be `ToplRpc.NodeView.BlockByIds.Reponse` it should be `ToplRpc.NodeView.BlocksInRange.Response`

scasplte2 (changes requested):
please address the previous comments

topl-rpc/src/main/scala/co/topl/rpc/ToplRpc.scala
scasplte2: @jingsting I believe there is still work to be completed here? Once you have addressed it please ping me to let me know it is updated

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
jingsting: Fixed, nice catch!

scasplte2 (approved):
👍

Issue #1565, opened 2021-08-24
Add RPC route to retrieve a segment of chain


scasplte2:
## Summary
It would be convenient to be able to retrieve multiple blocks at once from a node but we should limit the number of blocks that can be requested

## Acceptance
* An RPC should exst that can retrieve a segment of chain
* Errors should be handled if the user specifies an invalid range (not a positive number or too big)
* RPC should limit the maximum number of blocks requested (error if request too large)
* Unit tests should check error conditions and ensure correct behavior given valid parameters

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-764) by [Unito](https://www.unito.io)

PR #1782, opened 2021-11-29
RPC for head info and segment of chain

### LINKED ISSUE #1772: Add RPC method to get the blockId and height of the current best block
## Description:
The current `topl_head` rpc method gives the blockId, height, score, difficulty, and entire block data. We need a convenient method that just gets the blockId and height

## Acceptance Criteria:
* Add the `topl_headInfo` method
* Add test for the new method
* Add the request to the postman collection

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1026) by [Unito](https://www.unito.io)

jingsting:
## Purpose
- We need RRC methods for fetching the id and height of the best block without the other data about the block
- It way be convenient if we can retrieve a segment of the chain between a start height and an end height

## Approach
- Added a `topl_headInfo` RPC method
- Added a `topl_blocksInRange` RPC method

## Tests
- Added a test for `topl_headInfo` route
- Added a test for retrieving blocks with given start height and end height
- Added a test for the error response when start height and end height are incorrect (e.g. start height is greater than end height)
- Added a test for the response of the `topl_info` route

## Closes
#1772 
#1565


node/src/main/scala/co/topl/rpc/handlers/package.scala
```diff
@@ -19,4 +19,17 @@ package object handlers {
...
     } yield keys
 
+  private[handlers] def checkHeightRange(
+    bestBlockHeight: Long,
+    startHeight:     Long,
+    endHeight:       Long
+  ): Either[RpcError, (Long, Long)] =
+    for {
+      _ <- Either.cond(
+        startHeight >= 1 && endHeight >= startHeight && bestBlockHeight >= startHeight,
```
scasplte2, line 29: what happens when bestBlockHeight < endHeight ? Shouldn't the last check be bestBlockHeight >= endHeight?

topl-rpc/src/main/scala/co/topl/rpc/ToplRpcErrors.scala
```diff
@@ -10,6 +10,8 @@ object ToplRpcErrors {
 
   val NoBlockIdsAtHeight: CustomError = CustomError(-32000, "No block ids found from that block height")
 
+  val InvalidHeightRange: CustomError = CustomError(-32009, "Invalid height range")
```
scasplte2, line 13: didn't we also want to add a check that the user is not requesting too many blocks? Like the difference between end - start <=100. I also wonder if we should also add the `blockIdsInRange` so they can get back either a list of blocks of a list of ids

scasplte2 (changes requested):
please address the
- logic for getting blocks compared to best block
- error when the user specifies an end height greater than the current height
- error when the user asks for too many blocks
- add a `blockIdsInRange` method

I would say limit the number of blocks from `blocksInRange` to be 100 (configurable via settings) and the number of block ids in the `blockIdsInRange` to be 1000 (also configurable)

topl-rpc/src/main/scala/co/topl/rpc/ToplRpcErrors.scala
jingsting: That was added in the methodAlias-rpc branch, sorry it got mixed into that one

scasplte2: I just saw that. Thanks for letting me know 👍

node/src/main/scala/co/topl/rpc/handlers/package.scala
jingsting: We will get back blocks from startHeight to bestBlockHeight if bestBlockHeight < endHeight. Do you think we need to enforce getting the exact range of blocks? I could add that condition here

jingsting:
> please address the
> 
> * logic for getting blocks compared to best block
> * error when the user specifies an end height greater than the current height
> * error when the user asks for too many blocks
> * add a `blockIdsInRange` method
> 
> I would say limit the number of blocks from `blocksInRange` to be 100 (configurable via settings) and the number of block ids in the `blockIdsInRange` to be 1000 (also configurable)

- added a check enforcing that bestBlockHeight >= endHeight
- the checks for block/tx number limit were added in a different branch
- `topl_blockIdsInRange` was added in `blockIdsInRange-rpc` branch

scasplte2 (approved):
🔥

PR #1791, opened 2021-12-06
Confirmation Status RPC

jingsting:
## Purpose
Add a confirmation status check to see if transactions are still unconfirmed in the mempool or already included on chain

## Approach
- Added an RPC method `topl_confirmationStatus` that takes an array of transaction ids and returns the confirmation status
- If a transaction is in the mempool, return "Unconfirmed" with a depthFromHead of -1
- If a transaction is already on the chain, return "Confirmed" with the block height difference with the head of the chain as depthFromHead
- Return error if invalid ids are given or if no such ids are found in both mempool and history

## Tests
- Tested getting confirmation status for confirmed and unconfirmed transactions
- Tested getting error response for non existent transaction ids

## Closes
# 1166

### DIFF
--- a/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
+++ b/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
@@ -6,12 +6,14 @@ import cats.implicits._
 import co.topl.akkahttprpc.{CustomError, InvalidParametersError, RpcError, ThrowableData}
 import co.topl.attestation.Address
 import co.topl.consensus.ForgerInterface
+import co.topl.modifier.ModifierId
 import co.topl.modifier.block.Block
 import co.topl.modifier.box._
 import co.topl.network.message.BifrostSyncInfo
 import co.topl.nodeView.history.HistoryReader
 import co.topl.nodeView.state.StateReader
 import co.topl.nodeView.{NodeViewHolderInterface, ReadableNodeView}
+import co.topl.rpc.ToplRpc.NodeView.ConfirmationStatus.TxStatus
 import co.topl.rpc.{ToplRpc, ToplRpcErrors}
 import co.topl.settings.{AppContext, RPCApiSettings}
 import co.topl.utils.Int128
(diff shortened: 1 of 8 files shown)


node/src/main/scala/co/topl/rpc/handlers/package.scala
```diff
@@ -62,4 +64,31 @@ package object handlers {
...
     } yield (startHeight, endHeight)
 
+  private[handlers] def checkTxIds(
+    txIds: List[ModifierId],
+    view:  ReadableNodeView
+  ): Either[RpcError, ToplRpc.NodeView.ConfirmationStatus.Response] = {
+    val txStatusOption: List[Option[(ModifierId, TxStatus)]] = txIds.map { id =>
+      val mempoolStatus = view.memPool.modifierById(id)
+      val historyStatus = view.history.transactionById(id)
+      val bestBlockHeight = view.history.height
```
scasplte2, line 74: you should pull this out of the scope because it doesn't depend on `id`

scasplte2 (changes requested):
see previous comment

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
```diff
@@ -166,6 +175,24 @@ class NodeViewRpcHandlerImpls(
...
       .toList
 
+  private def getConfirmationStatus(
+    txIds: List[ModifierId],
+    view:  ReadableNodeView
+  ): List[Option[(ModifierId, TxStatus)]] =
+    txIds.map { id =>
+      val mempoolStatus = view.memPool.modifierById(id)
+      val historyStatus = view.history.transactionById(id)
+      val bestBlockHeight = view.history.height
```
scasplte2, line 185: the bestBlockHeight should still be moved out

scasplte2: @jingsting please ask me if this is unclear what I am asking

scasplte2 (approved):
💯

PR #1795, opened 2021-12-06
Fix Valhalla Sync Bug

b-vennes:
## Purpose

There is a bug with 1.8.X and 1.9.X nodes which have been unable to sync with Valhalla when starting from a Valhalla data backup.

## Approach

Change the `bestBlockIdKey` from a 32-byte array filled with `-1` to a 33-byte array, which is the current array used in Valhalla and 1.7.X nodes and earlier.

## Testing

No new tests have been added.

### Closes

- #1794


scasplte2 (approved):
👍

PR #1835, opened 2021-12-14
encodeTransfer RPC method

jingsting:
## Purpose
We need a new RPC method for encoding a transfer into the message to sign will allow users to only utilize the serializers in a Bifrost node, and pick the input boxes themselves.

## Approach
Added `topl_encodeTransfer` RPC method that returns a messageToSign string

## Testing
Added tests for the new method in `TransferRPCTestMethods.scala`

## Tickets
#1831


node/src/test/scala/co/topl/api/transaction/TransferRPCTestMethods.scala
```diff
@@ -53,6 +53,27 @@ trait TransferRPCTestMethods extends AnyWordSpec with Matchers with RPCMockState
...
   }
 
+  def testEncodeTransfer(rawTx: Json, messageToSign: String): Unit = {
+    val requestBody = ByteString(s"""
+      |{
+      | "jsonrpc": "2.0",
+      | "id": "2",
+      | "method": "topl_encodeTransfer",
+      | "params": [{
+      |   "rawTx": $rawTx
```
scasplte2, line 63: I would like to change the param to take the key `unprovenTransaction`

I also was thinking about making this take a list of inputs. You could then return a list of `messagesToSign`? Where the list is a tuple of transaction id with the base58 encoding. Do you think something like that would be easy to implement? This way we set ourselves up for handling multiple requests. You would probably want to limit the size of the input array though so perhaps sticking with the single value would be good enough for now. Thoughts?

scasplte2:
also this is the image I was referring to earlier
image

node/src/test/scala/co/topl/api/transaction/TransferRPCTestMethods.scala
jingsting: Yeah, I thought about taking an array here as well. I took just one because the other methods for creating and broadcasting transactions work on just one transaction at a time. Do you imagine people building multiple transactions at once? I think it won't be hard to turn it into an array

jingsting:
- Renamed parameter for `topl_encodeTransfer` to unprovenTransaction
- Changed ForgerStatus to a sealed trait
- Renamed variable "method" in Rpc to methods since we are supporting aliases

PR #1840, opened 2021-12-17
Tetra Operational Certificates

SeanCheatham:
## Purpose
- Tetra Block Headers carry an Operational Certificate containing:
  - A "parent" KesProduct VK
  - A parent's commitment to a child/linear Ed25519 VK
  - A child/linear Ed25519 VK
  - A signature of the block by the child/linear Ed25519 SK
- Creating an Operational Certificate requires a forward-secure KesProduct key that undergoes an update per operational period
  - This KesProduct SK commits to several child linear SKs
  - The child linear SKs are only created for slots in which the staker _could_ be eligible
- A SecureStore stores the "next" KesProduct key, which is consumed/erased when constructing the "next next" key
- Operational Certificates must be validated against the data in a block, the data in an address, and the data in a Registration Box

## Approach
- Rename KeyEvolver to OperationalCertificates
- Leverage SecureStore to consume and write KesProduct keys
- Use the consumed KesProduct key to produce linear child keys
- Use the consumed KesProduct key to update to the next time step before writing the new version to disk
- Update BlockHeaderValidation to verify commitments in the block, address, and registration

## Testing
- Updated existing BlockHeaderValidation tests to reflect new structures
- Implemented new operational certificate tests in BlockHeaderValidation
- Implemented OperationalKeysSpec tests

## Tickets
_* closes #1760_
_* closes #1667_
_* closes #1779_


consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
```diff
@@ -32,181 +33,238 @@ object BlockHeaderValidation {
       relativeStakeInterpreter: VrfRelativeStakeValidationLookupAlgebra[F],
       leaderElection:           LeaderElectionValidationAlgebra[F],
       registrationInterpreter:  RegistrationLookupAlgebra[F]
-    ): F[BlockHeaderValidationAlgebra[F]] = Ref
-      .of[F, Ed25519VRF](Ed25519VRF.precomputed())
-      .map(ref =>
-        new BlockHeaderValidationAlgebra[F] {
+    ): F[BlockHeaderValidationAlgebra[F]] =
+      (
```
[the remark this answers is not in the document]
SeanCheatham: `Ref.of` returns an `F[Ref[T]]` instead of immediately returning the `Ref[T]`, so instantiating the Refs inside the `new Impl` would require a bunch of `.flatMap` calls in the implementation code.

consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
```diff
@@ -32,181 +33,238 @@ object BlockHeaderValidation {
...
+              kesProduct -> kesProduct.verify(
+                header.operationalCertificate.parentSignature,
+                header.operationalCertificate.childVK.bytes.data ++ Bytes(Longs.toByteArray(header.slot)),
+                header.operationalCertificate.parentVK
               )
+            )
+            .map(isValid =>
+              if (isValid) header.asRight[BlockHeaderValidationFailure]
+              else
+                (BlockHeaderValidationFailures.InvalidOperationalParentSignature(
```
scasplte2, line 156: @SeanCheatham could you clarify for me what is happening in this if / else ?

demo/src/main/scala/co/topl/demo/AkkaSecureStore.scala
```diff
@@ -125,7 +125,9 @@ object AkkaSecureStoreActor {
       private[AkkaSecureStoreActor] def run(baseDir: Path): Unit = {
         val path = Paths.get(baseDir.toString, name)
         if (Files.exists(path) && Files.isRegularFile(path)) {
-          replyTo.tell(Bytes(Files.readAllBytes(path)).decoded[A].some)
+          val bytes = Bytes(Files.readAllBytes(path))
+          Files.delete(path)
```
scasplte2, line 129: this is just a "testing" store right? If I recall the disk persistent store does not just `'Files.delete(path)`

models/src/main/scala/co/topl/models/Box.scala
```diff
@@ -23,12 +23,13 @@ object Box {
...
-    // Note: We don't need to worry about these for phase 0
+    /**
+     * @param commitment message: Hash(vrfVK | poolVK), SK: 0th timestep of the KES
+     * @param activationSlot Slot in which this staker can start staking
+     */
     case class TaktikosRegistration(
-      vrfCommitment:    Sized.Strict[Bytes, Lengths.`32`.type],
-      extendedVk:       VerificationKeys.ExtendedEd25519,
-      registrationSlot: Slot
+      commitment:     Proofs.Knowledge.KesProduct,
```
scasplte2, line 31: is this a bug?

models/src/main/scala/co/topl/models/SecretKey.scala
```diff
@@ -38,20 +38,12 @@ object SecretKeys {
...
-    ///type Length = ???
-  }
-
   case class KesProduct(
-    superTree:    KesBinaryTree,
-    subTree:      KesBinaryTree,
-    nextSubSeed:  Sized.Strict[Bytes, KesProduct.SeedLength],
+    superTree:    KesBinaryTree, // Hour hand
+    subTree:      KesBinaryTree, // Minute hand
+    nextSubSeed:  Array[Byte],
```
scasplte2, line 44: why change this to be an array instead of fixing it's size?

demo/src/main/scala/co/topl/demo/AkkaSecureStore.scala
SeanCheatham: This is the actual store, but you're right that my change here isn't entirely correct.  I'll update the PR to call `erase` after this `run(...)` method returns. (`erase` overwrites the bytes before deleting the file)

models/src/main/scala/co/topl/models/Box.scala
SeanCheatham: I don't think it's a bug.  I believe it should just be a KesProduct signature generated by the 0th-timestep of the staker's ParentSK when they first register.

models/src/main/scala/co/topl/models/SecretKey.scala
SeanCheatham: I think this change was necessary to properly perform the byte erasing when calling `update`.  Before, it was just using the Sized.Strict and then calling `.toArray` which makes a new copy of the byte array (but preserves the `Sized.Strict` copy from before)

consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
SeanCheatham: I'll add some inline docs.  The lines just prior to this call `kesProduct.verify(...)` which returns a Boolean.  These lines highlighted here just decide what to do with the true/false result.

models/src/main/scala/co/topl/models/SecretKey.scala
scasplte2: got it 👍

models/src/main/scala/co/topl/models/Box.scala
scasplte2: ohhh... it is a proof not the msg itself, mea culpa

scasplte2 (approved):
👍

PR #1844, opened 2021-12-21
Added latestBlocks, latestBlockIds methods and tests

jingsting:
## Purpose
We could benefit from RPC methods that retrieve a number of blocks and block ids at the head of the chain

## Approach
- Added RPC method `topl_latestBlocks`
- Added RPC method `topl_latestBlockIds`

## Testing
- Added test for both methods

## Tickets
#1843

### DIFF
--- a/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
+++ b/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
@@ -91,36 +91,52 @@ class NodeViewRpcHandlerImpls(
 
   override val blocksInRange: ToplRpc.NodeView.BlocksInRange.rpc.ServerHandler =
     params =>
-      withNodeView(view =>
-        checkHeightRange(view.history.height, params.startHeight, params.endHeight, rpcSettings.blockRetrievalLimit)
-          .map(range => getBlocksInRange(view.history, range._1, range._2))
-      ).subflatMap(identity)
+      for {
+        headHeight <- withNodeView(_.history.height)
+        range <- EitherT.fromEither[Future](
+          checkHeightRange(headHeight, params.startHeight, params.endHeight, rpcSettings.blockRetrievalLimit)
+        )
+        blocks <- withNodeView(view => getBlocksInRange(view.history, range._1, range._2))
+      } yield blocks
 
   override val blockIdsInRange: ToplRpc.NodeView.BlockIdsInRange.rpc.ServerHandler =
     params =>
-      withNodeView(view =>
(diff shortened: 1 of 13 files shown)


jingsting:
- Separated some handler checks for height into more specific ones with error response
- Changed some of the RPC method implementations to use for comprehension and moved some procedures out from the inside of withNodeView method calls
- Added RPC error tests for securityRoot and assetCode

node/src/test/scala/co/topl/api/transaction/TransferRPCTestMethods.scala
```diff
@@ -201,7 +202,7 @@ trait TransferRPCTestMethods extends AnyWordSpec with Matchers with RPCMockState
       |      "assetCode" : "${assetCode.toString}",
```
b-vennes, line 202: The `.toString` can be removed now that asset code is a `String`.

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
```diff
@@ -78,28 +80,54 @@ class NodeViewRpcHandlerImpls(
...
+          checkHeightRange(headHeight, params.startHeight, params.endHeight, rpcSettings.blockIdRetrievalLimit)
+        )
+        ids <- withNodeView(view => getBlockIdsInRange(view.history, range._1, range._2))
+      } yield ids
+
+  override val latestBlocks: ToplRpc.NodeView.LatestBlocks.rpc.ServerHandler =
+    params =>
+      for {
+        headHeight <- withNodeView(_.history.height)
+        startHeight = headHeight - params.numberOfBlocks + 1
```
b-vennes, line 114: It might be nice to add a comment here for why the `+ 1` is necessary.

scasplte2: were you planning to add this comment @jingsting ?

node/src/test/scala/co/topl/api/transaction/TransferRPCTestMethods.scala
scasplte2: please address @jingsting

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
jingsting: Yeah, just added a comment there

node/src/test/scala/co/topl/api/transaction/TransferRPCTestMethods.scala
jingsting: Removed it 👍

PR #1846, opened 2021-12-22
Kes private key test

aaronschutza:
## Purpose
Update KES Spec to test the private key configuration byte for byte
## Approach
Private key configuration has been added to test vectors in readme files of reference crypto
## Testing
Two tests were added that generate the correct private key at the given time step
## Tickets
1755


models/src/main/scala/co/topl/models/utility/KesBinaryTree.scala
```diff
@@ -1,18 +1,88 @@
...
 object KesBinaryTree {
 
+  object PrivateKeyConstructor {
+    // args has a boolean for each entry corresponding to false <=> left/zero and true <=> right/one,
+    // false-true corresponds to left or right child node with respect to the parent being constructed
+    // witness elements and seeds should be provided for each level of the tree in order of highest to lowest
+    // sk and vk are the leaf (lowest) level private and public keys
+    type Args = (Boolean, (Bytes, Bytes, Bytes))
+
+    def build(sk: Bytes, vk: Bytes, args: Args*): KesBinaryTree =
```
b-vennes, line 16: The build function will fail with a stack overflow if provided with >3000 arguments. If these trees don't get that large then this should be fine.

models/src/main/scala/co/topl/models/utility/KesBinaryTree.scala
```diff
@@ -1,18 +1,88 @@
...
+
   case class MerkleNode(
     seed:         Array[Byte],
     witnessLeft:  Array[Byte],
     witnessRight: Array[Byte],
     left:         KesBinaryTree,
     right:        KesBinaryTree
-  ) extends KesBinaryTree
+  ) extends KesBinaryTree {
+    override def equals(obj: Any): Boolean = areEqual(this, obj)
```
scasplte2, line 80: I don't think we should override this method as it goes against the idea of the models package. We should explicitly call the `areEqual` function or implement an Eq instance

models/src/main/scala/co/topl/models/utility/KesBinaryTree.scala
scasplte2: @aaronschutza is this only for testing? If so then we should move this into the testing scope to avoid this stack overflow issue (in testing I am less worried that someone will use it incorrectly without realizing). In production, I think this could compile fine and result in a runtime error which is something I would like to minimize the likelihood of.

scasplte2:
@aaronschutza also please address the failing tests

Issue #1192, opened 2021-04-07
Return nothing after updating rewards address


b-vennes:
## Description
The API call to UpdateRewardsAddress currently returns the update address. This address is the same value that the user provides. Instead of returning the address, just return a confirmation that the address was updated.

## Acceptance Criteria
The UpdateRewardsAddress API endpoint only returns a confirmation that the update was successful.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-852) by [Unito](https://www.unito.io)

Issue #1032, opened 2021-02-11
Return error if expected modifier does not match the type asked for


scasplte2:
If a block is expected and a transaction id is given, I want to respond that the given modifier is not the correct type

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-829) by [Unito](https://www.unito.io)

Issue #982, opened 2021-01-29
[BUG] - Error messaging needed for invalid body in payload if any element is null or undefined


raularagonez:
API Call to Bifrost, tested using Private node. If any key in the body payload is undefined or no value is passed at all the system throws a null pointer exception. The only message to the user is "There was an internal server error."
This should be handled gracefully and an error should be thrown back stating the body/payload is invalid.
`
{
	"jsonrpc": "2.0",
	"id": "1", 
	"method": "topl_rawAssetTransfer", 
	"params": [
        {
            "propositionType": "PublicKeyCurve25519",
            "recipients": [
                    [
                        "AUAftQsaga8DjVfVvq7DK14fm5HvGEDdVLZwexZZvoP7oWkWCLoE", {
                            "type": "Asset",
                            "quantity":undefined,
                            "assetCode": "4Y7EsNHVwiZ488s2uvePrtNBpCFAsK132H7AUq2rxsBkJSJv7oda9yyZgb2"
                        }
                    ]
                ],
            "sender": ["AUAftQsaga8DjVfVvq7DK14fm5HvGEDdVLZwexZZvoP7oWkWCLoE"],
            "changeAddress": "AUAftQsaga8DjVfVvq7DK14fm5HvGEDdVLZwexZZvoP7oWkWCLoE",
            "minting": true,
            "fee": 1
        }
	]
}
`

`05:51:12.046| ERROR [bifrost-client-akka.actor.default-dispatcher-11] akka.actor.ActorSystemImpl - Error during processing of request: 'java.lang.NullPointerException (No error message supplied)'. Completing with 500 Internal Server Error response. To change default exception handling behavior, provide a custom ExceptionHandler.
java.lang.NullPointerException: null
	at co.topl.http.api.ErrorResponse.<init>(ApiResponse.scala:29)
	at co.topl.http.HttpService.$anonfun$basicRoute$3(HttpService.scala:49)
	at co.topl.http.HttpService.$anonfun$withAuth$1(HttpService.scala:92)
	at akka.http.scaladsl.server.util.ApplyConverterInstances$$anon$1.$anonfun$apply$1(ApplyConverterInstances.scala:14)
	at akka.http.scaladsl.server.directives.BasicDirectives.$anonfun$textract$2(BasicDirectives.scala:161)
	at akka.http.scaladsl.server.directives.BasicDirectives.$anonfun$mapRouteResult$2(BasicDirectives.scala:68)
[...]

Issue #864, opened 2020-12-03
API should send an informative error when the required params are not provided


scasplte2:
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-916) by [Unito](https://www.unito.io)

jingsting:
Already done

PR #1847, opened 2021-12-23
Updated error responses and tests for RPC methods

jingsting:
## Purpose
Added more error responses for RPC methods

## Approach
- For `topl_confirmationStatus`: return "Not Found" when a transaction is not found in mempool and history, instead of returning an error
- For `admin_updateRewardsAddress`: return a confirmation instead of the requested address

## Testing
- Added test for invalid securityRoot(not base58 or invalid length)
- Added test for invalid assetCode
- Updated RPC spec for `topl_confirmationStatus`
- Added test for unlocking a non-existent keyfile with `admin_unlockKeyfile`
- Added test for the response from `admin_updateRewardsAddress `

## Tickets
#864 
#982 
#1032 
#1192 
#1040


topl-rpc/src/main/scala/co/topl/rpc/ToplRpcCodecs.scala
```diff
@@ -119,6 +125,8 @@ trait TransactionRpcParamsEncoders extends SharedCodecs {
   implicit val transactionBroadcastTxParamsEncoder: Encoder[ToplRpc.Transaction.BroadcastTx.Params] =
     Encoder.forProduct1("tx")(_.tx)
 
+  implicit val transactionEncodeTransferParamsEncoder: Encoder[ToplRpc.Transaction.EncodeTransfer.Params] =
+    Encoder.forProduct1("unprovenTransaction")(_.unprovenTransaction)
```
b-vennes, line 129: Are we able to use `deriveEncoder` instead of `Encoder.forProduct1`?

node/src/test/scala/co/topl/api/AdminRPCSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+        """.stripMargin)
+
+      httpPOST(requestBody) ~> route ~> check {
+        val res: Json = parse(responseAs[String]).value
+        val error = res.hcursor.downField("error").as[Json].toString
+        error should include("Cannot find a unique matching keyfile")
+      }
+    }
+
+    "Return return a confirmation after successfully updating the reward address" in {
```
b-vennes, line 33: Possible typo here with `Return return`

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
```diff
@@ -157,17 +156,46 @@ class NodeViewRPCSpec extends AnyWordSpec with Matchers with RPCMockState with E
       view().mempool.remove(txs.head)
     }
 
+    "Return correct error response when an incorrect modifier id type is used" in {
```
b-vennes, line 159: It might be good to break this out into multiple tests, since you are testing three different API endpoints

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
```diff
@@ -157,17 +156,46 @@ class NodeViewRPCSpec extends AnyWordSpec with Matchers with RPCMockState with E
...
+        }
+      }
+
+      httpPOST(requestBody("blockId", "topl_blockById", txId)) ~> route ~> check {
+        val res: String = parse(responseAs[String]).value.hcursor.downField("error").as[Json].toString
+        res should include("The requested id's type is not an id type for Block")
+      }
+    }
+
     "Return correct error response when an id with non-base58 character is used for querying transactions" in {
```
b-vennes, line 188: We might also want to consider breaking this one out into multiple tests with each RPC function and id type.

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
b-vennes: I see we are doing this a lot inside this Spec, it might be better to make it an issue instead of trying to fix it all here

b-vennes (commented):
The changes to the `NodeViewRpcHandlerImpls` looks good. I like the new user-friendly error messages. I have a few comments on the added Json codes and tests.

node/src/test/scala/co/topl/api/AdminRPCSpec.scala
jingsting: Fixed that 😂

topl-rpc/src/main/scala/co/topl/rpc/ToplRpcCodecs.scala
jingsting: I just tried it but couldn't compile with this: could not find Lazy implicit value

scasplte2:
Please make a ticket about removing the RpcClient and RpcServer @jingsting

node/src/main/scala/co/topl/rpc/handlers/AdminRpcHandlerImpls.scala
```diff
@@ -71,7 +71,7 @@ class AdminRpcHandlerImpls(forgerInterface: ForgerInterface, keyManagerInterface
       keyManagerInterface
         .updateRewardsAddress(params.address)
         .leftMap(e => ToplRpcErrors.genericFailure(e.toString): RpcError)
-        .map(_ => ToplRpc.Admin.UpdateRewardsAddress.Response(params.address.toString))
+        .map(_ => ToplRpc.Admin.UpdateRewardsAddress.Response("Updated reward address"))
```
scasplte2, line 74: why did you remove the address from this message? I think a user would be more comfortable if the response was `Updated reward address to ${params.address}`

node/src/test/scala/co/topl/api/AdminRPCSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+        |     "address": "${keyRingCurve25519.addresses.head.toString}",
+        |     "password": "test"
+        |   }]
+        |}
+        """.stripMargin)
+
+      httpPOST(requestBody) ~> route ~> check {
+        val res: Json = parse(responseAs[String]).value
+        val error = res.hcursor.downField("error").as[Json].toString
+        error should include("Cannot find a unique matching keyfile")
```
scasplte2, line 29: where is this error defined?

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
scasplte2: I agree with Braden, we should break these into individual tests so that a unit test checks one thing (unless there is a good reason not to)

node/src/main/scala/co/topl/rpc/handlers/AdminRpcHandlerImpls.scala
jingsting: Just add  ${params.address} back in

node/src/test/scala/co/topl/api/AdminRPCSpec.scala
jingsting: It's in the `checkValid` method in KeyRing.scala. It gets thrown when the number of keyfile found with the address is not exactly one.

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
jingsting: These API endpoints are being used together here because this test wants the error response to be the same when the id type is incorrect. But I'm ok with separating them into individual test too

jingsting: Yeah, that makes sense to me

node/src/test/scala/co/topl/api/NodeViewRPCSpec.scala
jingsting: 👍  Separated these tests

PR #1854, opened 2022-01-05
Genus Filters To Mongo Bson

b-vennes:
## Purpose
* the custom Genus filtering types must be converted into Mongo Bson data to send to MongoDB during queries and subscriptions 

## Approach
* adds instances of the `MongoFilter` typeclass for `BlockFilter` and `TransactionFilter`
* changes `MongoFilter.toFilter` function to `MongoFilter.toBsonFilter` for better specificity

## Testing
* tests that `BlockFilter` and `TransactionFilter` types can be converted into non-empty Bson data without issue

## Tickets
* #1745


genus/src/main/scala/co/topl/genus/typeclasses/MongoFilterInstances.scala
```diff
@@ -1,10 +1,142 @@
 package co.topl.genus.typeclasses
```
scasplte2, line 1: is it possible to "extend" a package across several different modules? Just wondering if we should standardize putting TCs into co.topl.typeclasses but not sure if that is possible or even a good idea.

scasplte2 (commented):
👍 ready to merge in if you are all set @b-vennes

genus/src/main/scala/co/topl/genus/typeclasses/MongoFilterInstances.scala
b-vennes: I've thought about this too, and it seems like it should be possible, but I've failed at doing it each time I try. @SeanCheatham may know how to

PR #1855, opened 2022-01-05
Tetra VRF Simplification

SeanCheatham:
## Purpose
- The VRF mechanism can be simplified from two proofs into a single proof
- This leads to less data in the header and faster computations (signing is more costly than hashing)
- BlockHeaderV2.Unsigned now includes a "PartialOperationalCertificate"
- New codecs for new data types
- Added a `Stats[F[_]]` algebra for collecting arbitrary metrics during runtime

## Approach
- Create newtypes for Rho, RhoTestHash, and RhoNonceHash
- Update LeaderElection process to accept a Rho and turn it into a RhoTestHash
- Update EtaCalculation process to accept Rho values and turn them into RhoNonceHashes

## Testing
- Modified existing test cases to reflect changes

## Tickets
- _closes #1853_


minting/src/main/scala/co/topl/minting/BlockMint.scala
```diff
@@ -19,31 +21,44 @@ object BlockMint {
...
+                      address = address
+                    ),
+                    transactions
+                  )
             )
-            .flatMap(staker.certifyBlock)
+            .flatMap(staker.certifyBlock(parent.slotId, slot, _))
+        )
+        .semiflatTap(block =>
+          stats.write(
```
scasplte2, line 53: is this something you want to leave in for the time being so we can continue to collect data?

minting/src/main/scala/co/topl/minting/OperationalKeys.scala
```diff
@@ -125,66 +138,75 @@ object OperationalKeys {
...
       etaCalculation:          EtaCalculationAlgebra[F]
     )(implicit
-      kesProduct: KesProduct,
-      ed25519:    Ed25519
+      kesProductScheme: KesProduct,
+      ed25519Scheme:    Ed25519
     ): F[LongMap[OperationalKeyOut]] =
       for {
         epoch <- clock.epochOf(fromSlot)
         eta   <- etaCalculation.etaToBe(parentSlotId, fromSlot)
```
scasplte2, line 179: this seems a little odd. I understand why the eta is needed and therefore why the parent slot id has to be passed but at first glance I could not tell why operational keys needed parent slot. Thinking through how this would play out I think we may be alright but perhaps we should discuss just to be sure.

1. epoch boundary before parentSlot which is before fromSlot -> stable eta value
2. epoch boundary after parentSlot and fromSlot -> again, stable eta value
3. epoch boundary between parentSlot and fromSlot -> etaToBe probably handles this correctly 

I think we might also have made an important assumption in our code that we may need to enforce at startup. I think the epoch needs to contain an integer number of operational periods. Otherwise your eta would change midway through the operational period which is difficult to account for. But if (epochLength % operationalPeriodLength == 0) I think we avoid any issues there. This should be a config parameter though.

minting/src/main/scala/co/topl/minting/VrfProof.scala
```diff
@@ -50,41 +51,33 @@ object VrfProof {
...
                         )
-                        val rhoValues = LongMap.from(testProofs.view.mapValues(ed.proofToHash))
-                        ed -> (testProofs -> rhoValues)
+                        val rhoValues = LongMap.from(vrfProofs.view.mapValues(ed.proofToHash))
+                        ed -> (vrfProofs -> rhoValues)
                       }
                     )
-                    .flatTap { case (testProofsForEta, rhosForEta) =>
-                      (testProofs.put(eta)(testProofsForEta), rhos.put(eta)(rhosForEta)).tupled
+                    .flatTap { case (testProofsForEta, nonceRhoVaues) =>
```
scasplte2, line 66: I think this two variables are confusingly named, the `testProofsForEta` are the `vrfSigs` (or Proofs) and the `rhoValues` for that epoch right?

scasplte2:
also a couple of things that were added in the PR that aren't in the description
```
- modified unsigned block to include partial operational certificate
- defined codecs for partial operational certificates
- added minimal logging ability to allow for the collection of block production data
```

minting/src/main/scala/co/topl/minting/BlockMint.scala
SeanCheatham: I think the `Stats[F[_]]` algebra could come in handy for collecting various metrics.  We probably don't need it right now so it could be removed (or at least this specific invocation could be removed).  Or we could leave it and add other places for collecting metrics.  I have a feeling we'll come up with more use-cases pretty quickly.

minting/src/main/scala/co/topl/minting/OperationalKeys.scala
SeanCheatham: Needing eta and parentSlotId is definitely confusing at first.  I'll add some more inline docs explaining why.

As for the assumption in the code - yes, the epochLength needs to be divisible by operationalPeriodLength.  I know we considered this when coming up with `180` for the operationalPeriodLength, but it's not enforced in code.  I can add that in on startup though.

minting/src/main/scala/co/topl/minting/BlockMint.scala
scasplte2: I think that is reasonable, will definitely want to get some data on TPS at some point

consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
```diff
@@ -162,18 +149,24 @@ object BlockHeaderValidation {
...
                     header.operationalCertificate.childVK
                   )
                 )
                 .map(isValid =>
-                  if (isValid) header.asRight[BlockHeaderValidationFailure]
-                  else
+                  // Verification from the previous step returns a boolean, so now check the boolean verification result
+                  if (isValid) {
+                    // If the verification was valid, just return Right(header)
+                    header.asRight[BlockHeaderValidationFailure]
```
scasplte2, line 163: oh, I get it now, I was originally confused because I saw the .asRight[...Failure] and thought the error case was going to the right. 😵‍💫

PR #1870, opened 2022-01-12
Genus Query Interpreters

### LINKED ISSUE #1859: Genus - Query Service Interpreter
Implement the Query Service Interpreter object as an interpreter for the Query Service Algebra.

#### Acceptance Criteria
* queries can be made to the Genus server with expected response data
* queries will timeout if they take too long
* filters can be added to queries
* unit tests are added with 80% coverage

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1044) by [Unito](https://www.unito.io)

b-vennes:
## Purpose
- provide implementations for the `TransactionQuery` and `BlockQuery` interfaces generated by Akka gRPC
  - these interfaces are adjacent to the concept of an algebra, where the return functor types are always of `Future`
- pluralize usages of `Transaction` and `Block` in service, request, and response names

## Approach
- implement `TransactionQueryService` and `BlockQueryService` as interpreters for their respective algebras/interfaces
- use the `DatabaseClientAlg` to retrieve data from the data store
- use `MonadThrow` functor typeclass to catch errors that might occur, TimeoutExceptions in particular with long-running queries
- use a `ToFuture` functor typeclass to convert a given functor to `Future` type for mapping from Cats-Effect to Akka's use of `Future`

## Testing
- adds unit tests for the BlockQueryService and TransactionQueryService
  - tests that a query with no existing data returns a success result with no values
  - tests that a query with data returns a success result with data
  - tests that a query which takes longer than the configured timeout will return a timeout failure result
  - tests that an error with the database connection returns a database failure result 

## Tickets
- closes #1859


genus/src/main/protobuf/blocks_subscription.proto
```diff
@@ -0,0 +1,50 @@
...
+import "filters.proto";
+import "types.proto";
+
+package co.topl.genus.services;
+
+service BlocksSubscription {
+  rpc Create (CreateBlocksSubscriptionReq) returns (CreateBlocksSubscriptionRes);
+  rpc Delete (DeleteBlocksSubscriptionReq) returns (DeleteBlocksSubscriptionRes);
+  rpc Read (stream ReadBlocksSubscriptionReq) returns (stream ReadBlocksSubscriptionRes);
+}
```
scasplte2, line 12: couple of questions on this subscription
- is the point of this for the subscriber to just receive the blocks as they are added to the head of the local chain? If that is the case how are you planning to announce rollbacks to the subscriber? (perhaps this is why you included a depth but we likely still need a way to inform the subscriber because I could see a rollback triggering a flurry of updates as each "new" head is adopted when a tine is switched to.

- what is the purpose of the checkpoint option under read? I assume it is how the client updates the server to say "hey, I got data up to this point" but that feel more like an Update option to the subscription instead of a Read message option

genus/src/main/protobuf/blocks_query.proto
```diff
@@ -0,0 +1,33 @@
...
+
+package co.topl.genus.services;
+
+service BlocksQuery {
+  rpc Query (QueryBlocksReq) returns (QueryBlocksRes);
+}
+
+message QueryBlocksReq {
+  BlockFilter filter = 1;
+  uint32 confirmation_depth = 2;
```
scasplte2, line 14: What does confirmation depth mean in terms of a block query?

genus/src/main/protobuf/transactions_query.proto
```diff
@@ -0,0 +1,33 @@
...
+
+package co.topl.genus.services;
+
+service TransactionsQuery {
+  rpc Query (QueryTxsReq) returns (QueryTxsRes);
+}
+
+message QueryTxsReq {
+  TransactionFilter filter = 1;
+  uint32 confirmation_depth = 2;
```
scasplte2, line 14: Same question as for the block query

genus/src/main/scala/co/topl/genus/GenusApp.scala
```diff
@@ -19,11 +30,36 @@ object GenusApp extends IOApp.Simple {
...
+
+  val databaseClient: DatabaseClientAlg[IO, Source[*, NotUsed]] =
+    MongoDatabaseClient.Eval.make(
+      transactionsQuery,
+      blocksQuery,
+      MongoSubscription.Mock.make,
+      MongoSubscription.Mock.make
+    )
+
+  val server: HttpServer[IO] =
```
scasplte2, line 57: really cool app, complex, but I like this alot

genus/src/main/scala/co/topl/genus/interpreters/MongoDatabaseClient.scala
```diff
@@ -1,29 +1,26 @@
...
-      transactionQuery: QueryAlg[F, Source[*, NotUsed], Bson, ConfirmedTransactionDataModel],
-      blockQuery:       QueryAlg[F, Source[*, NotUsed], Bson, BlockDataModel],
-      transactionSubscription: SubscriptionAlg[F, Source[*, NotUsed], Bson, String, ChangeStreamDocument[
-        ConfirmedTransactionDataModel
-      ]],
-      blockSubscription: SubscriptionAlg[F, Source[*, NotUsed], Bson, String, ChangeStreamDocument[BlockDataModel]]
+      transactionQuery:        MongoQueryAlg[F, ConfirmedTransactionDataModel],
+      blockQuery:              MongoQueryAlg[F, BlockDataModel],
+      transactionSubscription: MongoSubscriptionAlg[F, ConfirmedTransactionDataModel],
+      blockSubscription:       MongoSubscriptionAlg[F, BlockDataModel]
```
scasplte2, line 23: I know this is basically the same as before but it is much easier to read

genus/src/main/scala/co/topl/genus/interpreters/queryservices/BlocksQueryService.scala
```diff
@@ -0,0 +1,106 @@
...
+import co.topl.genus.services.blocks_query.{BlocksQuery, QueryBlocksReq, QueryBlocksRes}
+import co.topl.genus.typeclasses.implicits._
+import co.topl.genus.types.Block
+
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+object BlocksQueryService {
+
+  object Eval {
```
scasplte2, line 21: good commenting! 🎉

scasplte2 (approved):
I would definitely like @SeanCheatham to go over this as well because admittedly this code is a bit over my head to see potential pitfalls but overall I think this looks great.

No need to add to this PR but came to mind while reviewing, I would like to have the option to protect the HTTP server via an API key.

genus/src/main/protobuf/blocks_query.proto
b-vennes: In terms of queries, I think confirmation depth would mean "give me all the data matching my filters that occurs in the blocks up to (the head block - the confirmation depth). This way the query will not return results which are in an unstable area of the chain.

b-vennes:
> I would definitely like @SeanCheatham to go over this as well because admittedly this code is a bit over my head to see potential pitfalls but overall I think this looks great.
> 
> No need to add to this PR but came to mind while reviewing, I would like to have the option to protect the HTTP server via an API key.

I agree. I haven't started on any implementation for a key, but I would imagine we just need to add a `authToken` parameter to every request type. Then we can just check the auth token against our auth cache whenever we need to see if a user can perform actions.

genus/src/main/protobuf/blocks_subscription.proto
b-vennes: I think users should be able to provide a confirmation depth, which means start subscribing to any new blocks and only send them once they have reached a certain confirmation depth. This will make it very likely that any blocks won't have rollback issues. I'm not sure exactly how to handle rollbacks yet, but I'll do some research at some point into handling data rollbacks in an event sourced environment.

b-vennes: > > is the point of this for the subscriber to just receive the blocks as they are added to the head of the local chain? If that is the case how are you planning to announce rollbacks to the subscriber? (perhaps this is why you included a depth but we likely still need a way to inform the subscriber because I could see a rollback triggering a flurry of updates as each "new" head is adopted when a tine is switched to.
> 
> I think users should be able to provide a confirmation depth, which means start subscribing to any new blocks and only send them once they have reached a certain confirmation depth. This will make it very likely that any blocks won't have rollback issues. I'm not sure exactly how to handle rollbacks yet, but I'll do some research at some point into handling data rollbacks in an event sourced environment.
> 
> > what is the purpose of the checkpoint option under read? I assume it is how the client updates the server to say "hey, I got data up to this point" but that feel more like an Update option to the subscription instead of a Read message option
> 
> Since the read request creates a two-way stream of n -> n messages, clients can send checkpoints as often or as sparse as they want. It could be an update request too, and it might need to change depending on how easy it is to implement or how easy it is for clients to handle sending updates and receiving data.

If a client wants to update their checkpoint after every tx or block, it might get slow to re-open an http connection after every request instead of using an existing connection.

scasplte2: Reasonable, makes sense as part of the stream. I think the depth on blocks is a good first step as well

genus/src/test/scala/co/topl/genus/interpreters/queryservices/BlocksQueryServiceSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.scalamock.scalatest.AsyncMockFactory
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.duration.DurationInt
+
+class BlocksQueryServiceSpec extends AsyncFlatSpec with AsyncMockFactory with Matchers {
+  behavior of "BlocksQueryService.Eval.make query"
+
+  val actorSystem: ActorSystem = ActorSystem("test")
```
SeanCheatham, line 24: Although it's just unit testing, it's still recommended to shutdown these ActorSystems to free up the threads and any associated resources.  The Akka testkit has some helpers for doing some of this automatically through trait mixins: https://doc.akka.io/docs/akka/current/testing.html

genus/src/test/scala/co/topl/genus/interpreters/queryservices/BlocksQueryServiceSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+    (databaseClient.queryBlocks _)
+      .expects(*)
+      .returns(Source(existingBlocks).pure[IO])
+
+    val underTest: BlocksQuery =
+      BlocksQueryService.Eval.make(databaseClient, 5.seconds)
+
+    val result = underTest.query(queryBlocksReq)
+
+    result map { r => r.result.success.get.blocks shouldNot be(empty) }
```
SeanCheatham, line 62: I _think_ `result` is a Future here.  There's a scalatest mixin `ScalaFutures` that gives you access to a `Future#futureValue` method that awaits and handles Failure cases.

genus/src/main/protobuf/blocks_subscription.proto
```diff
@@ -0,0 +1,50 @@
...
+
+  message Start {
+    string subscription_id = 1;
+  }
+
+  message Checkpoint {
+    uint64 timestamp = 1;
+  }
+
+  message Pause {}
```
SeanCheatham, line 45: I think it's outside the scope of this particular PR, but I'm curious what your plan/intention is behind the Pause/Checkpoint mechanism?

b-vennes: I talked about it a little in a [comment above](https://github.com/Topl/Bifrost/pull/1870#discussion_r784342853), but the checkpoint mechanism lets clients acknowledge that they've received an event (block or transaction) from the server. The server will then save this checkpoint to a cache somewhere so that they can reconnect on a failure without having to remember what their own state was. The pause mechanism might not be as helpful but I figured it could be if you wanted to maintain a connection to the server but not receive any more data for some time.

genus/src/test/scala/co/topl/genus/interpreters/queryservices/BlocksQueryServiceSpec.scala
b-vennes: I like the usages of these `mValue` types on `Futures` and `Eithers`. They give every test a similar structure regardless of the monad type.

genus/src/test/scala/co/topl/genus/interpreters/queryservices/BlocksQueryServiceSpec.scala
b-vennes: I wasn't able to mix in `TestKit` with `AsyncFlatSpec`, but I was able to use the `BeforeAndAfterAll` mixin with an override of `afterAll` to shutdown the actor system.

SeanCheatham: You should be fine now if it's covered in `afterAll`.  But to fix the `AsyncFlatSpec` issue, you should be able to mix in `AsyncFlatSpecLike` instead of extending `AsyncFlatSpec`.  `AsyncFlatSpec` is a class which is why you can't extend both `TestKit` and `AsyncFlatSpec`.  But for each of those `*Spec` classes, there's a `*SpecLike` trait that handles this particular issue.

genus/src/test/scala/co/topl/genus/interpreters/queryservices/BlocksQueryServiceSpec.scala
SeanCheatham: Actually on reviewing again, I realized you're using `AsyncFlatSpec` not `AnyFlatSpec`, so what you originally had would work just fine.  I was initially concerned with the `Future` not properly completing (thus raising a test failure), but in `AsyncFlatSpec`, you should be covered there.

genus/src/main/protobuf/blocks_subscription.proto
SeanCheatham: I think checkpointing (or more broadly, subscriptions) may introduce some additional headache.  Introducing persistent client state adds a bit of complexity.  For example, what happens if the Genus server crashes (if a temporary cache is used)?  Or what if Genus is behind a load balancer and the _second_ read request hits a different server?

I'm certainly not saying it's impossible, but it may be worth holding off on it for now and start with a simpler approach without subscriptions.  It seems like a client is already responsible for remembering a subscription ID, so they could instead just remember a block height.

SeanCheatham:
Overall I like this pull request and the approach you're taking with Genus!

genus/src/main/protobuf/blocks_subscription.proto
b-vennes: Yeah that's a good point about complexity. I'm not sure how clients should keep track of their stream state with transactions since they are generally unordered, but I suppose those do have the block height as well.

PR #1879, opened 2022-01-18
Genus Query Paging and Query Validation

### LINKED ISSUE #1861: Add Paging To Query Service
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1046) by [Unito](https://www.unito.io)

b-vennes:
## Purpose
- add paging to query requests so that larger queries can be broken down into smaller pieces
- validate the request parameters in queries (no negative values for paging size or number)
  - the gRPC implementations for clients should block against this since we are using unsigned, but its good to double-check

## Approach
- add a `Paging` type which can be optionally passed in with `QueryTxsReq` and `QueryBlocksReq`
- handle paging values in requests sent to MongoDB in the `MongoQuery` interpreter
- wrap steps in the query services in the `EitherT[F, Failure, *]` type to allow for converting thrown exceptions into failure cases  

## Testing
- adds input validation tests to the `TransactionsQueryServiceSpec` and `BlocksQueryServiceSpec`

## Tickets
- closes #1861


genus/src/main/scala/co/topl/genus/interpreters/MongoDatabaseClient.scala
```diff
@@ -36,14 +38,19 @@ object MongoDatabaseClient {
...
             .map(subscription => subscription.map(_.getFullDocument.transformTo))
 
-        override def queryTransactions(filter: TransactionFilter): F[Source[Transaction, NotUsed]] =
+        override def queryTransactions(
+          filter: TransactionFilter,
+          paging: Option[Paging]
+        ): F[Source[Transaction, NotUsed]] =
           transactionQuery
-            .query(filter.toBsonFilter)
+            // TODO: add some sorting options
```
scasplte2, line 46: are you thinking an enumerated list of optional ordering schemes? Just be careful with what indices you have in Mongo since I believe that could influence the response time (at least something I've hit using Mongoose during Annulus dev)

genus/src/main/scala/co/topl/genus/typeclasses/Validation.scala
```diff
@@ -0,0 +1,8 @@
+package co.topl.genus.typeclasses
+
+import cats.data.ValidatedNec
+import simulacrum.typeclass
+
+@typeclass trait Validation[T] {
+  def validate(value: T): ValidatedNec[String, T]
```
scasplte2, line 7: Validation also makes me think of sanitization. I would guess that our user inputs are not expressive enough to expose some "injection" type attack but wanted to raise the thought to get your opinion.

scasplte2 (approved):
overall seems pretty straightforward so I think we're alright to merge when you're ready

genus/src/main/scala/co/topl/genus/typeclasses/Validation.scala
b-vennes: I think you could have both Validation and Sanitation, since Validation represents a yes/no answer and Sanitation represents a modified input.

I should check and see if the Mongo client sanitizes string inputs for something like an equals query, because if not we could definitely leave ourselves open to an injection attack.

PR #1886, opened 2022-01-24
Genus - Streaming Queries

b-vennes:
## Purpose

- Add the option to stream a query from Genus.
  - use cases are for large queries which may take multiple seconds to complete, or queries where the client does not want to page the results

## Approach

- Refactors the existing gRPC query service implementations into a smaller `QueryServiceAlg` algebra which handles a specific query type `QueryRequest` with a standard query result `Either[QueryFailure, List[T]]` or `Either[QueryFailure, Source[T, NotUsed]]`
  - intepreters/programs which wish to use `QueryServiceAlg` can map their own query requests into the standard `QueryRequest` type which the `QueryServiceAlg` handles
- Removes the `DatabaseClientAlg` since it was not providing much value and made working with both streamed and list queries more difficult

## Testing

- adds some unit tests for `QueryServiceInterp`, `BlocksQueryProgram`, and `TransactionsQueryProgram`

## Tickets
- #1860


scasplte2 (approved):
Couple of questions but overall 🔥 
- any difficulty to transition to typed actor system?
- really like the reduction to a single query service and the abstraction to a datastore
- is the source functor because there is no Functor instance from TypeLevel?

b-vennes:
> is the source functor because there is no Functor instance from TypeLevel?

I was able to define a functor instance for Source, but I think it has something to do with the generic Filter type on the algebra. I'll keep investigating into it as we continue development.

b-vennes:
> any difficulty to transition to typed actor system?

Could you expand on what you mean by this?

scasplte2:
> > any difficulty to transition to typed actor system?
> 
> Could you expand on what you mean by this?

@b-vennes Your QueryServer is using a ClassicActorSystem, that got me thinking about transitioning to a typed actor model (which if I recall classic actor system can't support typed actors but typed actor system can support classic actors). So just wanted to see if there may be hurdles you foresee

b-vennes:
> So just wanted to see if there may be hurdles you foresee

I don't see any hurdles since we can always convert a typed actor system to an untyped one. I'm only using the actor system to power the grpc service and the streams materializer.

PR #1890, opened 2022-01-26
Binary Codecs: Networking Layer

b-vennes:
## Purpose

- replace existing usage of `BifrostSerializer` instances in the `co.topl.network` package
- optimize networking layer to take advantage of the type-class pattern and refactors to the message structures

## Approach
- changes all possible message types into an enumeration of `MessageV1`
- implements a `Transmission` type which parses incoming and outgoing data into a strongly-typed value containing the message header (message code and message length) and the message content (a checksum and the data)
  - this type is clear to work with and can be converted to and from message types
- adds codecs for all message types and uses these codecs to convert message content from transmissions
- `Synchronizer` types define handlers for specific subtypes of message and provide
- `Synchronizer` types register a set of message codes defining the messages they would like to receive from the `NetworkController`

## Testing
- adds compatibility unit tests for the new network message codecs

## Tickets
- #1768


node/src/main/scala/co/topl/network/PeerConnectionHandler.scala
```diff
@@ -279,28 +283,49 @@ class PeerConnectionHandler(
...
     }
 
+  /**
+   * Handles processing an incoming transmission and updating the state of the chunks buffer.
+   * @param transmission the transmission that is ready for processing
+   */
+  private def handleTransmission(transmission: Transmission): Unit = {
+    log.info("Received message " + transmission.header.code + " from " + connectionId)
+
+    networkControllerRef ! NetworkController.ReceivableMessages.TransmissionReceived(transmission, selfPeer)
```
scasplte2, line 321: I know this is a carry over but could we touch base on this `selfPeer`? I was initially having trouble recalling what the variable was for but now looking at it I think we may be able to get rid of it since I think it's only purpose is to be the internal reference (in the local ActorSystem) for communicating with a remote peer.

scasplte2: we should probably address in a separate PR but leaving the comment here so we can chat

node/src/main/scala/co/topl/network/codecs/scodecs/network/peer/PeerCodecs.scala
```diff
@@ -0,0 +1,53 @@
...
+            bytesCodec(length)
+              .exmapc[LocalAddressPeerFeature](bytes =>
+                localAddressPeerFeatureCodec.decode(BitVector(bytes)).map(_.value)
+              )(localAddress => localAddressPeerFeatureCodec.encode(localAddress).map(_.toByteArray))
+          )(localAddress =>
+            localAddressPeerFeatureCodec.encode(localAddress).map(_.length / 8).getOrElse(0L).toShortExact
+          )
+      )
+
+  implicit val peerSpecCodec: Codec[PeerSpec] =
```
scasplte2, line 46: `PeerSpec` is such an awful name for what this is

node/src/main/scala/co/topl/network/message/Transmission.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+case class TransmissionContent(checksum: Array[Byte], data: Array[Byte])
+
+/**
+ * Wrapper for a transmitted network message, whether it came from an external peer or was generated locally.
+ *
+ * @param header the header of the transmitted message
+ * @param content the message's checksum and encoded content
+ */
+case class Transmission(header: TransmissionHeader, content: Option[TransmissionContent]) {
```
scasplte2, line 26: This is a much easier thing to understand than how this was handled previously ✅

node/src/main/scala/co/topl/network/message/package.scala
```diff
@@ -0,0 +1,133 @@
...
+import co.topl.modifier.NodeViewModifier.ModifierTypeId
+import co.topl.network.peer.PeerSpec
+import co.topl.settings.Version
+
+package object message {
+
+  type MessageCode = Byte
+
+  /** Enumeration of app p2p messages in the network */
+  sealed trait Message {
```
scasplte2, line 13: ADTs for the win, this feels much more ergonomic and well defined now. Also thank you for including the descriptive comments

node/src/main/scala/co/topl/settings/AppContext.scala
```diff
@@ -1,8 +1,7 @@
 package co.topl.settings
 
-import co.topl.network.message._
+import co.topl.network.peer
```
scasplte2, line 3: Looking around the codebase, it seems like we might be able to rename this to something along the lines of `RuntimeNetworkConfig` (just to be more descriptive) and we can remove the `features` val as it seems the only place it is used is in the `NetworkController` where we assume all nodes have the `LocalAddressFeature`. Even if we want to retain the ability to distinguish application version features, this is probably the wrong place to specify them.

scasplte2 (approved):
👇 no other good place for that comment but overall this is really well done @b-vennes! It was a headache and a half when I first tried understanding how the messages interacted with one another and got into and out of the system. But this change feels like it draws the right lines in the right places by separating and disentangling complexities.

---
While reviewing, I was looking at `UByte` trying to see if it was used anywhere and noticed the type aliases in the valuetypes package object for `UByte, ULong, UShort, UInt`. Two questions came to mind. (1) Should `ULong` be a long? Shouldn't it need to go to a larger type similar to the others? (2) Since these are type aliases, are the type classes instances unique? Could we ever have a problem at runtime where `Codec[UByte]` is needed but `Codec[UShort]` is around so it gets used instead since `UByte` and `UShort` are both aliases for `Int`?

scasplte2 (changes requested):
apologies, meant to say lets chat about some of the comments before merging but overall I don't think there is much that needs to be changed

node/src/main/scala/co/topl/network/PeerConnectionHandler.scala
b-vennes: I think this might be a mistake. It should be sending the `TransmissionReceived` message with the connected peer who sent the transmission and not the self peer.

b-vennes: I think we might still need this concept so that we can send messages to the `PeerConnectionHandler` which in turns sends the transmissions to the connected peer.

b-vennes:
> While reviewing, I was looking at `UByte` trying to see if it was used anywhere and noticed the type aliases in the valuetypes package object for `UByte, ULong, UShort, UInt`. Two questions came to mind. (1) Should `ULong` be a long? Shouldn't it need to go to a larger type similar to the others? (2) Since these are type aliases, are the type classes instances unique? Could we ever have a problem at runtime where `Codec[UByte]` is needed but `Codec[UShort]` is around so it gets used instead since `UByte` and `UShort` are both aliases for `Int`?

An interesting thing about writing typeclasses in Scala and using type aliases is that the typeclass will only apply to the alias and not the aliased type. This means that if I want an implicit `Codec[ULong]`, then an implicit `Codec[Long]` will not work. It has to be the exact type. However, I can summon a `Codec[Long]` from implicit scope and call the `encode` function on a `ULong` value if I wish because the type alias system will automatically convert to the aliased type.

This is what has been biting us for things like the Proofs and Proposition codecs. The previous serialization scheme depended on inheritance of a BifrostSerializer in order to serialize the value which doesn't work for typeclasses.

node/src/main/scala/co/topl/network/PeerConnectionHandler.scala
scasplte2: yes, we do need the concept, but the PeerConnectionHandler just sends an ActorRef to the NetworkController to effectively say "data to this IP needs to be forwarded to this ActorRef" but I don't think we need the selfPeer variable as when that "registration" happens couldn't we just query the underlying ActorRef for the PeerConnectionHandler actor?

node/src/main/scala/co/topl/network/codecs/scodecs/network/peer/PeerCodecs.scala
scasplte2: we have a `PeerInfo` but `PeerMetadata` is at least a little more descriptive

node/src/main/scala/co/topl/network/PeerConnectionHandler.scala
b-vennes: I've created a new issue to remove usages of the `selfPeer` value. After looking some more it shouldn't be much work, but it will require some time for testing.

https://topl.atlassian.net/browse/CORE-1071

scasplte2 (approved):
🔥 ready when you are Branden

Issue #1438, opened 2021-07-15
Consensus Storage and Protocol Manager Actors


SeanCheatham:
Currently, the `co.topl.consensus` package object contains mutable references to a `ConsensusStorage` and a `ProtocolVersioner`.  These values are accessed and mutated by events taking place within the rest of the system.  This places these values in a JVM-global mutable singleton which may exit the encapsulation of a particular ActorSystem.  The values are also thread-unsafe, meaning two threads could attempt an update at the same exact time leading to inconsistent data.

ConsensusStorage and ProtocolVersioner should instead be captured in Actors which can maintain a thread-safe state of the two objects.  Access to these items will also need to be made asynchronous as well.

Acceptance Criteria:

* The `co.topl.consensus` package object does not contain mutable data
* ConsensusStorage and ProtocolVersioner are thread-safe

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-809) by [Unito](https://www.unito.io)

PR #1892, opened 2022-01-31
Consensus storage actor

jingsting:
## Purpose
We previously keep track of the consensus variables by modifying a consensusStorage object in the consensus package object. Using a stateless typed actor could make accessing and updating the consensus variables safer.

## Approach
Added a typed actor for keeping the consensus variables in memory and in storage. Update the variables by subscribing to newly appended semantically valid blocks. Provide methods to access the consensus variables, and roll back to a previous version.

## Testing
- Test if the consensus params are updated when a block is appended
- Test the loading consensus variables from storage
- Test returning the default values if there's no data in storage at start up
- Test using rollback with a block id that is not available in the saved versions
- Test rollback to a saved version

## Tickets
* #1438
* #1888


node/src/main/scala/co/topl/consensus/ConsensusVariables.scala
```diff
@@ -0,0 +1,254 @@
...
+
+        // Update the storage values
+        storage.update(versionId, Seq(), toUpdate)
+
+        if (storage.latestVersionId().exists(id => id sameElements versionId)) replyTo ! StatusReply.success(Done)
+        else replyTo ! StatusReply.error(new Exception("Failed to update consensus variables in storage"))
+
+        active(storage, updatedParams)
+
+      case (_, ReceivableMessages.RollbackConsensusVariables(blockId, replyTo)) =>
```
b-vennes, line 144: This might have already been discussed, but do we want to include this rollback implementation if we aren't planning on ever needing it?

node/src/main/scala/co/topl/nodeView/NodeViewHolder.scala
```diff
@@ -223,31 +237,48 @@ object NodeViewHolder {
           Behaviors.same
 
         case (context, ReceivableMessages.WriteBlock(block)) =>
+          context.pipeToSelf(consensusVariablesInterface.get.value) {
+            case Success(params) =>
```
b-vennes, line 241: I tried rewriting this section with cats and ended up with:

```
// add cats implicits at the top
context.pipeToSelf(consensusVariablesInterface.get.value)(
  _.fold(
    error => ReceivableMessages.Terminate(error),
    _.fold(
      error => ReceivableMessages.Terminate(error.reason),
      params => ReceivableMessages.WriteBlockWithConsensusParams(block, params)
    )
  )
)
```

Which may or may not be worse than what is currently there :smile:

node/src/main/scala/co/topl/nodeView/NodeViewHolder.scala
```diff
@@ -260,6 +291,19 @@ object NodeViewHolder {
     t
   }
 
+  private[nodeView] def getConsensusParams(
```
b-vennes, line 294: I don't think this function is ever used

node/src/main/scala/co/topl/tool/Exporter.scala
```diff
@@ -18,7 +19,8 @@ import scala.util.{Failure, Success}
 
 object Exporter extends Logging {
 
-  private def initHistory(settings: AppSettings, np: NetworkPrefix): History = History.readOrGenerate(settings)(np)
+  private def initHistory(settings: AppSettings, np: NetworkPrefix): History =
+    History.readOrGenerate(settings)(np, nxtLeaderElection = NxtLeaderElection(settings))
```
b-vennes, line 23: I think you can remove the explicit assignment of the `nxtLeaderElection` parameter

b-vennes (commented):
I like having the consensus parameters actor wrapped behind the interface and I like that we now have leader election as a parameter. I think we should consider making leader election an explicit parameter instead of an implicit one, since it is a standard value and tends to be explicitly provided anyways.

node/src/main/scala/co/topl/Heimdall.scala
```diff
@@ -81,7 +83,7 @@ object Heimdall {
     settings:              AppSettings,
     appContext:            AppContext,
     state:                 NodeViewHolderInitializingState
-  )(implicit timeProvider: TimeProvider): Behavior[ReceivableMessage] =
+  )(implicit timeProvider: TimeProvider, nxtLeaderElection: NxtLeaderElection): Behavior[ReceivableMessage] =
```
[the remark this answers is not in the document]
jingsting: That makes sense, I'll change that into an explicit parameter 👍

node/src/main/scala/co/topl/nodeView/NodeViewHolder.scala
jingsting: Oh I forgot to remove it, I'll remove it

node/src/main/scala/co/topl/consensus/Forger.scala
```diff
@@ -236,14 +261,18 @@ private class ForgerBehaviors(
...
-        nodeViewReader
-          .withNodeView(Forge.fromNodeView(_, keyView, minTransactionFee))
-          .leftMap(e => ForgingError(e.reason))
-          .subflatMap(_.leftMap(ForgeFailure))
+    for {
+      keyView <- EitherT[Future, ForgerFailure, KeyView](
+        fetchKeyView().map(Right(_)).recover { case e => Left(ForgingError(e)) }
       )
-      .subflatMap(_.make.leftMap(ForgeFailure))
+      consensusParams <- consensusVariablesInterface.get
```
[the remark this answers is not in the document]
jingsting: Yeah, if the consensus variables actor failed to update the params when a new block is appended, we would be using the wrong consensus params during forging. I added a check to see if the parent block's height and difficulty matches those in consensus storage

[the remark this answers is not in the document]
b-vennes: We probably just need to check height, but its probably okay either way.

node/src/main/scala/co/topl/consensus/ConsensusVariables.scala
jingsting: I think we said we were going to leave it there since we will need that function

node/src/main/scala/co/topl/consensus/Forger.scala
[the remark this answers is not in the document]
jingsting: Yeah, I realize that we really just need the height to match

jingsting:
@scasplte2 This PR is ready for review

b-vennes (approved):

PR #1893, opened 2022-01-31
Release Candidate v1.10

scasplte2:
## Tickets
*

**Delete everything after and including this line after you are done reading and understand what to put for each section**

>### Purpose
>
>*State in plain words what the scope of the pull request and what problem it is trying to solve*
>
>### Approach
>
>*Enumerate the steps you took to create the code/changes contained in this pull request*
>
>### Testing
>
>*Specify all unit, integration, or end-to-end testing that is contained in this pull request*
>
>### Tickets
>
>*List tickets if any that are connected to this pull request in the following format:*
>
>_* closes #1234_


node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
```diff
@@ -47,18 +65,60 @@ class NodeViewRpcHandlerImpls(
 
   override val transactionById: ToplRpc.NodeView.TransactionById.rpc.ServerHandler =
     params =>
-      withNodeView(_.history.transactionById(params.transactionId))
-        .subflatMap(
-          _.toRight[RpcError](InvalidParametersError.adhoc("Unable to find confirmed transaction", "modifierId"))
-        )
-        .map { case (tx, blockId, blockNumber) => ToplRpc.NodeView.TransactionById.Response(tx, blockNumber, blockId) }
+      getTxsByIds(List(params.transactionId)).map {
+        _.head match {
```
SeanCheatham, line 69: This seems like it could potentially throw an exception if the transaction is not found

node/src/main/scala/co/topl/rpc/handlers/package.scala
```diff
@@ -19,4 +29,92 @@ package object handlers {
...
+    blocksOption: List[Option[Block]],
+    sizeLimit:    Int
+  ): Either[RpcError, List[Block]] =
+    for {
+      _ <- Either.cond(
+        ids.size <= sizeLimit,
+        {},
+        ToplRpcErrors.unsupportedOperation("Number of ids given exceeded blockRetrievalLimit")
+      )
+      blocks <- Either.cond(
```
SeanCheatham, line 60: Stylistic preference:
```
import cats.implicits._

blocks <- blockOptions.sequence.toRight(ToplRpcErrors.NoBlockWithId)
```

b-vennes (approved):
This RC looks good :tada:

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
jingsting: The getTxByIds calls checkTxFoundWithIds at the end of the for comprehension, it returns a ToplRpcErrors.NoTransactionWithId if the id is not found

b-vennes: I just looked up the implementation of `getTxsByIds` and it does do a nonempty check on the result internally and returns a `NoTransactionWithId` in the case of an empty result.

node/src/main/scala/co/topl/rpc/handlers/package.scala
jingsting: This looks nice! I think I wanted something like this before 🔥 So I read this can turn `List[Option[Int]]` into a `Option[List[Int]]` and in this case if there is a None in the List, then it would get turned into a None right?

b-vennes: I also like `sequence`. I assume it does something like `blockOpts.flatMap(opt => opt.map(List(_)).getOrElse(List()))`?

jingsting: Oh I think there might be a few more places that need this

node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
scasplte2: I attempted to more obviously handle this here https://github.com/Topl/Bifrost/pull/1900/files#diff-e2af64b047560b250465f1e6d97319a86ddcdfb933e9d973278cdc284e482476R68

as previously stated I don't think we would get to an error state but I thought this handling was a little more explicit about it. Open to thoughts on the matter though.

PR #1895, opened 2022-02-01
Tetra Registration Box Remove activationSlot

SeanCheatham:
## Purpose
- Registrations only become valid 2 epochs after the initial registration
- This restriction will now be implicitly checked by the book-keeping of "Consensus State" which tracks registrations through the epochs.  (Consensus State book-keeping will be fully implemented later)
 
## Approach
- Remove `activationSlot` from Registration Box value

## Testing
- `preparePR` still passes

### DIFF
--- a/demo/src/main/scala/co/topl/demo/TetraDemo.scala
+++ b/demo/src/main/scala/co/topl/demo/TetraDemo.scala
@@ -80,8 +80,7 @@ object TetraDemo extends IOApp.Simple {
         commitment = kesProduct.sign(
           kesKey,
           new Blake2b256().hash(ed25519Vrf.getVerificationKey(stakerVrfKey).signableBytes, poolVK.bytes.data).data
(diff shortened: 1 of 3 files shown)


demo/src/main/scala/co/topl/demo/TetraDemo.scala
```diff
@@ -77,8 +77,7 @@ object TetraDemo extends IOApp.Simple {
               .hash((ed25519Vrf.getVerificationKey(stakerVrfKey).signableBytes ++ poolVK.bytes.data).toArray)
               .value
           )
-        ),
-        activationSlot = 0
```
b-vennes, line 81: I think you can remove the `commitment =` call above here now that you only have one parameter.

b-vennes (approved):

scasplte2 (approved):
☑️

PR #1899, opened 2022-02-01
Genus Subscriptions

### LINKED ISSUE #1862: Subscription Service Interpreter
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1047) by [Unito](https://www.unito.io)

b-vennes:
## Purpose

- Add implementations for Genus Subscription services

## Approach

- Slims down the subscription service gRPC interface to take in a single request with an optional block height checkpoint
- Implements the subscription service for blocks and transactions
- Refactors HTTP server creation into a single algebra and interpreter
- Adds MongoOplog algebra and interpreter for getting data from the Mongo Op Log collection to help with subscriptions
- Fixes validation to return multiple validation failures instead of just the first one
- Adds extension operations to types and companions defined in protobuf
  - new operations can be found in the `co.topl.genus.ops.protobufops` package 

## Testing
- todo: add tests after getting feedback

## Tickets
* closes #1862


scasplte2 (approved):
every time I think it is great, it just gets better!

PR #1900, opened 2022-02-02
Minor update to release candidate 1.10

### LINKED ISSUE #1873: Remove the serve method in class RpcServer
## Description
The method `serve` in class RpcServer is no longer needed since we use the method build for constructing the API routes.
Also check if there's any methods that is no longer useful in the class RpcClient

## Acceptance Criteria
* Remove the method `serve` in class RpcServer
* Remove the test specs for the `serve` method

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1055) by [Unito](https://www.unito.io)

scasplte2:
## Purpose
Minor refactoring of 1.10 release candidate. Separate PR to aid in review

## Approach
- Removed `.serve` method from `RpcServer` and transitioned tests to target the `RpcBuilder` instead
- Separated the retrieval limit check to happen before querying `view` in order to fail fast
- Handle potential edge case in `getTransatctionById` to avoid `.head` usage
- Removed `check___FoundWithIds` function defs and replaced with lambdas

## Testing
- Passes unit tests
- Passes integration tests

## Tickets
- closes #1873

### DIFF
--- a/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
+++ b/node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
@@ -66,7 +66,7 @@ class NodeViewRpcHandlerImpls(
   override val transactionById: ToplRpc.NodeView.TransactionById.rpc.ServerHandler =
     params =>
       getTxsByIds(List(params.transactionId)).subflatMap {
-        case (tx, blockId, blockNumber) :: Nil =>
+        case (tx, blockId, blockNumber) :: _ =>
           Either.right(ToplRpc.NodeView.TransactionById.Response(tx, blockNumber, blockId))
         case _ => Either.left(ToplRpcErrors.NoTransactionWithId)
       }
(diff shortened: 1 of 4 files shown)


node/src/main/scala/co/topl/rpc/handlers/NodeViewRpcHandlerImpls.scala
```diff
@@ -65,10 +65,10 @@ class NodeViewRpcHandlerImpls(
 
   override val transactionById: ToplRpc.NodeView.TransactionById.rpc.ServerHandler =
     params =>
-      getTxsByIds(List(params.transactionId)).map {
-        _.head match {
-          case (tx, blockId, blockNumber) => ToplRpc.NodeView.TransactionById.Response(tx, blockNumber, blockId)
-        }
+      getTxsByIds(List(params.transactionId)).subflatMap {
+        case (tx, blockId, blockNumber) :: Nil =>
```
b-vennes, line 69: should we change the `Nil` to `_` on the off-chance that we get multiple transaction IDs back? I don't think this should ever happen, but on the off chance it does, it might be a better response than saying no transactions were found

scasplte2: that is a better behavior 👍 I wasn't happy about the second case catching what you described but I think it would accurate if we make this replacement

PR #1915, opened 2022-02-10
Handle Block Cache Evictions

### LINKED ISSUE #1914: Delivery Tracker Not Reset On Block Cache Eviction
h3. Statement of Problem

When the block cache in {{NodeViewHolder}} overflows, the blocks are forgotten about, but the {{NodeViewSynchronizer}}  never updates its {{DeliveryTracker}}  to reflect the lost information. When the {{NodeViewSynchronizer}} then goes to get the blocks again from its peer, the {{DeliveryTracker}} show the status as being {{Requested}} , which causes the {{NodeViewSynchronizer}} to ignore requesting the evicted blocks.

h3. Acceptance Criteria

* Blocks evicted from the block cache in the {{NodeViewHolder}} will be requested on a future synchronization with a peer node

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1081) by [Unito](https://www.unito.io)

b-vennes:
## Purpose
The `NodeViewSynchronizer` needs to be notified if a block is evicted from the NodeViewHolder's block cache. If it is not notified, the delivery tracking mechanism will always ignore requests for that block in the future because it thinks its been successfully handled.

## Approach
- adds an `onEvict` function to the `SortedCache#apply` method which can be used as a callback for when a block is evicted
- `NodeViewHolder` publishes a message when a block is evicted from the cache
- `NodeViewSynchronizer` subscribes to cache overflows so that the delivery tracker can be updated to reflect lost blocks

## Testing
- no new tests added

## Tickets
- closes #1914

### DIFF
--- a/common/src/main/scala/co/topl/utils/actors/SortedCache.scala
+++ b/common/src/main/scala/co/topl/utils/actors/SortedCache.scala
@@ -39,15 +39,25 @@ object SortedCache {
    *                  overflows, entries at the end of the Ordering will be evicted
    * @param itemPopLimit The maximum number of times that a single item can be attempted to be popped.  If an item
    *                     is determined not to be a viable candidate after this many attempts, it is evicted
+   * @param onEvict A callback function to run on an item when it is evicted from the cache.
    */
   def apply[T: Ordering: Show](
     itemLimit:    Int = Int.MaxValue,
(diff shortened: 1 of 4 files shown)


scasplte2 (approved):
👍 merging and will tag to begin ecosystem integration testing

PR #1946, opened 2022-02-24
Dev Merge, Scodec Bits

SeanCheatham:
## Purpose
- The `tetra` branch is out-of-date with the `dev` branch
- The `tetra` branch uses the legacy encoding system for binary serialization and needs to be updated to the BV method

## Approach
- Merge `dev` branch
- Copy BV codec typeclasses from `common` into `byte-codecs`
- Create Identifiable, Signable, and Immutable typeclasses
- Create `tetra-byte-codecs` module with implementations of `byte-codecs` typeclasses
- Incorporate new typeclasses into Tetra codebase

## Testing
- Implemented new unit tests for the TetraScodecCodecs that are implemented
- NOTE: BloomFilterSpec had issues passing, so I adjusted it to pass

## Tickets
_* closes #1921_


json-codecs/src/main/scala/co/topl/codecs/json/tetra/ModelsJsonCodecs.scala
```diff
@@ -0,0 +1,166 @@
...
+import co.topl.models._
+import io.circe.syntax.EncoderOps
+import io.circe.{Encoder, Json}
+
+trait ModelsJsonCodecs {
+
+  implicit val dionAddressEncoder: Encoder[DionAddress] =
+    t => t.allBytes.toBase58.asJson
+
+  implicit def propositionEncoder: Encoder[Proposition] = {
```
b-vennes, line 13: does circe offer any sort of support for Sum Types?

SeanCheatham: Circe has _some_ support for that sort of thing: https://circe.github.io/circe/codecs/adt.html#a-more-generic-solution

But when doing so manually, you can either do so without a "discriminator" (which would attempt to find the correct sum type using field names) or with a "discriminator" (which places a type byte at the front of the encoded data).

byte-codecs/src/main/scala/co/topl/codecs/bytes/typeclasses/ImmutableCodec.scala
```diff
@@ -0,0 +1,71 @@
...
+}
+
+@typeclass trait ImmutableEncoder[T] {
+
+  /**
+   * Gets the byte representation of the value that should be persisted to a data store.
+   * @param value the value to convert into bytes
+   * @return an array of bytes representing the value which should then be persisted as-is
+   */
+  @op("immutableBytes") def immutableBytes(value: T): ByteVector
```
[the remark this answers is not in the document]
SeanCheatham: I agree - I dislike `ImmutableBytes` for the same reason.  Perhaps `StableBytes` might work?

brambl/src/main/scala/co/topl/credential/KeyCollection.scala
```diff
@@ -4,15 +4,17 @@ import cats._
...
+import co.topl.codecs.bytes.typeclasses.Persistable
 
+//todo: JAA - I am doing something non-ideal here for testing. I want to store keyfile as json (we can also have byte persistable)
+// JSON is the format that wallet and things might use, we are not necessarily tied to using JSON but it would break backwards compatibility
+// so we need to determine if it is worth it
 trait KeyCollection[F[_]] {
   def unlockedEntries: F[Set[TypedEvidence]]
-  def unlock[SK <: SecretKey: ByteCodec](evidence: TypedEvidence, password: Password): F[Option[SK]]
-  def lift[SK <: SecretKey](evidence:              TypedEvidence): F[Option[SK]]
+  def unlock[SK <: SecretKey: Persistable](evidence: TypedEvidence, password: Password): F[Option[SK]]
```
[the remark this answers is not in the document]
SeanCheatham: I think I prefer keeping `Persistable` as a context-bound to allow a user to override/specify the codec for their secret key.  This allows 3rd party developers to provide a specialized encoding scheme (that perhaps encrypts and decrypts the data bytes in addition to the encryption included internally).

brambl/src/main/scala/co/topl/credential/CredentialIO.scala
```diff
@@ -75,20 +75,21 @@ case class DiskCredentialIO[F[_]: Sync](basePath: Path) extends CredentialIO[F]
   def unlock(evidence: TypedEvidence, password: Password): F[Option[(Bytes, KeyFile.Metadata)]] =
     Sync[F].blocking {
       val keyFilePath = Paths.get(basePath.toString, s"${evidence.allBytes.toBase58}.json")
-      Option
-        .when(Files.exists(keyFilePath) && Files.isRegularFile(keyFilePath))(
-          Files.readString(keyFilePath, StandardCharsets.UTF_8)
-        )
-        .flatMap(io.circe.parser.parse(_).flatMap(_.as[KeyFile]).toOption)
-        .flatMap(keyFile => KeyFile.Encryption.decrypt(keyFile, password).toOption.map((_, keyFile.metadata)))
+      if (Files.exists(keyFilePath) && Files.isRegularFile(keyFilePath))
```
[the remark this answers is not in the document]
SeanCheatham: `Option.when` doesn't exist in Scala 2.12 :frowning:

I think if you're targeting an _application_ on Scala 2.13, then you're free to use Option.when.  But since `brambl` is a cross-versioned library, I figured it easier to just replace the usage of `Option.when`

brambl/src/main/scala/co/topl/credential/Credential.scala
```diff
@@ -45,38 +49,33 @@ object Credential {
...
       unprovenTransaction:      Transaction.Unproven
     )(implicit extendedEd25519: co.topl.crypto.signing.ExtendedEd25519)
         extends Credential {
+      val proposition: Proposition = sk.vk.asProposition
+      def prove(currentProof: Proof): Proof = (sk, unprovenTransaction).asProof
+    }
 
-      val proposition: Proposition =
-        sk.vk.asProposition
+    case class HashLock(salt: Digest32, value: Byte) extends Credential {
```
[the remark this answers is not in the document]
SeanCheatham: This is one that @scasplte2 implemented so he may be able to provide more details.

This is essentially a "Proof of Knowledge of Text".  It creates a Box in the chain which is locked by a proposition containing some hash value.  To unlock the box, the person proves knowledge of the text/string that, when hashed, matches the hash on the proposition.  (The salt value further locks it down like when storing password hashes in a database)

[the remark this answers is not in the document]
scasplte2: yes, Sean nailed the description. Hash Time Lock contracts are a useful primitve to have available (a HashLock && HeightLock) and can be used for atomic-swap contracts. 
https://academy.binance.com/en/glossary/hashed-timelock-contract

An atomic-swap is one where Alice has token A and wants token B, Bob has token B and wants A. So Alice will commit some amount of token A to a box that is locked with a HashLock. Bob can also create a box of token B using the same HashLock. Once, Bob's token B is locked then Alice can spend it since she knows the input value that produces the HashLock so she can create a valid proof. But in order to spend the funds that Bob locked up, Alice must reveal the secret value to claim the token B box. Finally, once the input is reveal, Bob can use it to claim the token A box. Addition of appropriate HieghtLocks then eliminate Alice's ability to claim both tokens and coupling with signing routines can allow for buybacks.

common/src/main/scala/co/topl/attestation/keyManagement/Keyfile.scala
```diff
@@ -88,10 +89,4 @@ trait KeyfileCompanion[S <: Secret, KF <: Keyfile[S]] {
...
 }
 
-object Keyfile {
-
-  implicit def jsonEncoder[KF <: Keyfile[_]]: Encoder[KF] = {
-    case kfc: KeyfileCurve25519 => KeyfileCurve25519.jsonEncoder(kfc)
-    case kfe: KeyfileEd25519    => KeyfileEd25519.jsonEncoder(kfe)
-  }
-}
+object Keyfile {}
```
scasplte2, line 92: should remove this when we can get to it

common/src/main/scala/co/topl/codecs/binary/typeclasses/BinaryShow.scala
```diff
@@ -0,0 +1,59 @@
+package co.topl.codecs.binary.typeclasses
```
scasplte2, line 1: Can we remove these typeclasses from `common` and have `common` depend on `byte-codecs`?

node/src/main/resources/application.conf
```diff
@@ -207,7 +207,7 @@ bifrost {
 
   application {
     # application version of this software
-    version { value = 1.7.6 }
+    version { value = 1.10.0 }
```
scasplte2, line 210: we don't need those values from above right? I think those are left over from Aaron trying to bring the Prosomo stakeholder over

node/src/main/scala/co/topl/consensus/genesis/HelGenesis.scala
```diff
@@ -20,13 +20,16 @@ case object HelGenesis extends GenesisProvider {
   implicit val networkPrefix: NetworkPrefix = NetworkType.HelTestnet.netPrefix
 
   override protected val blockChecksum: ModifierId =
-    ModifierId.fromBase58(Base58Data.unsafe("29jsARiMtRqoE5Z8okN6fddDVN4HHeDb7vdvsaEPVzgXD"))
+    Base58Data
+      .unsafe("vKjyX77HLRUiihjWofSsacNEdDGMaJpNJTQMXkRyJkP2")
+      .decodeTransmitted[ModifierId]
```
scasplte2, line 25: probably makes more sense if this is `Immutable`

common/src/main/scala/co/topl/codecs/binary/typeclasses/BinaryShow.scala
SeanCheatham: Dion still uses `Array[Byte]` everywhere which is the default input/output of the codec typeclasses in `common`.  Once Dion is updated to use `ByteVector`, we should be able to get rid of the codec typeclasses from `common`.

models/src/main/scala/co/topl/models/Box.scala
```diff
@@ -3,12 +3,15 @@ package co.topl.models
 import co.topl.models.utility.StringDataTypes.Latin1Data
 import co.topl.models.utility.{Lengths, Sized}
 
-case class Box[V <: Box.Value](evidence: Evidence, nonce: BoxNonce, value: V)
+import scala.util.Random
+
+case class Box[V <: Box.Value](evidence: TypedEvidence, nonce: BoxNonce, value: V, data: Int)
```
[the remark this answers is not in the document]
SeanCheatham: Removed `Box#data`.  It seems to primarily be used for the `RequiredBoxState` proposition.  Should that be removed as well?

node/src/main/resources/application.conf
SeanCheatham: Removed the Prosomo config values

node/src/main/scala/co/topl/consensus/genesis/HelGenesis.scala
SeanCheatham: Once Dion is ready to switch over to the typeclasses in `byte-codecs`, we can change this to use `Immutable`

crypto/src/main/scala/co/topl/crypto/crypto.scala
```diff
@@ -25,6 +25,12 @@ package object crypto {
   @newtype
   case class PublicKey(value: Array[Byte])
 
-  object implicits extends digest.Instances with digest.Digest.ToDigestOps with hash.Instances with PrivateKey.Instances
-
+  trait Implicits
```
scasplte2, line 28: @aaronschutza do you recall which branch you KES tests were on? I don't see them included here so just want to make sure we don't lose those updates and that they will (or are) included in Tetra

models/src/main/scala/co/topl/models/TaktikosAddress.scala
```diff
@@ -8,7 +8,7 @@ import co.topl.models.utility.{Lengths, Sized}
  * @param signature A commitment from the staker (paySK) to the poolVK
  */
 case class TaktikosAddress(
-  paymentVKEvidence: Evidence,
-  poolVK:            VerificationKeys.Ed25519,
-  signature:         Proofs.Knowledge.Ed25519
+  paymentVKEvidence: Evidence, // todo: should be typed evidence, rename to spendingEvidence
```
scasplte2, line 11: nit picky, but I want to change these member names to be something like (spendingEvidence, stakingEvidence, bindingCommit) the last one may be generalizable to a Proof of Knowledge since you can construct a commitment scheme with such a primitive. It also feels like this may be pointing to something more abstract.

node/src/it/resources/postman-tests/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.3"
+
+services:
+  bifrost:
+    build: ../../../../target/docker/stage
+    ports:
+      - 9085:9085
+    command: -s test --apiKeyHash As3ZuwnL9LpoW3wz8HoDpHtZqJ4dhPFFnv87GYrnCYKj
```
scasplte2, line 8: we talked about changing this, lets slate it for 1.11

typeclasses/src/main/scala/co/topl/typeclasses/ContainsEvidence.scala
```diff
@@ -20,7 +20,7 @@ object ContainsEvidence {
   trait Instances {
 
     implicit val ratioContainsEvidence: ContainsEvidence[Ratio] =
-      ratio => TypedEvidence(10: Byte, Sized.strictUnsafe(Bytes(blake2b256.hash(ratio.bytes.toArray).value)))
+      ratio => TypedEvidence(10: Byte, new Blake2b256().hash(ratio.immutableBytes))
```
scasplte2, line 23: Should we just incorporate the hash specification into the `ContainsEvidence` type class? Then we just have to pass in the type prefix and an instance of `ImmutableBytes` (since any evidence that we assign a type prefix must be remain immutably once it has been used on the network)

byte-codecs/src/main/scala/co/topl/codecs/bytes/typeclasses/ImmutableCodec.scala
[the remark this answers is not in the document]
scasplte2: That is a good point that I had not thought about. `StableBytes` is pretty descriptive but what about `BackwardsCompatibleBytes`? It is long but I feel like it conveys the appropriate level of information needed

models/src/main/scala/co/topl/models/Box.scala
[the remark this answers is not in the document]
scasplte2: It was really put there for `EnumeratedData` but needed to be used for the `RequiredBoxState` but no, the `RequiredBoxState` is a really really important piece of the extended-utxo contract system so I want to keep it around.

scasplte2 (changes requested):
Waiting to go over notes with the team so we can record tickets into the ice box and what not. Then this will be ready to merge

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/ops/ScodecOps.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.language.implicitConversions
+
+final class ScodecOps[T](private val codec: Codec[T]) extends AnyVal {
+
+  /**
+   * Maps an error with decoding using the [[Codec]] to another error type.
+   * @param fErr the mapping function to generate the new error
+   * @return a codec with a new error message on decoding failure
+   */
+  def mapDecodeErr(fErr: Err => Err): Codec[T] =
```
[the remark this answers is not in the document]
b-vennes: `mapDecodeErr` is effectively a `leftMap` for a Scodec `Codec` instance. There is currently a way to `leftMap` on an `Attempt` once you've attempted a decode or encode, but there is no way to globally set what the error message should be. I've used this in a few places to provide better error messages than the default `expected X bits in the bit vector`.

typeclasses/src/main/scala/co/topl/typeclasses/ContainsEvidence.scala
SeanCheatham: https://topl.atlassian.net/browse/CORE-1111

node/src/main/scala/co/topl/nodeView/NodeView.scala
```diff
@@ -0,0 +1,398 @@
...
+    state match {
+      case c: AutoCloseable =>
+        c.close()
+      case _ =>
+    }
+  }
+
+  protected val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  // todo: this method causes delays in a block processing as it removes transactions from mempool and checks
```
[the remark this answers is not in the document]
SeanCheatham: Removed comment from code per discussion in review

node/src/main/scala/co/topl/nodeView/MemPoolAuditor.scala
```diff
@@ -1,198 +1,167 @@
...
-    worker ! RunCleanup(state, mempool)
-    context become working // ignore other triggers until work is done
+    import scala.concurrent.duration._
+    implicit val timeout: Timeout = Timeout(10.seconds)
+    implicit val typedSystem: ActorSystem[_] = context.system
+    nodeViewHolderRef.askWithStatus[T](NodeViewHolder.ReceivableMessages.Read(f, _))
   }
 
-  private def rebroadcastTransactions(ids: Seq[ModifierId]): Unit = {
+  // TODO: Jing - Consider changing this to just sending the ids to the networkController and let it form the message
```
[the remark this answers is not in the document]
SeanCheatham: Removed comment from code per discussion in review

models/src/main/scala/co/topl/models/Box.scala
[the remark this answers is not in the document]
SeanCheatham: Removed the commented code from ProofVerifier as discussed in review

crypto/src/main/scala/co/topl/crypto/crypto.scala
scasplte2: Aaron recalled the PR as https://github.com/Topl/Bifrost/pull/1846 and I verified that the tests are in the correct place

PR #1948, opened 2022-02-28
Event Tree Abstraction

SeanCheatham:
## Purpose
- Classic event sourcing takes a single linear chain of events to construct a view of the state
- Blockchains aren't necessarily a single linear chain of blocks; there are small forks and branches along the way
- An "Event Sourced Tree" is like classical event sourcing except that it supports "undoing" events in order to traverse various branches
- Unrelated: Moved several interpreters from `demo` module into new `commonInterpreters` module
## Approach
- Created `EventSourcedState`
  - Takes an event store, undo store, state store, and parent-child tree to construct stateful views at any arbitrary point in a tree
- Created `ParentChildTree`
  - Builds associations between two nodes in a directed, acyclic graph
## Testing
- Added `EventSourcedStateSpec` which tests a mock ledger
## Tickets
_* closes #1894_


event-tree/src/test/scala/co/topl/eventtree/EventSourcedStateSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+    val deltaStore = RefStore.Eval.make[F, LedgerUnapply]().value
+    val tree = ParentChildTree.FromRef.make[F, TypedIdentifier].value
+
+    txData
+      .foldLeftM[F, Unit](()) { case (_, (txId, from, amount, to)) =>
+        eventStore.put(txId, Tx(txId, (from, amount), to))
+      }
+      .value
+
+    txAssociations.foldLeftM[F, Unit](()) { case (_, (c, p)) => tree.associate(c.asTxId, p.asTxId) }.value
```
scasplte2, line 43: This separation makes me think about possible issues that may arise from non-atomic updates. It isn't exactly the case here but could the stores get out of sync if one them is updated before the others?

scasplte2 (approved):
👍 just the one comment but thinking on it a bit more, I think I am misunderstanding something so if we can clear that up then I'm happy to merge

PR #1990, opened 2022-03-10
Genesis config

jingsting:
## Purpose
We are adding options to generate genesis blocks from block Json files, and generating genesis block for different networks, not just for private testnet (generated addresses with funds will be left unlocked)

## Approach
- Added genesisFromBlockJson under genesisProvider
- Added generatedGenesis under genesisProvider
- Updated app settings for new functions
- Removed PrivateTestnet settings in forging settings
- Moved genesisBlock() out of forger and into a separate file
- Added genesis block Json files for Toplnet, Valhalla, and Hel

## Testing
- Add tests in GenesisSpec to make sure the genesis block generated from block Json are the same as the ones from previous genesisProviders

## Tickets
* #1837
* #1947


node/src/test/scala/co/topl/consensus/ConsenesusVariablesSpec.scala
```diff
@@ -34,9 +34,12 @@ class ConsenesusVariablesSpec
 
   behavior of "ConsensusStorage"
```
scasplte2, line 35: lets rename this to ``ConsensusInterfaceSpec`

scasplte2: or `ConsensusViewerSpec`

scasplte2 (approved):
nicely done @jingsting

PR #1993, opened 2022-03-11
Tetra Eligibility Simulator

SeanCheatham:
## Purpose
- Extracting runtime data about Tetra's leader election mechanism is slow if using TetraDemo
  - TetraDemo uses a clock/delay mechanism which requires waiting for the completion of each slot
- The new EligibilitySimulator runs almost the same protocol as TetraDemo
  - The simulator does not delay between slots
  - Header validation is disabled
  - Operational Keys uses an in-memory secure store; no disk persistence
- Also add a new ThresholdSimulator which generates threshold values under various configurations
## Approach
- Create new eligibilitySimulator SBT module
- Copy/paste TetraDemo.  Updated to reflect changes described above

## Testing
N/A
## Tickets
_* closes #1943_


scasplte2 (approved):
🚀

PR #1995, opened 2022-03-14
Genus Query Sorting

### LINKED ISSUE #1867: Genus - Add Sorting To Queries
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1052) by [Unito](https://www.unito.io)

b-vennes:
## Purpose
- Add option to sort response data in a Genus query using some pre-selected set of fields.

## Approach
- Adds a new parameter to the grpc service requests
- Implements a set of sorting options using Mongo BSON
- Removes default values for filtering and sorting
  - The default value is now determined when converting a grpc query request into a query interpreter input

## Testing
- Fixes some tests broken by type changes

## Tickets
- Closes #1867


genus/src/main/scala/co/topl/genus/ops/protobufops/TxsQueryStreamReqOps.scala
```diff
@@ -2,15 +2,20 @@ package co.topl.genus.ops.protobufops
...
 import scala.language.implicitConversions
 
-final class TxsQueryStreamReqOps(val value: TxsQueryStreamReq) extends AnyVal {
+final class TxsQueryStreamReqOps(private val value: TxsQueryStreamReq) extends AnyVal {
 
-  def toQueryRequest: QueryRequest[TransactionFilter, Bson] =
-    QueryRequest(value.filter, None, None, value.confirmationDepth)
+  def toQueryRequest: QueryRequest[TransactionFilter, TransactionSorting] =
+    QueryRequest(
+      value.filter.getOrElse(TransactionFilter(TransactionFilter.FilterType.Empty)),
```
scasplte2, line 14: it seems like you have to have this `.getOrElse` in several places. Is there any way to avoid this? I realize these are are the query level so maybe not. Just wondering since they are all QueryRequest and it feels a little repetitive. But this may be the most appropriate way to do it so just let me know your thoughts.

genus/src/main/scala/co/topl/genus/ops/protobufops/TxsSubscriptionResCompanionOps.scala
```diff
@@ -9,7 +9,7 @@ import co.topl.genus.types.Transaction
 
 import scala.language.implicitConversions
 
-final class TxsSubscriptionResCompanionOps(val companion: TxsSubscriptionRes.type) extends AnyVal {
+final class TxsSubscriptionResCompanionOps(private val companion: TxsSubscriptionRes.type) extends AnyVal {
```
scasplte2, line 12: several places have these as private members now, what was the motivation behind this? Anything more than better defining the scope and usage?

scasplte2 (approved):
🪨 well done, just a couple of questions but looks good to me

genus/src/main/scala/co/topl/genus/ops/protobufops/TxsSubscriptionResCompanionOps.scala
b-vennes: I realized that leaving the value of an implicit class used for defining an extension method also adds the value as a parameter.

I'm not sure if this link will still be accurate, but I defined this program which demonstrates this issue:
[https://scastie.scala-lang.org/SIguGF92T6GSKb44gbf7BQ](https://scastie.scala-lang.org/SIguGF92T6GSKb44gbf7BQ)

genus/src/main/scala/co/topl/genus/ops/protobufops/TxsQueryStreamReqOps.scala
b-vennes: The primary reason for this is that any non-primitive field in protobuf will be of type Option[T] instead of just T with a default. This means that its the job of the server/client to decide what the default should be.

I do believe there is a `valueOrDefault` option which effectively does the same thing as `.getOrElse(...Empty)`. I was thinking it might be better to be specific about what the default value is, but I think its also a fair assumption for the default to always  be the `Empty` case.

scasplte2: Reasonable, we can keep that in mind for the future if it comes up. Thanks for the explanation.

genus/src/main/scala/co/topl/genus/ops/protobufops/TxsSubscriptionResCompanionOps.scala
scasplte2: That is a good idea and thank you for linking an example. Could you put this code and a little explainer on a page under Scala Learning in Confluence?
https://topl.atlassian.net/wiki/spaces/CORE/pages/363855942/Scala+Learning

scasplte2: [CORE-ExtensionOperations-150322-1837.pdf](https://github.com/Topl/Bifrost/files/8256234/CORE-ExtensionOperations-150322-1837.pdf)

☝️ copy of document linked above for those without access to Confluence

PR #1999, opened 2022-03-14
Typed Protocols

SeanCheatham:
## Purpose
- Typed Protocols is a concept borrowed from Cardano's network implementation
  - It is a specialized state machine
  - It is a two-party state protocol, meaning the two participating parties share the same protocol state
  - Only one party has agency at any given state of the protocol, meaning only one party can send a message
- This PR adds the basic abstraction for a typed protocol, as well as a subset of the protocols we will need to implement networking
- This PR also moves SlotData from `consensus` to `models`
  - **Note:** A majority of the files changed in this PR are related to this refactor

## Approach
- Define an abstraction of the type system required to execute a typed protocol
- Introduce a few basic building blocks for common typed protocols
- Implemented BlockNotification, TransactionNotification, SlotDataRequestResponse, BlockHeaderRequestResponse, BlockBodyRequestResponse, TransactionRequestResponse, and BlockHeightRequestResponse

## Testing
- Added basic test cases for the two classifications of typed protocol
- Included basic example/demo typed protocols

## Tickets
_* closes #1954_


consensus/src/main/scala/co/topl/consensus/package.scala
```diff
@@ -22,4 +24,19 @@ package object consensus {
 
     def stringToCacheKey(key: String): String = key
   }
+
+  implicit class BlockHeaderV2Ops(blockHeaderV2: BlockHeaderV2) {
```
scasplte2, line 28: nice 👍 this is a convenient conversion

networking/src/main/scala/co/topl/networking/typedprotocols/TypedProtocolInstance.scala
```diff
@@ -0,0 +1,109 @@
+package co.topl.networking.typedprotocols
```
scasplte2, line 1: the type manipulation in this file feels like magic 🪄

scasplte2 (approved):
nice work here @SeanCheatham 🎉

scasplte2 (approved):
🔥

PR #2000, opened 2022-03-15
Tetra async program

SeanCheatham:
## Purpose
- The current DemoProgram implementation iterates over each slot and invokes the available minters to attempt to mint a block
  - The slot-by-slot approach is really only relevant in the context of minting new blocks
  - The program itself only needs to supervise the minters, but it doesn't need to control their clocks
  - The program will eventually be responsible for accepting asynchronous information from the network, and a synchronous clock will not permit that behavior

## Approach
- Add a new PerpetualMintAlgebra which emits a stream of blocks
- Add a new interpreter which wraps the existing BlockMint and produces attempts in a streamed fashion
- Updated the DemoProgram to run the block streams from each perpetual minter

## Testing
- Ran TetraDemo program locally and witnessed blocks minted, even across epoch boundaries

## Tickets
_* closes #1998_


demo/src/main/scala/co/topl/demo/DemoProgram.scala
```diff
@@ -1,183 +1,76 @@
...
-        if (localChainIsWorseThan)
-          EitherT(headerValidation.validate(nextBlock.headerV2, canonicalHead))
+      _ <- Monad[F].ifElseM(
+        localChainIsWorseThan.pure[F] ->
+        Sync[F].defer(
+          EitherT(
+            OptionT(blockStore.get(nextBlock.headerV2.parentHeaderId))
+              .getOrElseF(MonadThrow[F].raiseError(new NoSuchElementException(nextBlock.headerV2.parentHeaderId.show)))
+              .flatMap(parent => headerValidation.validate(nextBlock.headerV2, parent.headerV2))
+          )
```
scasplte2, line 73: What are your thoughts about having multiple adoptions? it might make the storage and retrieval of data a little more difficult but we could keep this idea here of validating and adopting the header. Similar to how the adoption of a body (i.e. the final portion of a block) will trigger the notification protocols to notify of the newly adopt block id. Then adoption of a header could trigger the corresponding message to request the transaction id's. 

Just an idea, and I am not sure how to deal with adopting headers that may not have a valid body but I like that there seems to be a symmetry in handling the requests

scasplte2 (approved):
🏃 🎽 🏃‍♂️

demo/src/main/scala/co/topl/demo/DemoProgram.scala
SeanCheatham: The storage abstraction I've been running with is a `Store[F, T]` where `T` is the value type being stored.  There are separate `Store[F, BlockHeaderV2]` and `Store[F, BlockBodyV2]`.  (The Block Body store will probably change to store a list of Transaction IDs, just like in the typed protocols).  And eventually there will be `Store[F, Transaction]`.

We can store the data as we receive it.  We may receive a header that we validate, and then subsequently receive a body which is invalid.  In this case, we would just remove the header from its store and penalize the peer who told us about it.

PR #2012, opened 2022-03-24
Tetra Basic Networking

SeanCheatham:
## Purpose
- Tetra nodes need the ability to connect to other Tetra nodes to exchange blocks
  - The connection should be over the network
  - A node can accept incoming connections
  - A node initiates outbound connections
- For each established connection,
  - A multiplexer should aggregate several communication channels over a single connection
  - Each communication channel should be backed by a Typed Protocol
  - The local node should listen to remote-adopted block IDs, fetch each missing header, and fetch each missing body.  Once a remote block has been retrieved, it should run through the chain selection, validation, and adoption mechanism
## Approach
- Introduce the Multiplexer from a recent akka-stream workshop
  - NOTE: Also introduced a DynamicMultiplexer which is not yet needed and is not as robust as the basic Multiplexer
- Introduce P2PServer and AkkaP2PServer to receive and initiate peer connections
- Create helpers/factories for transforming a TypedProtocol into an akka stream flow
- Modify DemoProgram to launch the AkkaP2PServer and fetch blocks for each remote connection
- Modify command-line arguments to allow for creating a deterministic set of genesis stakers, also allow for defining remote peers
## Testing
- Created test suites for multiplexer
- Created test suites for ConnectionLeaderFlow
- NOTE: There is still more testing to do, specifically around running TypedProtocols within a Multiplexer
## Tickets
_* closes #2001_
_* closes #2010_
_* closes #2011_


byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
```diff
@@ -12,6 +12,16 @@ import scala.collection.immutable.ListMap
 import scala.reflect.ClassTag
 
 trait ValuetypesCodecs {
+
+  def emptyCodec[T](instance: T): Codec[T] =
```
b-vennes, line 16: what do we use this for?

cats-akka/src/main/scala/co/topl/catsakka/RunnableGraphOps.scala
```diff
@@ -0,0 +1,23 @@
...
+import cats.arrow.FunctionK
+import cats.effect.kernel.Sync
+import cats.~>
+
+trait RunnableGraphOps {
+
+  implicit def runnableGraphToF[F[_]: Sync](implicit mat: Materializer): RunnableGraph ~> F =
+    FunctionK.liftFunction(rg => Sync[F].delay(rg.run()))
+
+  implicit class RunnableGraphSupport[Mat](runnableGraph: RunnableGraph[Mat]) {
```
b-vennes, line 15: Is `implicit class` preferred to creating a `class` alongside an implicit conversion from the value to the class?

demo/src/main/scala/co/topl/demo/DemoProgram.scala
```diff
@@ -51,40 +87,51 @@ object DemoProgram {
...
-  ): F[Unit] =
+  ): F[Boolean] =
     for {
-      _                     <- Logger[F].info(show"Minted block ${nextBlock.headerV2}")
-      _                     <- blockStore.put(nextBlock.headerV2.id, nextBlock)
-      slotData              <- ed25519VrfResource.use(implicit ed25519Vrf => nextBlock.headerV2.slotData.pure[F])
-      localChainIsWorseThan <- localChain.isWorseThan(slotData)
-      _ <- Monad[F].ifElseM(
-        localChainIsWorseThan.pure[F] ->
+      _ <- OptionT(headerStore.get(block.headerV2.id))
```
b-vennes, line 99: Why do you want to check that the block id exists before updating it in the store?

demo/src/main/scala/co/topl/demo/DemoProgram.scala
```diff
@@ -51,40 +87,51 @@ object DemoProgram {
...
         )
       )(
-        Logger[F].info(show"Ignoring weaker block header id=${nextBlock.headerV2.id}")
+        Sync[F]
+          .defer(Logger[F].info(show"Ignoring weaker block header id=${block.headerV2.id.asTypedBytes}"))
+          .as(false)
       )
-    } yield ()
+      a = adopted
+    } yield a
```
b-vennes, line 135: We could probably yield `adopted` directly here.

networking/src/main/scala/co/topl/networking/blockchain/BlockchainMultiplexerCodecs.scala
```diff
@@ -0,0 +1,95 @@
...
+import co.topl.networking.multiplexer.MultiplexerCodecBuilder
+import co.topl.networking.blockchain.NetworkTypeTags._
+import co.topl.networking.typedprotocols.TypedProtocol
+import scodec.{Attempt, Codec, DecodeResult, Decoder, Encoder, Err}
+import scodec.bits.BitVector
+
+object BlockchainMultiplexerCodecs {
+
+  implicit val commonMessagesStartTransmittable: Transmittable[TypedProtocol.CommonMessages.Start.type] =
+    Transmittable.instanceFromCodec(emptyCodec(TypedProtocol.CommonMessages.Start))
```
b-vennes, line 18: I see why we need the `emptyCodec` now. Is it just so that we can have the `Start` message without needing to send any data?

networking/src/main/scala/co/topl/networking/p2p/ConnectionLeaderFlow.scala
```diff
@@ -0,0 +1,156 @@
...
+  def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Mat]) = {
+    val matPromise: Promise[Mat] = Promise()
+    val logic =
+      new GraphStageLogic(shape) with StageLogging {
+
+        def setUninitializedHandlers(): Unit = {
+          var pullOnInit = false
+          var localNumberSent = false
+          var remoteNumber = none[Int]
+          val localNumber = scala.util.Random.nextInt() // TODO Secure Random
```
b-vennes, line 55: Couldn't an adversary just always use `Int.max` or `Int.min` to select if they want to go first?

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
SeanCheatham: Added a scaladoc comment giving a bit of context.

You already figured it out later in the PR, but for others reading: `emptyCodec` is meant for representing a serialized object that doesn't really need to serialize any data.  For example, a case object.  In the MultiplexerCodec, a discriminator byte is already embedded and consumed in the byte data indicating the type of information serialized.

demo/src/main/scala/co/topl/demo/DemoProgram.scala
SeanCheatham: If the value already exists in the database, I can avoid the penalty of serializing the value.

I think it makes more sense if I do a `.contains` check instead of a `.get` call.  Doing a `.contains` simply checks if the key exists in the database, while `.get` will decode the value as well.  So with a .get, I'm just trading one penalty for another.

I added `.contains` to this PR after your comment

networking/src/main/scala/co/topl/networking/blockchain/BlockchainMultiplexerCodecs.scala
SeanCheatham: Yup that's correct!  The typed protocol helper already inserts the type prefix byte, so we don't need any further information to decode these simple case objects like `Start`

networking/src/main/scala/co/topl/networking/p2p/ConnectionLeaderFlow.scala
SeanCheatham: Yes :100: .  There's a comment up near the top that outlines a TODO plan for improving the implementation that involves exchanging the hash of a number before the actual number.

cats-akka/src/main/scala/co/topl/catsakka/RunnableGraphOps.scala
b-vennes: > I think I could improve things slightly by having the implicit class extend AnyVal though

Are implicit classes allowed to extend AnyVal? if so, using implicit classes might be the best pattern going forward.

Whatever we decide, I'll be sure to update my doc here and update any usages of implicit classes that I see: https://topl.atlassian.net/wiki/spaces/CORE/pages/389349389/Extension+Operations

demo/src/main/scala/co/topl/demo/DemoProgram.scala
```diff
@@ -51,40 +88,51 @@ object DemoProgram {
...
-          EitherT(
-            OptionT(blockStore.get(nextBlock.headerV2.parentHeaderId))
-              .getOrElseF(MonadThrow[F].raiseError(new NoSuchElementException(nextBlock.headerV2.parentHeaderId.show)))
-              .flatMap(parent => headerValidation.validate(nextBlock.headerV2, parent.headerV2))
+      _ <- headerStore
+        .contains(block.headerV2.id)
+        .ifM(Applicative[F].unit, headerStore.put(block.headerV2.id, block.headerV2))
+      _ <- bodyStore
+        .contains(block.headerV2.id)
+        .ifM(Applicative[F].unit, bodyStore.put(block.headerV2.id, block.blockBodyV2))
```
b-vennes, line 105: could you explain again what `ifM` does?

b-vennes: Actually, I think I see. Does `ifM` only work on a `M[Bool]` type where `M` is a `Monad`. And it runs the left case when `true` and the right case when `false`?

SeanCheatham: Yup that's correct!  It's one of my new favorites

networking/src/main/scala/co/topl/networking/p2p/ConnectionLeaderFlow.scala
SeanCheatham: I updated the PR with the new implementation which uses the hash-based approach.  It's not really the integer values themselves that get compared; it's a mechanism that hashes the concatenation of the bytes of the integer values and then turns the hashes into BigInt values.  Because the comparison requires hashing values from both parties, choosing Int.MaxValue or Int.MinValue doesn't necessarily benefit the adversary.

I think there's still definitely room for improvement for sure.  For now, there's not really much of a benefit to an adversary if they choose to be the connection leader or not because all of our typed protocols are reciprocated pairs (they come in client/server pairs).

cats-akka/src/main/scala/co/topl/catsakka/RunnableGraphOps.scala
SeanCheatham: An implicit class may extend AnyVal, but in implementing the change in this particular file, I realized the compiler does not let you define a nested class that extends AnyVal.  Meaning you can't have
```
trait Foo {
  class Bar(val x: Int) extends AnyVal
}
 ```
 
 That `class Bar` needs to be moved out of the `trait`, but doing that means you can no longer use `implicit class` because an `implicit class` has the opposite requirement; it needs to be defined inside of another class.

So with all that said, I think I understand now why people use the more verbose separation of `class ... extends AnyVal` from `implicit def`.  I think you can leave your document as-is since it's probably the more useful approach.

scasplte2 (approved):
🐙 <> :octocat:

PR #2016, opened 2022-03-30
Add test vectors for Curve25519 key generation

### LINKED ISSUE #1989: BN-481 ⁃ From entropy test vectors - Bifrost
### Statement of Problem
Generating a Curve25519 key pair should be standardized across implementations in different languages to ensure portability of keys and deterministic regeneration of the key pair from the same given entropy. Including a standard set of test vectors will ensure implementations in other languages (JavaScript, Python, etc.) will produce the same keys.

### Acceptance Criteria
* Create or use a preexisting set of test vectors for Curve25519 keys
* Implement a set of unit tests to fully test Curve25519 key generation using given entropy

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/BN-481) by [Unito](https://www.unito.io)

Tuxman:
## Purpose
Generating Curve25519 public/private key pairs requires deterministic output from a given seed. The generation of this seed should be standardized across implementations using a set specification. We have chosen to use a modified version of [SLIP-0023](https://github.com/satoshilabs/slips/blob/master/slip-0023.md)  to derive the entropy of Curve25519 keys using the same key derivation scheme as ED25519 keys. The KDF and generation of Curve25519 keys is already in Bifrost, so this PR is adding unit tests to verify generating a seed from a given entropy. The test vectors used here will be placed in Topl's [Reference Crypto](https://github.com/Topl/reference_crypto) repo to define the specification for implementations in other languages and libraries, namely [BramblJS](https://github.com/Topl/BramblJS) and [BramblPy](https://github.com/Topl/BramblPy).

## Testing
* 

## Tickets
* closes #1989


b-vennes:
Instead of moving `mnemonic` into the new `generation` package, would it be more clear to instead rename `mnemonic` to `generation`?

crypto/src/test/resources/Curve25519AxolotlTestVectors.json
```diff
@@ -0,0 +1,121 @@
+[  {
```
b-vennes, line 1: do we ever use these test vectors?

b-vennes (approved):
Looks good! But I had one suggestion with the `TestVector` abstract class.

scasplte2 (changes requested):
I would like this to be better organized, there is no need for you to do anything @Tuxman. I just wanted to block this PR from being merged as I am not fully satisfied with where it is at but I will take care of getting it there.

crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
```diff
@@ -227,7 +215,7 @@ class ExtendedEd25519
 
 object ExtendedEd25519 {
 
-  def precomputed(): ExtendedEd25519 = {
+  lazy val instance: ExtendedEd25519 = {
```
SeanCheatham, line 218: If I remember correctly, `ExtendedEd25519` is not thread safe, so I don't think we want to include a shared companion object instance

crypto/src/test/scala/co/topl/crypto/utils/Generators.scala
```diff
@@ -23,4 +24,12 @@ object Generators {
 
   lazy val stringGen: Gen[String] = Gen.alphaNumStr.suchThat(_.nonEmpty)
 
+  def pickMnemonicSize(int: Int): MnemonicSize = int % 5 match {
```
SeanCheatham, line 27: This could probably be replaced by a `Gen[MnemonicSize]` if I understand its usage correctly.  i.e. 

```

val mnemonicSizeGen: Gen[MnemonicSize] =
  Gen.oneOf(MnemonicSizes.`12`, MnemonicSizes.`15`, MnemonicSizes.`18`, MnemonicSizes.`21`, MnemonicSizes.`24`)
  
  ```

SeanCheatham (approved):

crypto/src/test/scala/co/topl/crypto/utils/Generators.scala
scasplte2: thanks for the suggestion @SeanCheatham

scasplte2 (approved):
🔍

PR #2056, opened 2022-04-21
Genus X-Api-Key Authentication

### LINKED ISSUE #1864: Token-Based Authorization System
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1049) by [Unito](https://www.unito.io)

b-vennes:
## Purpose

Implementation of the X-Api-Key HTTP header protocol is necessary for basic security and restriction of Genus services running in production. The HTTP header `x-api-key` must be provided if required by the server and must hash to a value stored in the service.

## Approach

Implements `x-api-key` authentication using the `optionalHeaderValueByName` Akka directive, which secures access to a `Route` unless a specific header is provided. 

## Testing

- no new tests are added

## Tickets

- closes #1864


genus/src/main/resources/application.conf
```diff
@@ -11,6 +11,10 @@ genus {
 
       port = 9089
 
+      disableAuth = false
```
scasplte2, line 14: @b-vennes could you add some comments describing the settings in the `application.conf`. Some are obvious but in particular you should at least give the expected format for the `apiKeyHash` as the Base58 encoded hash output resulting from executing the Blake2b256 hash function on the the UTF-8 encoded API key (or something like that)

scasplte2 (changes requested):

SeanCheatham:
Were you able to test this out manually in some way?  I'm curious how things look from the client-side of akka-grpc (or other gRPC frameworks) when attaching headers

genus/src/main/resources/application.conf
scasplte2: @b-vennes any movement here?

b-vennes: I've just added the comments

b-vennes:
> Were you able to test this out manually in some way? I'm curious how things look from the client-side of akka-grpc (or other gRPC frameworks) when attaching headers

I tested the headers using the `grpcurl` tool, but I've not yet tried using a generated client.

b-vennes:
> Were you able to test this out manually in some way? I'm curious how things look from the client-side of akka-grpc (or other gRPC frameworks) when attaching headers

I just tested and it is possible to set headers via the akka-grpc library. We will definitely want to provide some examples, and probably want to bake it into any official clients we create.

b-vennes:
@scasplte2 I have made the requested changes, so this should be ready for review again

SeanCheatham (approved):

scasplte2 (approved):

scasplte2 (changes requested):
@b-vennes please ping the tech core channel once the conflicts are resolved but otherwise 👍

scasplte2 (approved):
👍

PR #2058, opened 2022-04-21
Signable Instance Of Tetra Unproven Transaction

### LINKED ISSUE #1945: Dion-Compatible Tetra Transaction Proofs
### Statement of Problem
Proofs should be created on Tetra transactions such that when the Tetra transaction is converted into a dion transaction, that it can be verified correctly.

### Acceptance Criteria
* Attestations can be created for Tetra transactions

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-1099) by [Unito](https://www.unito.io)

b-vennes:
## Purpose

Generating a Dion-compatible valid signature for a Tetra Transaction requires an instance of the `Signable` typeclass. The instance must generate the same bytes for a Tetra transaction as the equivalent Transaction.TX instance.

## Approach

In order to generate a signable instance, multiple new codecs are added in the `co.topl.codecs.binary.typeclasses.SignableInstances.SignableCodecs` object. These codecs are intended for maintaining compatibility with Dion signature generation and are not 100% compatible with the other codecs that currently exist in bifrost.

## Testing

Unit tests were added which test the correctness of the bytes generated by the signable instance for various types of transactions.

## Tickets

- closes #1945


scasplte2 (approved):
I think there could be some duplication of codecs in `SignableInstances` so we may want to come back to these at some point

PR #2061, opened 2022-04-22
Resolve Warnings

mgrand-topl:
See CORE-1143


mgrand-topl:
@scasplte2 There are still two warnings that occur during the build. The message is
> Bifrost/crypto/src/main/scala/co/topl/crypto/accumulators/merkle/MerkleTree.scala:87:35: value WrappedArray in package mutable is deprecated (since 2.13.0): Use ArraySeq instead of WrappedArray; it can represent both, boxed and unboxed arrays
Because `ArraySeq` is not available in Scala 2.12 and we are currently supporting 2.12 and 2.13, I cannot resolve this by making the suggested change. When we move to Scala 3, we will need to change our approach as WrappedArray does not exist in Scala 3

I see two way to resolve this. My preference would be to stop supporting multiple versions of Scala for Bifrost. If we no longer support Scala 2.12 then we can make the recommended change. Support for multiple versions of Scala should only be necessary for libraries, not for executable binaries.

If there is a reason that we want to support multiple versions, then before we support Scala 3 we will need to have multiple versions of the code specific to different versions of Scala.

b-vennes:
> value WrappedArray in package mutable is deprecated (since 2.13.0): Use ArraySeq instead of WrappedArray

These are both really unusual types. I wonder if it might be preferable to test it with `List` or `Chain` instead.

b-vennes (approved):

b-vennes (approved):

PR #2091, opened 2022-04-27
Core-1183: Tetra Transaction Codecs

SeanCheatham:
## Purpose
- Tetra needs the ability to serialize and deserialize a Transaction for networking and storage purposes
- Serializing a Transaction also involves serializing propositions, proofs, and boxes
## Approach
- Added codecs for Propositions, Proofs, Boxes, and Transactions
- Separated Tetra Scodec Codecs into multiple traits by Group
- Removed typed parameter from Box model for simplification

## Testing
- Added several new ModelGenerators
- Added codec tests for the new codecs

## Tickets
_* closes #2081_

### DIFF
--- a/byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
+++ b/byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
@@ -1,5 +1,6 @@
 package co.topl.codecs.bytes.scodecs.valuetypes
 
+import cats.data.NonEmptyChain
 import cats.implicits._
 import co.topl.codecs.bytes.ZigZagEncoder._
 import co.topl.codecs.bytes.scodecs.valuetypes.Constants._
@@ -134,6 +135,12 @@ trait ValuetypesCodecs {
 
   implicit def seqCodec[T: Codec]: Codec[Seq[T]] =
     listCodec[T].xmap(list => list.toSeq, seq => seq.toList)
+
+  implicit def nonEmptyChainCodec[T: Codec]: Codec[NonEmptyChain[T]] =
+    listCodec[T].exmap(
+      list => Attempt.fromOption(NonEmptyChain.fromSeq(list), Err("Expected non-empty seq, but empty found")),
+      seq => Attempt.successful(seq.toList)
+    )
 
   implicit def arrayCodec[T: Codec: ClassTag]: Codec[Array[T]] =
     listCodec[T].xmap(list => list.toArray, array => array.toList)
--- a/byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
+++ b/byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
@@ -2,17 +2,38 @@ package co.topl.codecs.bytes.scodecs.valuetypes
 
 import cats.data.NonEmptyChain
 import cats.implicits._
+import cats.{Monad, Monoid}
 import co.topl.codecs.bytes.ZigZagEncoder._
 import co.topl.codecs.bytes.scodecs.valuetypes.Constants._
 import co.topl.codecs.bytes.scodecs.valuetypes.Types._
 import scodec.bits.BitVector
 import scodec.{Attempt, Codec, DecodeResult, Err, SizeBound}
 
+import scala.annotation.tailrec
 import scala.collection.SortedSet
-import scala.collection.immutable.ListMap
+import scala.collection.immutable.{ListMap, ListSet}
 import scala.reflect.ClassTag
+import scala.util.Try
 
 trait ValuetypesCodecs {
+
+  implicit private val bitVectorMonoid: Monoid[BitVector] =
+    Monoid.instance(BitVector.empty, _ ++ _)
+
+  implicit val attemptMonad: Monad[Attempt] =
+    new Monad[Attempt] {
+      def flatMap[A, B](fa: Attempt[A])(f: A => Attempt[B]): Attempt[B] = fa.flatMap(f)
+
+      @tailrec
+      def tailRecM[A, B](a: A)(f: A => Attempt[Either[A, B]]): Attempt[B] =
+        f(a) match {
+          case Attempt.Successful(Left(a))  => tailRecM(a)(f)
+          case Attempt.Successful(Right(b)) => Attempt.successful(b)
+          case Attempt.Failure(cause)       => Attempt.Failure(cause)
+        }
+
+      def pure[A](x: A): Attempt[A] = Attempt.successful(x)
+    }
 
   /**
    * A valuetype codec which encodes zero bytes and decodes zero bytes.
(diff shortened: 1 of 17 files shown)


byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
```diff
@@ -98,54 +120,109 @@ trait ValuetypesCodecs {
...
+                  throw new Exception(cause.messageWithContext)
+              }
+            }
+          )
+        )
+        .map(DecodeResult(_, remaining))
+    }
   }
 
+  def sizedSeqCodec[T: Codec](size: Int): Codec[Seq[T]] = new Codec[Seq[T]] {
```
b-vennes, line 157: Could this use the sized array codec?

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
```diff
@@ -81,6 +100,29 @@ trait TetraScodecCodecs {
       }
     )
 
+  implicit val networkPrefixCodec: Codec[NetworkPrefix] =
+    Codec[Byte].xmapc(b => NetworkPrefix(b))(_.value)
```
b-vennes, line 104: I think you can do `Codec[Byte].as[NetworkPrefix]`.

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
```diff
@@ -81,6 +100,29 @@ trait TetraScodecCodecs {
...
+
+  implicit val typedEvidenceCodec: Codec[TypedEvidence] =
+    (Codec[TypePrefix] :: Codec[Evidence])
```
b-vennes, line 107: I think this could be `(Codec[TypePrefix] :: Codec[Evidence]).as[TypedEvidence]`

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
```diff
@@ -81,6 +100,29 @@ trait TetraScodecCodecs {
...
+        HList(t.networkPrefix, t.typedEvidence)
+      )
+
+  implicit val rhoCodec: Codec[Rho] =
+    Codec[Sized.Strict[Bytes, Lengths.`64`.type]].xmap(Rho(_), _.sizedBytes)
+
+}
+
+trait TetraScodecCryptoCodecs {
+  self: TetraScodecPrimitiveCodecs =>
```
b-vennes, line 124: Could you explain what this is doing?

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
```diff
@@ -294,7 +573,15 @@ trait TetraScodecCodecs {
...
+    (Codec[Slot](uLongCodec) :: Codec[TypedIdentifier])
+      .xmapc { case slot :: id :: HNil => SlotId(slot, id) }(t => HList(t.slot, t.blockId))
+
+  implicit val slotDataCodec: Codec[SlotData] =
+    (Codec[SlotId] :: Codec[SlotId] :: Codec[Rho] :: Codec[Eta] :: Codec[Long](uLongCodec))
+      .xmapc { case slotId :: parentSlotId :: rho :: eta :: height :: HNil =>
+        SlotData(slotId, parentSlotId, rho, eta, height)
+      }(t => HList(t.slotId, t.parentSlotId, t.rho, t.eta, t.height))
 
   implicit val unsignedBlockHeaderV2Codec: Codec[BlockHeaderV2.Unsigned] =
```
b-vennes, line 586: I think you could also use the `.as[BLockHeaderV2.Unsigned]` function here since the sub-types are in the same order as in the case class.

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
```diff
@@ -1,18 +1,40 @@
...
+import scala.collection.immutable.{ListMap, ListSet}
 import scala.reflect.ClassTag
+import scala.util.Try
 
 trait ValuetypesCodecs {
 
+  implicit private val bitVectorMonoid: Monoid[BitVector] =
+    Monoid.instance(BitVector.empty, _ ++ _)
+
+  implicit val attemptMonad: Monad[Attempt] =
```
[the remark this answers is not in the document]
SeanCheatham: :+1: Nice - I added the dependency and removed my implementations

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
SeanCheatham: As a side-note: I replaced the `sizedListCodec` with `sizedSeqCodec` because the List codec assembled the list one element at a time through an append operation (which Lists don't do very well).

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
SeanCheatham: Per your suggestion, I was able to incorporate the `.as` syntax in a bunch of places in this file which reduced the amount of code significantly.  Unfortunately, for `@newtype`s like this, it wasn't allowing me to use that `.as` syntax so I continued the `.xmap` approach in those cases.

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
b-vennes: I'll be sure to use this from now on when I need access to certain typeclass instances, but don't want to have to manage imports from the same package.

b-vennes (approved):

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
```diff
@@ -98,54 +101,112 @@ trait ValuetypesCodecs {
...
-    override def encode(value: List[T]): Attempt[BitVector] =
-      (0 until size).foldLeft(Attempt.successful(BitVector.empty)) {
-        case (Attempt.Successful(bits), index) =>
-          for {
-            item     <- value.get(index).map(Attempt.successful).getOrElse(Attempt.failure(Err("invalid list size")))
-            itemBits <- Codec[T].encode(item)
-          } yield bits ++ itemBits
-        case (failure, _) => failure
-      }
+    override def encode(value: Array[T]): Attempt[BitVector] =
```
scasplte2, line 106: is the refactor here just to avoid the iteration from looping?

scasplte2: is this the same logic as replacing sizedListCodec ?

scasplte2 (approved):

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
SeanCheatham: Another note about the "known size" thing - we encode the length of the collection we're serializing so that we know how many items to expect when we deserialize.  `List#length` is also `O(n)` unfortunately.  There's not much we can really do about it since we only get the List's length when _encoding_, and we already have the List at that point.

SeanCheatham: To answer your question:
> is this the same logic as replacing sizedListCodec

It's different logic, but the resulting bytes should be the same as before.

Issue #1120, opened 2021-03-18
Create a "Known Peers" Command-Line Option


SeanCheatham:
The ability to specify a list of `knownPeers` as a command-line startup option would assist with integration testing efforts.  Test setups often involve dynamically allocated IP addresses, thus static addresses are difficult to use.

Acceptance Criteria:
- When BifrostApp is started with the flag `-p 1.2.3.4,5.6.7.8` or `--knownPeers 1.2.3.4,5.6.7.8`, the service starts with `knownPeers` containing `1.2.3.4` and `5.6.7.8`

Issue #1116, opened 2021-03-17
Environment Variable Configurations


SeanCheatham:
The current strategy for configuring a Bifrost application on startup is to provide pre-defined configuration files for each environment.  For fixed/static/known environments, this is helpful for sharing and connecting to the same specific networks.  For dynamically created environments (such as docker testing), this makes it tricky to configure individual instances for testing purposes.

Acceptance Criteria:

* Evaluate each of our configuration settings (in application.conf) to determine which ones make sense to allow environment variables.  At minimum include, `bifrost.rpcApi.bindAddress`, `bifrost.network.bindAddress`, and `bifrost.network.knownPeers`.
* For each of those settings, specify an optional environment variable override (in application.conf)
* The optional environment variable overrides do not interfere with command line arguments.  (i.e. Command-line arguments still take precedence)

Example:

```
bifrost.rpcApi.bindAddress = "0.0.0.0:9085"
bifrost.rpcApi.bindAddress = ${?BIFROST_RPC_API_BIND_ADDRESS}
```

This pattern will choose the environment variable if one exists; otherwise, it falls back to the static value.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/CORE-841) by [Unito](https://www.unito.io)

PR #2105, opened 2022-05-05
added cli option for known peers

jingsting:
## Purpose
We are adding the ability to specify a list of knownPeers, network address, and rpc api address as command-line startup options to assist integration testing efforts. This will make tests with dynamic IP addresses easier.

## Approach
Added `knownPeers` `networkBindAddress` `rpcBindAddress` options for cli arguments.
`knownPeers` supports multiple peer addresses separated with commas

## Testing
Tested running with the added arguments

## Tickets
* #1120 
* #1116

### DIFF
--- a/node/src/main/scala/co/topl/settings/StartupOpts.scala
+++ b/node/src/main/scala/co/topl/settings/StartupOpts.scala
@@ -115,8 +115,8 @@ final case class RuntimeOpts(
       .modify(configAddr =>
         networkBindAddress match {
           case Some(addrStr) =>
-              val split = addrStr.split(":")
-              new InetSocketAddress(split(0), split(1).toInt)
+            val split = addrStr.split(":")
+            new InetSocketAddress(split(0), split(1).toInt)
           case None => configAddr
         }
       )
--- a/node/src/main/scala/co/topl/settings/StartupOpts.scala
+++ b/node/src/main/scala/co/topl/settings/StartupOpts.scala
@@ -52,11 +52,12 @@ final case class RuntimeOpts(
       "If API key protection is enabled, this argument specifies the Blake2b256 hash of API key required by the JSON-RPC server "
   )
   apiKeyHash: Option[String] = None,
-  @arg(name = "knownPeers", short = 'k', doc = "List of IP addresses and ports of known peers, separated by commas")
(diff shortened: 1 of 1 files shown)


b-vennes (approved):

node/src/main/scala/co/topl/settings/StartupOpts.scala
```diff
@@ -49,7 +51,18 @@ final case class RuntimeOpts(
...
-  apiKeyHash: Option[String] = None
+  apiKeyHash: Option[String] = None,
+  @arg(
+    name = "knownPeers",
+    short = 'k',
+    doc = "List of IP addresses and ports of known peers, separated by commas. " +
+      "For example: 0.0.0.0:11,0.0.0.0:22"
+  )
+  knownPeers: Option[String] = None,
+  @arg(name = "networkBindAddress", short = 'a', doc = "Network address and port to bind to")
```
SeanCheatham, line 62: Should we use `p2pBindAddress` instead?  `networkBindAddress` was ambiguous to me at first glance; you only really know what it is once you realize the _other_ is `rpcBindAddress`.

SeanCheatham: `network` is already used throughout the codebase, so this wouldn't be the right time to change it.

node/src/main/scala/co/topl/settings/StartupOpts.scala
```diff
@@ -49,7 +51,18 @@ final case class RuntimeOpts(
     doc =
       "If API key protection is enabled, this argument specifies the Blake2b256 hash of API key required by the JSON-RPC server "
   )
-  apiKeyHash: Option[String] = None
+  apiKeyHash: Option[String] = None,
+  @arg(
+    name = "knownPeers",
+    short = 'k',
```
SeanCheatham, line 57: The ticket technically says the shortname should be `-p` instead of `-k`.  Does anyone have opinions between the two?

b-vennes: shouldn't `-p` be port?

scasplte2: @SeanCheatham I agree that `-p` seems more common

scasplte2 (approved):

PR #2109, opened 2022-05-10
Core-1191: Tetra Transaction Restructure

SeanCheatham:
## Purpose
- Refactors the existing Tetra transaction structure into the new model proposed in the [Tetra Transactions](https://topl.atlassian.net/wiki/spaces/CORE/pages/395215079/Transactions) design
  - Transaction Inputs include transactionId, transactionOutputIndex, proposition, proof, and the box value
  - Transaction fees are implicitly defined as the unclaimed inputs of the transaction

## Approach
- Update Tetra Transaction model.  Add `Transaction.Input`.  Replaced `Transaction.CoinOutput` with `Transaction.Output`
- Update binary and JSON codecs for the new model
- Partially updated the Tetra -> Dion converter, but left incomplete for now
  - https://topl.atlassian.net/browse/CORE-1215 
- Deleted the Dion -> Tetra converter
- Updated Brambl Credentials Playgrounds

## Testing
- Updated the ModelGenerators with new types

## Tickets
_* closes #2088_


byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
```diff
@@ -135,75 +138,60 @@ trait ValuetypesCodecs {
...
-                case Attempt.Failure(cause) =>
-                  throw new Exception(cause.messageWithContext)
-              }
-            }
-          )
-        )
-        .map(DecodeResult(_, remaining))
-    }
+    override def decode(bits: BitVector): Attempt[DecodeResult[Chain[T]]] =
+      (0 until size).foldLeft(Attempt.successful(DecodeResult(Chain.empty[T], bits))) {
```
b-vennes, line 151: did this code block get reverted?

b-vennes (approved):

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
SeanCheatham: I didn't revert it, but I updated it yet again to use `Chain` as the default base/primitive for serializing collections.  Now that Tetra Transactions directly use Chains, it made sense to implement them here.

models/src/main/scala/co/topl/models/Transaction.scala
```diff
@@ -1,34 +1,43 @@
...
+    inputs:    Chain[Transaction.Unproven.Input],
+    outputs:   NonEmptyChain[Transaction.Output],
+    timestamp: Timestamp,
+    data:      Option[TransactionData]
   )
 
+  object Unproven {
+
+    case class Input(
+      transactionId:          TypedIdentifier,
```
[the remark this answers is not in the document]
SeanCheatham: The primary goal here is to get us back to something a bit closer to "Unspent Transaction Outputs".  In Dion, the inputs of a transaction reference a Box.  But in Tetra, the inputs of a transaction will reference the outputs of a previous transaction (which subsequently point to boxes).

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
scasplte2: yes it looks like it @SeanCheatham do you think you could merge in `dev` changes or would that bork things for you?

models/src/main/scala/co/topl/models/Transaction.scala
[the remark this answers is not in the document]
scasplte2: additionally, we are also attempting to address an issue with generating box ids from evidence and nonce (where nonce is a 64 bit number). In a pure privacy preserving application layer (where we could assume SecretKeys are used only once) we probably wouldn't have had an issue because even if the nonce was repeated the evidence is always changing. However, in the extended-utxo contract situation we should expect evidence to be reused so then we fall back to a potential collision at 10^19 boxes (Long.MaxValue)

We can estimate the timescale for 10^19 boxes to exist by considering some numbers
- outputs per transaction: 100 (10^2)
- transactions per second: 100 (10^2)
- seconds per day: 86400 (~10^5)
- days per year: 365 (~10^2)

this gives us 10^11 so probabilistically we would've only had a collision after 10^8 years (which would've been fine) but I am concerned that this isn't enough security to mitigate an adversary grinding on transaction data (like the metadata or timestamp) to force a collision.

In this new method, we avoid arbitrarily reducing the space of available nonce outputs (as is done when computing box nonces in Dion by only taking the first 8 bytes of the transaction hash). This change made sense to me once I realized that we're just expanding the nonce from 8 bytes to 34 bytes (transaction id + index). So the space of possible nonce values is ~10^81. Given the above consideration of ~10^11 boxes created per year, then we still have a sizable margin of 10^70 possible nonce values, which I believe is a robust enough safety margin given current hardware.

typeclasses/src/main/scala/co/topl/typeclasses/ProofVerifier.scala
```diff
@@ -109,28 +109,22 @@ object ProofVerifier {
       context:     VerificationContext[F]
     ): F[Boolean] = {
       def compareBoxes(propositionBox: Box)(sourceBox: Box): Boolean = propositionBox match {
-        case Box(TypedEvidence.empty, 0, value) =>
+        case Box(TypedEvidence.empty, value) =>
```
scasplte2, line 112: @SeanCheatham we should reassess the `RequiredBoxState` proposition at some point. I'd like it to be able to reference a specific instance of a `Box` in addition to its spending logic and value. I don't think this needs to change box, but just how we provide arguments into the box state proposition.

typeclasses/src/main/scala/co/topl/typeclasses/TransactionOps.scala
```diff
@@ -11,28 +13,44 @@ object TransactionOps {
...
+          .Unproven(
+            transaction.inputs.map(i =>
+              Transaction.Unproven.Input(i.transactionId, i.transactionOutputIndex, i.proposition, i.value)
+            ),
+            transaction.outputs,
+            transaction.timestamp,
+            transaction.data
+          )
+
+      def unclaimedInputValues: Chain[Box.Value] = {
```
scasplte2, line 26: if we allow unclaimed arbits, I think we should figure out how to do unclaimed assets. Or if we want to hold off for now we could just support claiming polys for now and come back to the other two at a later date.

scasplte2 (approved):
overall looks solid ✅

byte-codecs/src/main/scala/co/topl/codecs/bytes/scodecs/valuetypes/ValuetypesCodecs.scala
SeanCheatham: Merged the other PR and incorporated the changes in this PR.  (I had to disable some more Tetra -> Dion Transaction unit tests)

typeclasses/src/main/scala/co/topl/typeclasses/TransactionOps.scala
SeanCheatham: :+1: I'll most likely change this implementation later anyway.  I added it when I initially attempted to update the Tetra -> Dion adapter to calculate Dion fees.

typeclasses/src/main/scala/co/topl/typeclasses/ProofVerifier.scala
SeanCheatham: Yea let's play with it once we implement the box state tracking functionality.  Hopefully that will clarify exactly what we need.

PR #2113, opened 2022-05-11
Genus - Added Scalacheck generators for data stored in mongo

b-vennes:
## Purpose

In order to make the tests for Genus more robust, especially around mocking the underlying Mongo collection, instances of the Scalacheck `Gen` typeclass should be created for the various data models used in Mongo. 

## Approach

Implemented `Gen` instances for several types:
- `BlockSummaryDataModel`
- `SimpleValueDataModel`
- `AssetValueDataModel`
- `TokenValueDataModel`
- `TokenBoxDataModel`
- `ConfirmedTransactionDataModel`
- `BlockDataModel`
- `Document` values representing encoded `ConfirmedTransactionDataModel` and `BlockDataModel` values

## Testing

No new tests have been added

## Tickets
* [Jira CORE-1217](https://topl.atlassian.net/browse/CORE-1217)


SeanCheatham (approved):
Totally up to you if you want to change the PR, but I've started to prefer implementing implicit `Arbitrary` instances rather than `Gen` instances.  You can convert between the two at any time, but since Arbitrary wants to be implicit, it's easier if it starts out that way.  When using a Gen in a test, you need to do so explicitly anyway.

scasplte2 (approved):

b-vennes:
> Totally up to you if you want to change the PR, but I've started to prefer implementing implicit `Arbitrary` instances rather than `Gen` instances. You can convert between the two at any time, but since Arbitrary wants to be implicit, it's easier if it starts out that way. When using a Gen in a test, you need to do so explicitly anyway.

I forgot that `Arbitrary` can support more cases. I'll change `Gen` to `Arbitrary` whenever it makes sense.

PR #2115, opened 2022-05-11
CORE-1211: Track-Latest Flow and Current Chain Height In Genus 

b-vennes:
## Purpose

Create a system that can retrieve the latest block height of a replicated chain in Mongo.

## Approach

- Implemented a new type of Akka Streams `Flow` that will keep track of the latest value that has passed through the stream at the point of the flow.
- Made use of the new `TrackLatestFlow` type combined with a continuous poll against an underlying blocks mongodb collection to determine the current height of the replicated chain

## Testing

- adds two tests for the `TrackLatestFlow`:
  - `it should materialize to a function which returns the latest value seen`
    - checks that the latest value seen at the point when the source closes is the expected value
  - `it should materialize to a function which returns the default value when no values seen`
    - checks that the default value gets returned if no values have been emitted to the stream

## Tickets
* closes [Jira CORE-1211](https://topl.atlassian.net/browse/CORE-1211)


genus/src/main/scala/co/topl/genus/flows/TrackLatestFlow.scala
```diff
@@ -0,0 +1,66 @@
...
+ */
+class TrackLatestFlow[T](initial: T) extends GraphStageWithMaterializedValue[FlowShape[T, T], () => T] {
+
+  private val inlet = Inlet[T]("RetriveLatestValue.Inlet")
+  private val outlet = Outlet[T]("RetrieveLatestValue.Outlet")
+
+  override def shape: FlowShape[T, T] = FlowShape(inlet, outlet)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, () => T) = {
+    var latestValue: T = initial
```
SeanCheatham, line 20: Thread safety probably isn't much of a concern in the context this is used, but if you wanted to make this thread-safe, you could using `getAsyncCallback`.  It allows you to "step into" the Graph Stage in a thread-safe way.  I'd be happy to explain that a bit further if it's something you want to explore

genus/src/test/scala/co/topl/genus/flows/TrackLatestFlowSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import akka.stream.scaladsl.{Keep, Sink, Source}
+import org.scalatest.flatspec.AsyncFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+
+class TrackLatestFlowSpec extends ScalaTestWithActorTestKit with AsyncFlatSpecLike with Matchers {
+  behavior of "Track Latest Flow"
+
+  it should "materialize to a function which returns the latest value seen" in {
+    val upstream = Source(List(1, 2, 3, 4))
```
SeanCheatham, line 12: I recommend using TestSource which gives you a probe for the materialized value.  You can then pass in values to the probe one-at-a-time and verify that your Sink's materialized value function works correctly for each element.

scasplte2:
this bring up a good point @b-vennes , perhaps we can chat about what is possible to replicate between the gRPC and json-rpc sometime as I've found it nice to describe to Chris and others that Genus and Topl-RPC are interchangeable. Admittedly, I have not been sure that is (or even could entirely) be the case though.

genus/src/main/scala/co/topl/genus/flows/TrackLatestFlow.scala
```diff
@@ -0,0 +1,66 @@
...
+            latestValue = next
+            push(outlet, next)
+          }
+
+          override def onUpstreamFinish(): Unit = super.onUpstreamFinish()
+
+          override def onUpstreamFailure(ex: Throwable): Unit = super.onUpstreamFailure(ex)
+
+        }
+      )
```
mgrand-topl, line 38: This is large enough that it should be moved to a separate function with a descriptive name.

genus/src/main/scala/co/topl/genus/interpreters/MongoChainHeight.scala
```diff
@@ -0,0 +1,69 @@
...
+
+    // get height will ask the tracker for the current chain height
+    val getHeight: () => BlockHeight =
+      Source
+        .tick(0.seconds, 10.seconds, NotUsed)
+        .flatMapConcat(_ =>
+          // get the current largest value
+          Source
+            .fromPublisher(
+              blocksCollection.find().sort(descendingHeightBlockSorting)
```
scasplte2, line 55: will this by default run against the entire collection? Really just thinking about indexing but if you're able to run this against a full synced Valhalla chain then probably nothing to be concerned about.

mgrand-topl: I'm a bit fuzzy on what the source is and how it should be handled. An ordering that we will need so often is a good candidate for a MongoDB index.  Indexing amortizes the cost of sorting over each insert and update. 

If this is coming from the database, we should be using an index so that we don't need to do a sort here. Mongo allows us to fetch the max value in an index directly.

If this is not coming from the database, then we may need to be concerned with the algorithm. We really need to just find the highest block, which does not require the work of doing a full sort. If we believe that the data is mostly sorted, then there are algorithms such as bubble sort that work well for that case. 

What is the source?

genus/src/main/scala/co/topl/genus/interpreters/MongoChainHeight.scala
```diff
@@ -0,0 +1,69 @@
...
+            )
+            .take(1)
+        )
+        // decode the documents into blocks and get the height
+        .mapConcat(document =>
+          DocumentDecoder[BlockDataModel].fromDocument(document).map(block => BlockHeight(block.height)).toSeq
+        )
+        // track the latest value seen
+        .toMat(TrackLatestFlow.create(defaultValue).to(Sink.ignore))(Keep.right)
```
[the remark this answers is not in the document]
b-vennes: Yes, I've updated the implementation of `TrackLatestFlow` to better handle this scenario. It no longer requires an initial value, but will instead make callers of `ChainHeight#get` wait for an initial value to come in

genus/src/main/scala/co/topl/genus/interpreters/MongoChainHeight.scala
b-vennes: The source is going to periodically emit the largest block height in the blocks collection.

PR #2112, opened 2022-05-11
fixed merge collision with google/protobuf files

b-vennes:
## Purpose

Errors occur during the assembly process when trying to merge the akka-grpc and protoc-java jars due to both jars containing the same .proto files.

## Approach

Filter for `google/protobuf` files during the assembly process and use `MergeStrategy.last` to use the google types from the latest upstream source (akka-grpc). 

## Testing

No new tests have been added

## Tickets
closes #2094

PR #2116, opened 2022-05-11
CORE-1216: Update action branch filters to include our GitFlow branches

admiraladmirable:
## Purpose
After updating our branching strategy, some of our GitHub actions have filters that don't match the new naming schemes (for example, this PR didn't run a check: https://github.com/Topl/Bifrost/pull/2112).

## Approach
* Update the filters to include our release branch patterns on PR and push.

## Tickets
_* closes #1216_

### DIFF
--- a/.github/workflows/sbt_checkPR.yml
+++ b/.github/workflows/sbt_checkPR.yml
@@ -2,9 +2,9 @@ name: SBT checkPR
 
 on:
   pull_request:
-    branches: ['*']
+    branches: [dev, main, release-*, rc-*]
   push:
-    branches: [dev, main]
+    branches: [dev, main, release-*, rc-*]
 
 jobs:
   build:
--- a/.github/workflows/postman_integration_tests.yml
+++ b/.github/workflows/postman_integration_tests.yml
@@ -2,7 +2,7 @@ name: Postman Integration Tests
 
 on:
   pull_request:
-    branches: [dev, main]
+    branches: [dev, main, release-*, rc-*]
 
 jobs:
 


.github/workflows/sbt_checkPR.yml
```diff
@@ -2,9 +2,9 @@ name: SBT checkPR
 
 on:
   pull_request:
-    branches: ['*']
+    branches: [dev, main, release-*, rc-*]
```
scasplte2, line 5: Are you sure this is what you want and will not cause PRs from a branch like `CORE-???-someFix` that is PR'd into `dev` to not run the `checkPR` script? Looking at the example here, this looks to me like the qualifiers here are acting like filters.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request

admiraladmirable: I believe this should work.

* In the case of this PR (with branch name`CORE-1216-update-bifrost-git-hub-action-branch-paths-to-reflect-our-updated-branch-strategy)`, it ran https://github.com/Topl/Bifrost/actions/runs/2309685341.
* I also tested in another repo and confirmed that the GitHub actions that are run for PRs are reflected by the PR trying to merge. (to confirm that this wasn't matching all branches like it was previously).

From my understanding, `pull_request` means "requests to merge into the branches listed in the array". i.e. `[dev, main, etc]`

b-vennes (approved):

scasplte2 (approved):

PR #2132, opened 2022-05-17
Environment Variable Configurations

SeanCheatham:
## Purpose
- Many modern deployment systems prefer environment variables to command-line arguments
  - But when _developing_ locally, environment variables can be a nuisance
- Adds optional support for environment variable setting overrides for a subset of configurations

## Approach
- Added `environment.conf` resource that implements environment variable substitutions for a few common settings
- Updated AppSettings config parser to parse environment.conf
  - Also added a helper to fix array-like setting overrides

## Testing
- Manually tested
- RPC Bind Address set via environment.  Forging disabled through environment but overridden by command-line.
  - image
  - image
  - image
- P2P Known Peers set via environment.
  - image
  - image

## Tickets
_* closes #1116_

### DIFF
--- a/node/src/main/resources/environment.conf
+++ b/node/src/main/resources/environment.conf
@@ -1,0 +1,22 @@
+bifrost {
+  application {
+    dataDir = ${?BIFROST_APPLICATION_DATA_DIR}
+    keyFileDir = ${?BIFROST_APPLICATION_KEY_FILE_DIR}
+  }
+
+  rpcApi {
+    bindAddress = ${?BIFROST_RPC_API_BIND_ADDRESS}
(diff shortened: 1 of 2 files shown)


scasplte2 (commented):
👍

On a separate note, could we make a ticket for the backlog to document the available environment variables and command line args? We can chat tomorrow morning about specifics if needed.

scasplte2 (approved):

SeanCheatham:
> On a separate note, could we make a ticket for the backlog to document the available environment variables and command line args? We can chat tomorrow morning about specifics if needed.

@scasplte2 

https://topl.atlassian.net/browse/CORE-1231

PR #2136, opened 2022-05-19
Filter query and subscription data using confirmation depth, chain height, and block height

b-vennes:
## Purpose

Data retrieved using queries and subscriptions should only be sent to the client if the associated block height is below the requested confirmation depth.

## Approach

Confirmation depth had to be implemented slightly differently for queries and subscriptions.

For queries, any data returned from mongo is filtered downstream based on the current chain height at the time of the request.

For subscriptions, the block height associated with each document retrieved in a batch will be checked against the current chain height. It is only after the entire filtered batch has been materialized into a `Seq` that the subscription index is incremented. 

## Testing

No new tests have been added.

## Tickets
* closes [CORE-1212 in Jira](https://topl.atlassian.net/browse/CORE-1212)


genus/src/main/scala/co/topl/genus/interpreters/MongoChainHeight.scala
```diff
@@ -63,7 +64,9 @@ object MongoChainHeight {
             .toSeq
         )
         // track the latest value seen
-        .toMat(TrackLatestFlow.create[BlockHeight].to(Sink.ignore))(Keep.right)
+        .viaMat(TrackLatestFlow.create[BlockHeight])(Keep.right)
+        // ignore value after tracked
+        .toMat(Sink.ignore)(Keep.left)
```
SeanCheatham, line 69: You _could_ consider keeping both materialized values here.  The `left` is a function that returns a Future, and the `right` is just a Future.  So you could do something like `() => Future.firstCompletedOf(rightFuture, leftFunction())`.  That way, once the stream completes, these function calls will immediately fail instead of returning stale data.

genus/src/main/scala/co/topl/genus/interpreters/BatchedMongoSubscription.scala
```diff
@@ -38,6 +42,18 @@ object BatchedMongoSubscription {
                   .getDocuments(filter.toBsonFilter.some, sort.toBsonSorting.some, batchSize.some, index.some)
                   .mapFunctor[Future]
               )
+              .flatMapConcat(document =>
+                Source
+                  .asyncF(chainHeight.get)
+                  .mapConcat(height =>
+                    documentToHeightOpt(document)
+                      .flatMap(documentHeight =>
+                        if (documentHeight.value <= height.value - confirmationDepth) document.some
```
SeanCheatham, line 51: Minor detail: Could probably use `.collect`

PR #2140, opened 2022-05-20
Tetra Syntactic Validation

SeanCheatham:
## Purpose
- A Transaction case class _can_ be created with data that is invalid for the blockchain (i.e. a negative timestamp)
- Syntactic Validation performs stateless checks on the data of the transaction
- Unrelated:
  - Box.value.value renamed to Box.value.quantity
  - Transaction#securityRoot is now strict-sized to 32 bytes

## Approach
- Add new `ledger` SBT module
- Add SyntacticValidationAlgebra
- Add SyntacticValidation interpreter
  - Validate non-empty inputs
  - Validate timestamp
  - Validate positive output quantities
  - Validate sufficient input funds

## Testing
- Added SyntacticValidationSpec

## Tickets
_* closes #2077_

### DIFF
--- a/common/src/main/scala/co/topl/modifier/transaction/builder/TransferBuilder.scala
+++ b/common/src/main/scala/co/topl/modifier/transaction/builder/TransferBuilder.scala
@@ -2,6 +2,8 @@ package co.topl.modifier.transaction.builder
 
 import cats.implicits._
 import co.topl.attestation.{Address, EvidenceProducer, Proposition}
+import co.topl.models.utility.HasLength.instances.bytesLength
+import co.topl.models.utility.Sized
 import co.topl.models.{Box => TetraBox, Bytes, FullAddress, Transaction}
 import co.topl.modifier.box._
 import co.topl.modifier.implicits._
@@ -386,7 +388,7 @@ object TransferBuilder {
           .map(assetCode =>
             Transaction.Output(
               consolidationAddress,
-              TetraBox.Values.Asset(asset._2.toSized, assetCode, Bytes.empty, None),
+              TetraBox.Values.Asset(asset._2.toSized, assetCode, Sized.strictUnsafe(Bytes.fill(32)(0: Byte)), None),
               minting
             )
           )
--- a/ledger/src/main/scala/co/topl/ledger/algebras/SyntacticValidationAlgebra.scala
+++ b/ledger/src/main/scala/co/topl/ledger/algebras/SyntacticValidationAlgebra.scala
@@ -1,6 +1,6 @@
 package co.topl.ledger.algebras
 
-import cats.data.ValidatedNec
+import cats.data.{Chain, ValidatedNec}
 import co.topl.models.{Box, Timestamp, Transaction}
 
 trait SyntacticValidationAlgebra[F[_]] {
@@ -17,4 +17,5 @@ object InvalidSyntaxErrors {
   case object EmptyInputs extends InvalidSyntaxError
   case class InvalidTimestamp(timestamp: Timestamp) extends InvalidSyntaxError
   case class NonPositiveOutputValue(outputValue: Box.Value) extends InvalidSyntaxError
+  case class InsufficientInputFunds[V <: Box.Value](inputs: Chain[V], outputs: Chain[V]) extends InvalidSyntaxError
 }
--- a/ledger/src/main/scala/co/topl/ledger/algebras/SyntacticValidationAlgebra.scala
+++ b/ledger/src/main/scala/co/topl/ledger/algebras/SyntacticValidationAlgebra.scala
@@ -1,0 +1,20 @@
+package co.topl.ledger.algebras
+
+import cats.data.ValidatedNec
+import co.topl.models.{Box, Timestamp, Transaction}
+
+trait SyntacticValidationAlgebra[F[_]] {
+
+  /**
+   * Determines the syntactic integrity of the given transaction
+   */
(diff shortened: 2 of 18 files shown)


ledger/src/main/scala/co/topl/ledger/interpreters/SyntacticValidation.scala
```diff
@@ -0,0 +1,107 @@
...
+  private def nonEmptyInputsValidation(transaction: Transaction): ValidatedNec[InvalidSyntaxError, Unit] =
+    Validated.condNec(transaction.inputs.nonEmpty, (), InvalidSyntaxErrors.EmptyInputs: InvalidSyntaxError)
+
+  /**
+   * Verify that the timestamp of the transaction is positive (greater than 0).  Transactions _can_ be created
+   * in the past.
+   */
+  private def positiveTimestampValidation(transaction: Transaction): ValidatedNec[InvalidSyntaxError, Unit] =
+    Validated.condNec(
+      transaction.timestamp > 0,
```
b-vennes, line 39: is 0 ever a valid timestamp?

ledger/src/main/scala/co/topl/ledger/interpreters/SyntacticValidation.scala
```diff
@@ -0,0 +1,107 @@
...
+      nonEmptyInputsValidation,
+      positiveTimestampValidation,
+      positiveOutputValuesValidation,
+      sufficientFundsValidation
+    )
+
+  /**
+   * Verify that this transaction contains at least one input
+   */
+  private def nonEmptyInputsValidation(transaction: Transaction): ValidatedNec[InvalidSyntaxError, Unit] =
```
b-vennes, line 30: would it be beneficial to make these validation functions public so that they could be tested independently?

ledger/src/main/scala/co/topl/ledger/interpreters/SyntacticValidation.scala
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Perform validation based on the quantities of boxes grouped by type
+   * @param f an extractor function which retrieves a BigInt from a Box.Value
+   */
+  private def quantityBasedValidation(transaction: Transaction)(
+    f: PartialFunction[Box.Value, BigInt] => ValidatedNec[InvalidSyntaxError, Unit]
+  ): ValidatedNec[InvalidSyntaxError, Unit] =
+    NonEmptyChain(
+      // Extract all Poly values and their quantities
+      f { case Box.Values.Poly(quantity) => quantity.data },
```
b-vennes, line 92: Could this partial function ever fail because of a match error?

common/src/main/scala/co/topl/modifier/transaction/builder/TransferBuilder.scala
```diff
@@ -386,7 +388,7 @@ object TransferBuilder {
           .map(assetCode =>
             Transaction.Output(
               consolidationAddress,
-              TetraBox.Values.Asset(asset._2.toSized, assetCode, Bytes.empty, None),
+              TetraBox.Values.Asset(asset._2.toSized, assetCode, Sized.strictUnsafe(Bytes.fill(32)(0: Byte)), None),
```
scasplte2, line 391: @SeanCheatham I believe at some point we had an optimization to not encode strings of zero data when going to the byte repr. Even if we didn't, do you think such an optimization would be worthwhile to save on bandwidth / storage costs?

Obviously not something needed for now but just wanted to note as something to discuss and a possible todo

SeanCheatham: Should this instead be an `Option[Sized.Strict[Bytes, Lengths.`32`.type]]`?

ledger/src/main/scala/co/topl/ledger/interpreters/SyntacticValidation.scala
SeanCheatham: We could probably technically allow `0` here.  I'll make the change to `>=`

ledger/src/main/scala/co/topl/ledger/algebras/SyntacticValidationAlgebra.scala
```diff
@@ -0,0 +1,21 @@
+package co.topl.ledger.algebras
+
+import cats.data.{Chain, ValidatedNec}
+import co.topl.models.{Box, Timestamp, Transaction}
+
+trait SyntacticValidationAlgebra[F[_]] {
```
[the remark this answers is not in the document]
SeanCheatham: Semantic and Authorization Validation will inherently require an F-context, so I figured I'd setup Syntactic Validation to be the same.

With that said, I'm on the fence about the return type being `F[Validated]`.  It could be better to have it just be `F[Transaction]` and have the interpreter require `F[_] : MonadError[InvalidSyntaxError, *]`.  What I have right now is simpler and easier, but for future improvement I might change it.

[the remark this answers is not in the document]
b-vennes: `F[Validated]` is pretty nice because it returns the domain-specific validation failures from the function. I was thinking about when `MonadError` should be used and I realized it could be best used to capture any truly-unexpected behaviors (like a data connection error), but in a way that the caller can recover from.

So in this case I think `F[Validation` is a great choice.

PR #2149, opened 2022-06-01
Partial Integration Test Fixes

SeanCheatham:
## Purpose
- The application configuration file format has changed in recent versions of Bifrost, but the integration tests still use the old format
  - Forging is also now disabled by default
- The client side of the Topl RPC codecs has an infinite self-recursive call
- BlockValidation failures are not explicitly logged

## Approach
- Update configurations used in integration tests to new format
- Send StartForging signal in SingleNodeTest
- Remove invalid RPC Codec entry
- Improve History BlockValidator logging
- Added integration test compilation to checkPR step
## Testing
- SingleNodeTest now passes
- The remaining integration tests still fail due to other bugs, but the tests at least start and run now
## Tickets
_* closes #1987_

### DIFF
--- a/.github/workflows/postman_integration_tests.yml
+++ b/.github/workflows/postman_integration_tests.yml
@@ -21,7 +21,7 @@ jobs:
         java-version: graalvm-ce-java11@21.1.0
         
     - name: Stage Docker Image
-      run: sbt bifrost/docker:stage
+      run: sbt node/docker:stage
 
     - name: Integration Tests
       run: docker-compose -f ./node/src/it/resources/postman-tests/docker-compose.yml up --exit-code-from postman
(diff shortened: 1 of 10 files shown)


.github/workflows/postman_integration_tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         java-version: graalvm-ce-java11@21.1.0
         
     - name: Stage Docker Image
-      run: sbt bifrost/docker:stage
+      run: sbt node/docker:stage
```
SeanCheatham, line 24: Both commands work, but `bifrost/docker:stage` will attempt to build any SBT module for which we have Docker enabled.  Figured I'd save time here and just do the node.

topl-rpc/src/main/scala/co/topl/rpc/ToplRpcCodecs.scala
```diff
@@ -297,12 +297,6 @@ trait TransactionRpcResponseDecoders extends SharedCodecs {
     ToplRpc.Transaction.UnprovenPolyTransfer.Response(x)
   )
 
-  implicit def transactionBroadcastTetraTransferResponseDecoder(implicit networkPrefix: NetworkPrefix): Decoder[
-    ToplRpc.Transaction.BroadcastTetraTransfer.Response
```
SeanCheatham, line 301: `type ToplRpc.Transaction.BroadcastTetraTransfer.Response = Transaction.TX` so this decoder was both redundant and infinitely recursive.

b-vennes (approved):

scasplte2:
@SeanCheatham should this target `rc-1.11` since it is a bugfix to the release? We will need to also merge into dev but I think this is immediately applicable to the rc

PR #2246, opened 2022-06-07
Tetra Transaction Notification Protocol

SeanCheatham:
## Purpose
- Transactions must be exchanged over the network so that they can be included in newly minted blocks
- Peers can be *notified* of a transaction's existence using the Transaction Notification Protocol
## Approach
- Added a new typed protocol to notify of transaction IDs
- When a peer notifies us:
  - Syntactically validate
  - Save transaction
  - Add to mempool
  - Rebroadcast
- "We" notify a peer when:
  - A Transaction is added to the mempool
## Testing
- Minor unit test updates
- Verified manually
  - image
## Tickets
_* closes #2073_


scasplte2 (approved):
📬

PR #2247, opened 2022-06-08
Tetra Box State

SeanCheatham:
## Purpose
- To support semantic validation, the system needs to know if a Box is spendable at some point in some chain of blocks
- A transaction outputs boxes.  These boxes can be spent by subsequent transactions and are included in box state.
- A transaction's input spend a previous transaction's output.  Once spent, the box is no longer spendable and is removed from box state
## Approach
- Implement BoxStateAlgebra and interpreter
- Interpreter based on Event Sourced Trees
## Testing
- Added unit test for BoxState
- Modified other ledger test suites to use MUnit
  - This involved some other modifications to build.sbt
  - This required a new SBT module which provides helpers/support for mocking in MUnit
## Tickets
_* closes #2076_

### DIFF
--- a/demo/src/main/scala/co/topl/demo/DemoProgram.scala
+++ b/demo/src/main/scala/co/topl/demo/DemoProgram.scala
@@ -6,7 +6,7 @@ import akka.util.ByteString
 import cats.data.{EitherT, OptionT, Validated}
 import cats.effect._
 import cats.implicits._
-import cats.{Applicative, MonadThrow, Parallel, Show}
+import cats.{Applicative, Monad, MonadThrow, Parallel, Show}
 import co.topl.algebras.{Store, StoreReader, ToplRpc, UnsafeResource}
 import co.topl.catsakka._
 import co.topl.codecs.bytes.tetra.instances._
@@ -58,6 +58,14 @@ object DemoProgram {
     for {
       (locallyAdoptedBlocksSource, locallyAdoptedBlocksSink) <-
         BroadcastHub.sink[TypedIdentifier].preMaterialize().pure[F]
+      (locallyAdoptedTransactionsSource, locallyAdoptedTransactionsSink) <-
+        BroadcastHub.sink[TypedIdentifier].preMaterialize().pure[F]
+      (remoteBlockSink, remoteBlockSource) <-
+        MergeHub.source[BlockV2].preMaterialize().pure[F]
+      (remoteTransactionsSink, remoteTransactionsSource) <-
+        MergeHub.source[Transaction].preMaterialize().pure[F]
+      ((offerRpcTransaction, terminateRpcTransactions), rpcTransactionsSource) =
+        Source.backpressuredQueue[F, Transaction]().preMaterialize()
       blockProcessor =
         (blockV2: BlockV2) =>
           MonadThrow[F].recoverWith(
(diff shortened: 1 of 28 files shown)


demo/src/main/scala/co/topl/demo/DemoProgram.scala
```diff
@@ -169,44 +192,57 @@ object DemoProgram {
                 )
             ),
             Sync[F]
-              .defer(Logger[F].info(show"Ignoring weaker block header id=${block.headerV2.id.asTypedBytes}"))
+              .defer(Logger[F].info(show"Ignoring weaker (or equal) block header id=${block.headerV2.id.asTypedBytes}"))
               .as(false)
           )
     } yield adopted
 
   private def toplRpcInterpreter[F[_]: Async: Logger: FToFuture](
```
scasplte2, line 200: I know we will likely just add to this as we go but I think these are the same pieces needed in the Dion RPC layer, right? Just to tie to an example we know. Thinking on it, we have NodeView (LedgerView??) and ConsensusView in some requests for Dion so we would probably want to add accessors for block store and epoch parameters.

Nothing needs to be done now, just noting a comparative example.

event-tree/src/main/scala/co/topl/eventtree/EventSourcedState.scala
```diff
@@ -51,24 +62,28 @@ object EventSourcedState {
     ) extends EventSourcedState[F, State] {
 
       def stateAt(eventId: TypedIdentifier): F[State] =
+        useStateAt(eventId)(_.pure[F])
```
scasplte2, line 65: does `.pure` come from `Async`? Is it worth reviewing (or at least pointing out to people) the typeclass hierarchy we are operating in? Or perhaps we should build out a typeclass hierarchy including our own?

networking/src/main/scala/co/topl/networking/blockchain/BlockchainClientHandler.scala
```diff
@@ -87,7 +87,38 @@ object BlockchainClientHandler {
                         .liftTo[F]
                     )
                 )
-                f2 = Async[F].fromFuture(
+                transactionsReceivedBackgroundF = Async[F].fromFuture(
```
scasplte2, line 90: would it be any nicer to inline `Async[F].fromFuture(...)`? Did you have plans to refactor this at some point since it seems like the for comprehension in the client `make` is getting a bit hard to follow

scasplte2 (approved):
🗳️ ✅

event-tree/src/main/scala/co/topl/eventtree/EventSourcedState.scala
SeanCheatham: `.pure` comes from `Applicative`: https://typelevel.org/cats/typeclasses/applicative.html

I'm fairly confident that it's out-of-date, but here's an infographic of the hierarchy for anyone interested: https://github.com/tpolecat/cats-infographic

Async includes Applicative in its hierarchy, so I'm able to get access to `.pure` through Async.

networking/src/main/scala/co/topl/networking/blockchain/BlockchainClientHandler.scala
SeanCheatham: I would expect cats-effect to provide syntax on a `F[Future[T]]` to transform it into `F[T]` using `Async`, but I haven't found it in the standard library.  I think it would be a very helpful syntax, but my gut says they probably don't have it for a reason.  I'm not quite sure though.

I think this file grew a bit large and is due for a bit of a refactor.

PR #2249, opened 2022-06-14
ContainsEvidence from ImmutableCodec

SeanCheatham:
## Purpose
- The `ContainsEvidence[T]` typeclass instance implementations repeat the same code
  - To abide by the Don't-Repeat-Yourself principle, this can be refactored into a helper method
## Approach
- Added a helper method to ContainsEvidence which constructs the instance from an ImmutableCodec instance (plus a type prefix)
## Testing
- Added ContainsEvidenceSpec
## Tickets
_* closes #1991_

### DIFF
--- a/typeclasses/src/main/scala/co/topl/typeclasses/ContainsEvidence.scala
+++ b/typeclasses/src/main/scala/co/topl/typeclasses/ContainsEvidence.scala
@@ -1,15 +1,12 @@
 package co.topl.typeclasses
 
 import co.topl.codecs.bytes.tetra.instances._
+import co.topl.codecs.bytes.typeclasses.ImmutableCodec
 import co.topl.codecs.bytes.typeclasses.implicits._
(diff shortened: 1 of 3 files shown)


typeclasses/src/main/scala/co/topl/typeclasses/ContainsEvidence.scala
```diff
@@ -1,217 +1,115 @@
...
-//            Bytes(
-//              blake2b256
-//                .hash(Ints.toByteArray(t.index) ++ t.address.allBytes.toArray)
-//                .value
-//            )
-//          )
-//        )
+      fromImmutableCodec(10)
 
     implicit val requiredInputBoxStateContainsEvidence: ContainsEvidence[Propositions.Contextual.RequiredBoxState] =
```
scasplte2, line 83: I know names don't matter here much, just pointing out that it is no longer "input" box state

scasplte2 (approved):
🗻

PR #2259, opened 2022-06-17
Block Body Semantic Validation

SeanCheatham:
## Purpose
- Transactions must not spend an output that has already been spent
- A block body can contain two transactions that spend the same input, and this would not be caught by simple Transaction Semantic Validation
## Approach
- Added Block Body Syntactic and Semantic Validation Algebras
- Added AugmentedBoxState which augments an underlying BoxStateAlgebra with ephemeral state
- BodySyntaxValidation runs TransactionSyntaxValidation on all of its transactions
- BodySemanticValidation runs TransactionSemanticValidation on all of its transactions
  - But while processing transactions, it includes the state of box spends for previous transactions within the block, and injects that data into Transaction Semantic Validation

## Testing
- Added new BodySyntaxValidationSpec and BodySemanticValidationSpec
## Tickets
- #2255


ledger/src/main/scala/co/topl/ledger/interpreters/BoxState.scala
```diff
@@ -104,3 +104,43 @@ object BoxState {
...
+      }
+    }
+
+  case class StateAugmentation(spentBoxIds: Set[Box.Id], newBoxIds: Set[Box.Id]) {
+
+    def augment(transaction: Transaction): StateAugmentation = {
+      val transactionSpentBoxIds = transaction.inputs.map(_.boxId).toIterable.toSet
+      val transactionId = transaction.id.asTypedBytes
+      val transactionNewBoxIds =
+        transaction.outputs.mapWithIndex((_, idx) => Box.Id(transactionId, idx.toShort)).toIterable.toSet
```
scasplte2, line 134: does `.mapWithIndex` implicitly convert the transaction output from `Short` to `Long` or something? Just wondering why we need `idx.toShort` or if that could be an issue.

ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
```diff
@@ -20,6 +20,7 @@ object TransactionSyntaxValidation {
   private[interpreters] val validators: Chain[Transaction => ValidatedNec[TransactionSyntaxError, Unit]] =
```
scasplte2, line 20: What you did for `TransactionSyntaxValidation` is much easier to understand each component since it is broken up with comments. I think `BodySyntaxValidation` could benefit from something similar. Also, not sure if you are doing this somewhere else but I think `nonEmptyInputsValidation & distinctInputsValidation` could also be applicable to `Body` as well? You got me realizing that the inputs for 
- boxes => transactions
- transaction => bodies
- bodies => headers
- headers => epochs

then all 4 of those things create new boxStates as their output. This is what got me thinking about looking for similarities in Syntax validation between those types of modifiers.

scasplte2 (approved):
🧊 🔍

ledger/src/main/scala/co/topl/ledger/interpreters/BoxState.scala
SeanCheatham: `.mapWithIndex` is a syntax operation method for anything with `Traverse[F[_]]`.  That method uses integers for the index value, but `Box.Id` requires a `Short` for the index value.  I would have assumed an implicit existed for Int => Short, but apparently it doesn't :shrug:

ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
SeanCheatham: I split up BodySyntaxValidation like you suggested, even though for now it's just a single little helper method that gets invoked.  In the future, I foresee some extra validation steps once we incorporate Rewards Transactions into the system.

On the subject on `nonEmptyInputsValidation`, I think a Block Body _can_ be empty.  In Dion, it can't be empty because of rewards transactions.  But in Tetra, I intend to change `type BlockBodyV2 = List[TypedIdentifier]` to `case class BlockBodyV2(transactionIds: List[TypedIdentifier], rewardsTransactionId: TypedIdentifier)`.  Doing it that way alleviates the need for a non-empty validation step.

And on the subject of `distinctInputsValidation`, this technically gets caught by BodySemanticValidation, but you're right that we could easily do that sooner.  I think I'd like to do two things here:
- Change `type BlockBodyV2 = List[TypedIdentifier]` to `type BlockBodyV2 = ListSet[TypedIdentifier]` (use a ListSet)
- Incorporate distinct Transaction.Input validation into BodySyntaxValidation

PR #2261, opened 2022-06-22
BlockBodyV2 List to ListSet

SeanCheatham:
## Purpose
- Naturally enforces a transaction isn't duplicated within a block
- Also validates that a block does not attempt to spend the same Transaction Output more than once
## Approach
- Change BlockBodyV2 type from List to ListSet
- Added new BodySyntaxValidation check to detect duplicate spends
## Testing
- New unit test for Body duplicate spend prevention
## Tickets
- #2255

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/BodySyntaxValidation.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/BodySyntaxValidation.scala
@@ -1,13 +1,25 @@
 package co.topl.ledger.interpreters
 
-import cats.data.{NonEmptyChain, Validated, ValidatedNec}
+import cats.data.{NonEmptyChain, NonEmptySet, ValidatedNec}
 import cats.effect.Sync
 import cats.implicits._
+import cats.{Foldable, Order}
 import co.topl.ledger.algebras._
 import co.topl.ledger.models._
-import co.topl.models.{BlockBodyV2, Transaction, TypedIdentifier}
+import co.topl.models._
+
+import scala.collection.immutable.SortedSet
 
 object BodySyntaxValidation {
+
+  implicit private val orderBoxId: Order[Box.Id] = {
+    implicit val orderTypedIdentifier: Order[TypedIdentifier] =
+      Order.by[TypedIdentifier, Bytes](_.allBytes)(Order.from[Bytes]((a, b) => a.compare(b)))
+    Order.whenEqual(
+      Order.by(_.transactionId),
+      Order.by(_.transactionOutputIndex)
+    )
+  }
 
   def make[F[_]: Sync](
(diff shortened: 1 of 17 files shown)


ledger/src/main/scala/co/topl/ledger/interpreters/BodySyntaxValidation.scala
```diff
@@ -1,14 +1,26 @@
...
 import co.topl.ledger.algebras._
 import co.topl.ledger.models._
-import co.topl.models.{BlockBodyV2, Transaction, TypedIdentifier}
+import co.topl.models._
+
+import scala.collection.immutable.SortedSet
 
 object BodySyntaxValidation {
 
+  implicit private val orderBoxId: Order[Box.Id] = {
```
scasplte2, line 15: @SeanCheatham could you describe the expected behavior? It is mostly not clear to me why an implicit `Order[TypedIdentifier]` needs to be present

ledger/src/main/scala/co/topl/ledger/interpreters/BoxState.scala
```diff
@@ -60,7 +60,7 @@ object BoxState {
     fetchTransaction: TypedIdentifier => F[Transaction]
   )(state:            State[F], blockId: TypedIdentifier): F[State[F]] =
     for {
-      body         <- fetchBlockBody(blockId)
+      body         <- fetchBlockBody(blockId).map(_.toList)
```
scasplte2, line 63: I believe this changed to needing a `.map` because a body is a case class now instead of just an alias for `List[Transaction]` right? So is this `.toList` method an implicit ops method on `BlockBody` to convert it to a list? If so where is that implicit coming from?

ledger/src/main/scala/co/topl/ledger/interpreters/BodySyntaxValidation.scala
SeanCheatham: I'm using a Cats NonEmptySet a bit further down in the file.  NonEmptySets are backed by SortedSets which require an implicit `Order[T]` in scope.  The actual ordering/sorting of the BoxIds isn't important for this use-case; it's only important in determining the equality of two BoxIds.

ledger/src/main/scala/co/topl/ledger/interpreters/BoxState.scala
SeanCheatham: `BlockBodyV2` is still just a type alias `type BlockBodyV2 = ListSet[TypedIdentifier]`, so the `.toList` is just acting on the ListSet.  In the future I plan to make it a `case class` once we incorporate rewards into the mix.

scasplte2:
@SeanCheatham 👍

PR #2267, opened 2022-06-29
Transaction/Body Validation Integration

SeanCheatham:
## Purpose
- Transaction and Body validation are already implemented as separate components
- They must be integrated into the rest of the program
  - When minting a new block
  - When validating a remote peer's block
  - When validating a remote peer's transaction (from mempool gossip)
  - When validating a user's transaction (from RPC)  

## Approach
- Created `blockchain` module
  - Migrated DemoProgram into this module, renamed to `Blockchain`
  - Migrated BlockchainPeerHandlerImpl
  - Migrated ToplRpcServer implementation
  - Added wrappers around LocalChain and Mempool to support rebroadcasting
- Integrated Body Validation into BlockMint
## Testing
- Added new basic unit tests for LocalChainBroadcaster and MempoolBroadcaster
- More tests will follow at a later date for BlockchainPeerHandler and ToplRpcServer
## Tickets
* #2256


blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
```diff
@@ -0,0 +1,147 @@
...
+          .map(Validated.Valid(_))
+          .tapAsyncF(1)(localChain.adopt)
+          .toMat(Sink.ignore)(Keep.right)
+          .liftTo[F]
+      _ <- rpcServer.use(binding =>
+        Logger[F].info(s"RPC Server bound at ${binding.localAddress}") >>
+        Async[F].fromFuture(blockStreamCompletionFuture) >>
+        p2pFiber.join
+      )
+    } yield ()
```
scasplte2, line 145: couple of questions on this file
1. Why the usage of `BroadcastHub.sink` for the notification streams vs. `Sink.ignore` for the `blockStreamCompletionFuture`?
2. Is the `blockStreamCompletionFuture` always pulling from the downstream because we are terminating the GraphStave with a `Sink.ignore` and can always materialize a value?

blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
```diff
@@ -0,0 +1,442 @@
...
+      (client: BlockchainPeerClient[F]) =>
+        for {
+          _adoptions <- client.remotePeerAdoptions
+          adoptions  <- _adoptions.withCancel[F]
+          _fetchAndValidateMissingHeaders = fetchAndValidateMissingHeaders(
+            client,
+            headerValidation,
+            slotDataStore,
+            headerStore
+          ) _
```
scasplte2, line 73: is this `_` a sugar thing?

blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
```diff
@@ -0,0 +1,442 @@
...
+          parentOf(data.head._1, data.head._2)
+            .flatMap(parentId =>
+              store
+                .contains(parentId)
+                .ifM(
+                  (data, true).pure[F],
+                  fetchRemoteData(parentId).map(parentData => (data.prepend((parentId, parentData)), false))
+                )
+            )
+        }(_._2)
```
scasplte2, line 313: this pattern is a little tricksy the iterate M testing on a single boolean value but building a chain. I like it just wondering what your experience is with clever implementations like this. I find I will probably be able to remember (or figure out) how it works in 6 months but I can almost guarantee it is going to take me a minute or two when I return to it.

I agree that inline documentation really messes with the style though and it seems cumbersome to articulate in the description. At least as far as the mechanism goes. What you have for the behavior description is fine.

blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
scasplte2: oh, nevermind. I see it is just taking the client input argument

scasplte2 (approved):
⬅️ 📦 ⬅️ 📦 ☁️ 🌩️

blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
SeanCheatham: 1. The notification streams are meant for outbound notifications of block IDs and transaction IDs.  As a block ID comes into the source, it gets broadcast to all subscribers of the BroadcastHub sink (meaning, all connected peers, gRPC listeners, and eventually the block mint).  `blockSteamCompletionFuture` is only meant to handle the locally minted blocks now.  I renamed it to `mintedBlockStreamCompletionFuture` for clarity.
2. `mintedBlockStreamCompletionFuture` captures the `Future[Done]` that is materialized from a `Sink.ignore`, so it simply captures the completion of the stream.  In this particular case, we don't _actually_ ever expect the stream to complete.

PR #2272, opened 2022-07-08
BN-155: Integration Test CI/CD

admiraladmirable:
## Purpose
Implement Scala integration tests for Bifrost via post commit workflow.

## Approach
This PR includes a few different optimizations and work to split things up into a PR build, and a post submit build. This allows us to test the quicker and more important parts before allowing submissions, and the slower parts post submit for every commit.

* Use reusable workflows to break apart common tasks into jobs.
  * In `Bifrost_PR` and `Bifrost_CI`, implement these common jobs. 
  * `Bifrost_CI` includes the Scala integration tests.
* Cache `target` folders (excluding `test-report` folders.
  * For `Bifrost_PR`, cache between builds.
  * For `Bifrost_CI`, only cache between jobs of the same build instance.
* Update `Bifrost_PR` build command to use `testQuick`, which combined with the project cache, only runs tests for source that changes.
* Expose test results folder from Scala integration tests so we can publish them.

## Testing
* Ran `Bifrost_PR`
  * Ensured build and test cache was used between re-runs.
  * Ensured changing one test rebuilt the source for that test, and ran the tests in that test class.
* Ran the `Bifrost_CI`
  * Ensured that the cache was not used between different builds.
  * Ensured Scala tests were running on the build.

Also, benefit of caching and `testQuick`, the PR time went from ~30min -> ~3.5 min best case (total time since combining Postman tests with PR is ~10min).

## Tickets
_*closes #2173*_

PR #2326, opened 2022-08-02
Fix integration test cache

admiraladmirable:
## Purpose
Quick fix after merging: https://github.com/Topl/Bifrost/pull/2272

## Approach
Remove download cache step for the scala integration test reusable workflow.

## Testing
https://github.com/Topl/Bifrost/actions/runs/2783012947


scasplte2 (approved):
👍

PR #2327, opened 2022-08-02
Fix order of classloaders

### LINKED ISSUE #2268: BN-549 ⁃ Wrong class loader order
When attempting to build from source using an ARM based Mac, the build fails with a “wrong class loader”.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/BN-549) by [Unito](https://www.unito.io)

mundacho:
## Purpose

Make the build and tests work on a mac.

## Approach

The error message was indicating a problem loading a class. Upon inspection it was found that the class loader used was the system class loader. The code tried to load the class using first the system class loader and then the class's class loader. The normal order is normally the opposite. The order was changed and now the code compiles and the tests pass.

## Testing

Tested on my mac and it works.

## Tickets
* closes #2268

### DIFF
--- a/node/src/main/scala/co/topl/db/LDBFactory.scala
+++ b/node/src/main/scala/co/topl/db/LDBFactory.scala
@@ -126,7 +126,7 @@ object LDBFactory extends Logging {
   }
 
   lazy val factory: DBFactory = {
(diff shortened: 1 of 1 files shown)


SeanCheatham (approved):

scasplte2 (approved):
💯  nicely done, thank you @mundacho

PR #2330, opened 2022-08-03
Fetch Block Body RPC

SeanCheatham:
## Purpose
- RPC Clients need the ability to fetch the transaction IDs associated with a block ID
## Approach
- Added FetchBlockBody RPC to topl_grpc.proto
- Interpreted using Block Body Store
## Testing
- Added unit tests for FetchBlockBody
## Tickets
* #2322


scasplte2 (approved):
👍

PR #2372, opened 2022-08-09
Added Proto model definitions

SeanCheatham:
## Purpose
- To support a "Fetch Transaction" RPC, there must be a `Transaction` message defined in the protobuf descriptor
  - A Transaction message necessarily involves defining messages for Boxes, Proofs, and Propositions

## Approach
- Created separate grpc sub-package which roughly reflects the co.top.models package

## Testing
- N/A
## Tickets
- #2323 (partially)
  - There will be follow-up PRs which _use_ these proto descriptor messages

### DIFF
--- a/topl-grpc/src/main/protobuf/models/box.proto
+++ b/topl-grpc/src/main/protobuf/models/box.proto
@@ -1,0 +1,45 @@
+syntax = "proto3";
+
+package co.topl.grpc.models;
+
+import 'models/common.proto';
+
+message Box {
+  message Id {
+    TransactionId transactionId = 1;
+    int32 transactionOutputIndex = 2;
+  }
+}
+
+message BoxValue {
+  oneof sealed_value {
+    EmptyBoxValue empty = 1;
+    PolyBoxValue poly = 2;
+    ArbitBoxValue arbit = 3;
+    DionAssetBoxValue asset = 4;
+    OperatorRegistrationBoxValue operatorRegistration = 5;
+  }
+}
+
+message EmptyBoxValue {}
+message PolyBoxValue {
+  Int128 quantity = 1;
+}
+message ArbitBoxValue {
+  Int128 quantity = 1;
+}
+message DionAssetBoxValue {
+  Int128 quantity = 1;
+  Code assetCode = 2;
+  bytes securityRoot = 3;
+  bytes metadata = 4;
+
+  message Code {
+    int32 version = 1;
+    bytes issuerAddress = 2;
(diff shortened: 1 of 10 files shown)


topl-grpc/src/main/scala/co/topl/grpc/MessageIsomorphism.scala
```diff
@@ -5,7 +5,7 @@ import cats.data.ValidatedNec
 import cats.implicits._
 import co.topl.codecs.bytes.tetra.instances._
 import co.topl.codecs.bytes.typeclasses.implicits._
-import co.topl.models
+import co.topl.{models => bifrostModels}
```
SeanCheatham, line 8: The new `.proto` files define messages in package `co.topl.grpc.models`.  This file here is located in `co.topl.grpc` (the parent of `co.topl.grpc.models`, thus `models` now references proto stuff.  I needed to alias `co.topl.models` to avoid the new collision.

topl-grpc/src/main/protobuf/models/box.proto
```diff
@@ -0,0 +1,45 @@
...
+  oneof sealed_value {
+    EmptyBoxValue empty = 1;
+    PolyBoxValue poly = 2;
+    ArbitBoxValue arbit = 3;
+    DionAssetBoxValue asset = 4;
+    OperatorRegistrationBoxValue operatorRegistration = 5;
+  }
+}
+
+message EmptyBoxValue {}
```
scasplte2, line 24: Dion separability might give us a clean way to handle the Asset v1 models as well

topl-grpc/src/main/protobuf/models/verification_key.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package co.topl.grpc.models;
+
+message VerificationKey {
```
scasplte2, line 5: let's talk through an example where we replace the VerificationKey with a Witness and see if there is a different language we want to adopt

scasplte2 (commented):
couple of thoughts
- do you want to review grpc (and record a video) for introing new team members?
- should we incorporate these models into a single way of approaching Dion and Tetra? What are the implications for the future here?

topl-grpc/src/main/protobuf/models/box.proto
SeanCheatham: :+1: For now, I've explicitly named it `DionAssetBoxValue` in the .proto

PR #2373, opened 2022-08-09
Update FetchHeader and FetchBody to use Isomorphisms

SeanCheatham:
## Purpose
- A previous PR established a set of protobuf messages/models to represent headers/bodies/transactions
- This PR uses some of those models (Block Header and Block Body) to replace the ones used previously
- This does not yet include implementations for fetching a Transaction
## Approach
- Rename MessageIsomorphism to Isomorphism.  Update syntax usage.
- Implemented Isomorphism instances for anything required in to transmit a BlockHeader or BlockBody
## Testing
- Update BlockHeader gRPC tests to verify isomorphic reciprocity of BlockHeader type
## Tickets
- #2323 (partially)
  - There will be follow-up PRs which implement FetchTransaction and the related Isomorphisms


scasplte2 (approved):
🐶 😶

PR #2415, opened 2022-08-11
Get Transaction RPC

SeanCheatham:
## Purpose
- Adds an RPC to read a Transaction by ID from a node
## Approach
- Added new RPC `FetchTransaction` to ToplGrpc service
- Also updated BroadcastTx RPC to accept a full Transaction (instead of Transaction bytes)
## Testing
- Unit test of FetchTransaction RPC
## Tickets
- #2323


scasplte2 (commented):
I don't quite understand what happened to this PR @SeanCheatham, perhaps what you mentioned this morning but I see some new stuff and some stuff from other PRs. Everything looks fine otherwise

scasplte2 (approved):
👍

PR #2416, opened 2022-08-11
Get Block ID at Height RPC

SeanCheatham:
## Purpose
- Implements a new RPC which retrieves the block ID associated with the requested height
## Approach
- Added new .proto RPC
- Implemented using LocalChain + BlockHeights
- Also updated BramblTetra to use new RPCs to determine genesis block+transactions
## Testing
- Unit test for new RPC
## Tickets
- #2324


scasplte2 (approved):
🆔 📏

PR #2424, opened 2022-08-15
GraalVM -> Temurin JDK

SeanCheatham:
## Purpose
- We currently do not leverage GraalVM for its main benefits
- GraalVM has a very heavy installation footprint which slows down builds/CI
- Until we _need_ GraalVM, a standard JDK distribution will suffice
## Approach
- Switch to Eclipse's Temurin JDK (successor to AdoptOpenJDK)
- Update GitHub workflows to specify Java 11 as the target version
- Update SBT Build workflow to install Temurin Java as a separate step
  - SBT Install is now a separate manual workflow step
  - (setup-scala uses jabba which has not been updated with Temurin support)
## Testing
- preparePR locally
- GitHub Workflow Checks
## Tickets
* #2423

### DIFF
--- a/topl-grpc/src/main/protobuf/topl_grpc.proto
+++ b/topl-grpc/src/main/protobuf/topl_grpc.proto
@@ -15,6 +15,7 @@ service ToplGrpc {
   rpc FetchTransaction (FetchTransactionReq) returns (FetchTransactionRes);
 
   rpc FetchBlockIdAtHeight (FetchBlockIdAtHeightReq) returns (FetchBlockIdAtHeightRes);
+
 }
 
 message BroadcastTransactionReq {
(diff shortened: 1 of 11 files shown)


topl-grpc/src/main/protobuf/topl_grpc.proto
```diff
@@ -15,6 +15,7 @@ service ToplGrpc {
   rpc FetchTransaction (FetchTransactionReq) returns (FetchTransactionRes);
 
   rpc FetchBlockIdAtHeight (FetchBlockIdAtHeightReq) returns (FetchBlockIdAtHeightRes);
+
```
admiraladmirable, line 18: Was this newline intentional?

admiraladmirable (approved):
🚀

scasplte2 (approved):
🦐

topl-grpc/src/main/protobuf/topl_grpc.proto
SeanCheatham: It was an erroneous change, but it's harmless so I figure I'll keep it for now

PR #2425, opened 2022-08-16
Bn 552 bifrost node definition

### LINKED ISSUE #2264: BN-544 ⁃ Bifrost Helm Chart Options for Bifrost Node
To deploy Bifrost to GKE, we need to have Kubernetes files created that define the components. We should go a step further and make this into a Helm chart, that way we can configure how it is deployed in the future (for now, focus on getting just GCP working).

The components should be thought out more, but thoughts on a first pass:

* Define the Bifrost node pod (probably stateful set deployment).
* Define persistent storage in GCP.
* Define load balancer between pods.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/BN-544) by [Unito](https://www.unito.io)

admiraladmirable:
## Purpose
To help with deployment in cloud native environments, create a Helm chart for running a Bifrost node

## Approach
This is the first pass of a simple Bifrost node helm chart. Use [node besu helm chart](https://github.com/hyperledger/bevel/tree/main/platforms/hyperledger-besu/charts/node_besu) for inspiration.

The Helm chart consists of a `StatefulSet` deployment for the `bifrost-node` Docker image, which will also spin up a `PersistentVolume` automatically.

The `Service` routes traffic back to the pods.

### Creation
1. Use `helm create bifrost` to create a template Helm chart.
2. Remove extra files and scope the chart to a `StatefulSet` and a `Service`.

Also ran `checkov` on the chart and fixed any issues that appeared.

## Testing
Deployed basic example in GCP which just attached to the Valhalla network. Verified connection via logs.

## How to use

Locally:
1. Install Docker Desktop, enable Kubernetes.
2. Install Helm
3. In root of Bifrost repo: `helm upgrade --install bifrost-node-deployment-name ./helm/bifrost/ -f values.yaml`
4. `-f values.yaml` is a user specified override for the values. This part is optional.

### DIFF
--- a/helm/bifrost/.helmignore
+++ b/helm/bifrost/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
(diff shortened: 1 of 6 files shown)


scasplte2 (approved):
🎛️

PR #2426, opened 2022-08-16
LevelDB Store Interpreter

SeanCheatham:
## Purpose
- Currently, the only interpreter for `Store` is in-memory and non-persistent
- This PR introduces a `Store` interpreter that persists data to disk
## Approach
- Created new separate LevelDB SBT module which interprets the `Store` algebra
  - Separate SBT module allows us to isolate the unit tests, allowing us to disable _just_ that module when unit testing on machines that don't support LevelDB very well
## Testing
- Added LevelDbStoreSpec
  - Incorporated FS2 to help with the File IO (because deleting a directory with java NIO is rather clunky)
## Tickets
- #2383

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -661,6 +661,20 @@ lazy val toplGrpc = project
     munitScalamock % "test->test"
   )
 
+lazy val levelDbStore = project
+  .in(file("level-db-store"))
+  .settings(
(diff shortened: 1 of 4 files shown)


scasplte2 (approved):
💾

PR #2428, opened 2022-08-17
Incorporate Proto Definition and Models Feedback

SeanCheatham:
## Purpose
- Proto v3.15.0 added support for the `optional` syntax in proto3 specs.  This avoids the need to use a pattern of implementing new message types to allow optionality of scalar types
- Some proto definitions specified int64 where uint64 would be better
- Rename Box.Values.Asset to Box.Values.AssetV1
- More type aliases in the `models` package
- Note: There may be future changes/PRs to the proto definitions.  I'm hoping to incorporate the feedback as it comes so that we can make sure it holds together on the server-side
## Approach
- Incorporate `optional bytes` into proto definitions that defined separate Data/Metadata message types (header, transaction, asset box)
- Update timestamp and slot fields to be uint64 type
- Renaming of classes and type aliases
## Testing
- Unit tests still pass
## Tickets
N/A

### DIFF
--- a/common/src/main/scala/co/topl/modifier/ops/AssetCodeOps.scala
+++ b/common/src/main/scala/co/topl/modifier/ops/AssetCodeOps.scala
@@ -2,7 +2,7 @@ package co.topl.modifier.ops
 
 import cats.implicits._
 import co.topl.attestation.Address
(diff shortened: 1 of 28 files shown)


scasplte2 (approved):
👍

PR #2434, opened 2022-08-19
Implement checkov for the PR and CI

admiraladmirable:
## Purpose
Add Helm/Kubernetes security analyzer to the PR.

## Approach
Split from: https://github.com/Topl/Bifrost/pull/2425

[Checkov](https://www.checkov.io/) is an open source static security analysis tool that will catch common misconfigurations in various DevOps file types. In this case, we will use it for Kubernetes yaml file scanning to make sure we are following best practices for our Helm charts.

## Testing
Local testing for Helm chart branch https://github.com/Topl/Bifrost/pull/2425.

### DIFF
--- a/.github/workflows/_validate_helm.yml
+++ b/.github/workflows/_validate_helm.yml
@@ -1,0 +1,25 @@
+name: Validate Helm Chart
+
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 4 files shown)


scasplte2 (approved):
🤖

PR #2437, opened 2022-08-22
Rename Transaction.Chronology to Transaction.Schedule

SeanCheatham:
## Purpose
- Transaction.Chronology is not descriptive enough for the field's purpose
## Approach
- Rename Transaction.Chronology to Transaction.Schedule
## Testing
- N/A
## Tickets
- #2374


scasplte2 (approved):
🕐

PR #2446, opened 2022-08-26
Tetra "main" Application

SeanCheatham:
## Purpose
- The new `nodeTetra` needs a proper implementation of the "main" (the entrypoint for the JVM)
  - NOTE: There are follow-up tickets which introduce command-line arguments and configurations, so for now, this node is a single local node with no other peers
## Approach
- Leverage the code from `TetraSuperDemo` as reference
- Create "PrivateTestnet" concept
  - The network's "big bang timestamp" is used as the "seed" for the generated stakers
- Refactor the minting process using a `BlockProducer`
  - In preparation for an improved "Block Packer" mechanism
## Testing
- Ran the Bifrost Tetra docker image locally
  - Ran "BramblTetra" to produce/emit transactions and verified they are included in the chain
  - Ran "BramblTetraMempoolReader" to determine if transactions make their way out of the mempool
- BlockProducer unit tests 
## Tickets
- #2379


scasplte2:
I'd like to chat about magic bytes that are derived from a shared public string. Just a thought that occurred to me the other day.

scasplte2 (approved):
- What is the use-case of the `Abandoner` and is this replicable in FS2?

SeanCheatham:
> * What is the use-case of the `Abandoner` and is this replicable in FS2?

@scasplte2 The `Abandoner`'s use-case is to allow the "Block Producer" to short-circuit its current block creation if a new canonical head is adopted.  If a new block is adopted by the node, then we want the Block Producer to immediately start building upon the new one instead of waiting for it to complete its previous attempt (which would waste precious slots/time).

PR #2460, opened 2022-09-08
Enforce Transaction.Schedule

SeanCheatham:
## Purpose
- A Transaction can define constraints on _when_ it can be included in the blockchain.
- These constraints need to be enforced during minting and validation
## Approach
- Modify Transaction and Body Semantic Validations to use additional context information in its validation
- Enforce Transaction.Schedule constraints in TransactionSemanticValidation
- Also validate Transaction.Schedule values in TransactionSyntaxValidation
## Testing
- New Transaction Semantic and Syntax unit tests for schedule
## Tickets
#2435


scasplte2 (approved):
🔥

PR #2462, opened 2022-09-08
ParentChildTree Persistence

SeanCheatham:
## Purpose
- The backbone of the event-sourced-state mechanism is the ability to traverse a _tree_ of IDs
- The tree of IDs is updated by the application as data arrives.  Upon restart, the tree would not be preserved
## Approach
- Re-interpret the ParentChildTree using a `Store` (which _may_ be a persistent store)
  - The child ID is the key
  - A tuple (child height, parent ID) is the value
## Testing
- Update EventSourcedStateSpec to use ParentChildTree.FromStore
## Tickets
- #2385 
- #2252


scasplte2 (approved):
💾

PR #2463, opened 2022-09-09
Current Event ID Persistence

SeanCheatham:
## Purpose
- The Event-Sourced-Tree mechanism frequently updates its current state to reflect data at a particular ID
- The ID of the state needs to be persisted whenever it changes, such that it can be recovered upon node restart
## Approach
- Create a new Store which stores an ID entry for each of the event-sourced trees that are used (consensus data, epoch boundaries, block heights, box state, and the mempool).
  - Also store the "canonical head" ID
## Testing
- Node still operates normally
## Tickets
- #2386


scasplte2 (approved):
🎄

PR #2480, opened 2022-09-20
Tetra Command-Line Args

SeanCheatham:
## Purpose
- Introduces a set of command-line arguments to modify the behavior of a node
- Fixes a couple of bugs when **re**launching a node
## Approach
- Added more structural definition to the Args case class
  - Startup args for arguments needed during initialization
  - Runtime args for arguments which modify the runtime behavior
 - Apply Args to ApplicationConfig
 - Incorporate MonadCancel#uncancelable to prevent cancellation of critical "save" operations
## Testing
- Node still runs locally
- Created node-tetra/example/docker-compose.yaml
## Tickets
- #2381 
- #2287


scasplte2 (approved):
🖥️ 📻 📶

PR #2481, opened 2022-09-21
Update Scalacache Dependency (Dev)

SeanCheatham:
## Purpose
- scalacache is a library for caching within an effect context
- scalacache 1.0.0-M6 introduces a backwards incompatible change which allows CaffeineCache to specify a Key type
  - (This is a really helpful change because previously, the keys needed to be converted to strings which is inefficient)
## Approach
- Update scalacache dependency
- Update usages of CaffeineCache.apply to specify a key type
- In cases which used TypedIdentifier as the key type, the type was changed to just `Bytes` due to an issue with TypedBytes being an `@newtype` which does not work as a key
## Testing
- N/A
## Tickets
#2419


scasplte2 (approved):
🤖

PR #2483, opened 2022-09-22
Use Headless Service for Bifrost for direct pod communication (for peer to peer)

admiraladmirable:
## Purpose
To enable p2p networking for our Bifrost, also expose the application using a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services).

## Approach
Create an additional service in Helm, use that to route p2p traffic, and keep the existing RPC service.

## Testing
Deployed with ArgoCD. Made sure passing `knownPeers` on the CLI worked (logs showed peers communicating).
Ensured RPC traffic worked via Postman.


SeanCheatham (approved):

scasplte2 (approved):
💻 ⬅️➡️ 💻

PR #2484, opened 2022-09-22
Cache Store Interpreter

SeanCheatham:
## Purpose
- Blockchain data is saved to disk using a LevelDb Store Interpreter
- This PR introduces a new interpreter which wraps an existing Store with a caching layer
## Approach
- Create new CacheStore interpreter which accepts a base Store and a configuration function
- Configurations for the store caches
## Testing
- New unit tests for CacheStore
## Tickets
#2469


scasplte2 (approved):
☕ 💾

PR #2485, opened 2022-09-26
Transaction Schedule Non-Negative Minimum Slot

SeanCheatham:
## Purpose
- Slots can never practically be non-negative in the protocol
- Users should not use negative slots when declaring transaction schedule bounds
## Approach
- Verify Transaction.schedule.minimumSlot >= 0
## Testing
- Update unit test to verify non-negative schedules
## Tickets
#2473


scasplte2 (approved):
➕

PR #2486, opened 2022-10-04
Implement CatsUnsafeResource

SeanCheatham:
## Purpose
- Reduce dependency on Akka
- Enable a parallelism limit on UnsafeResource creation
## Approach
- Interpret UnsafeResource a cats-effect queue
  - Uninitialized resources are enqueued at the start, up to max parallelism
  - As requests arrive, resources are dequeued.  If uninitialized, initializes.
## Testing
- Modified ActorPoolUnsafeResourceSpec to new implementation
## Tickets
- BN-654


scasplte2 (approved):
🚧 🐱 🛢️

PR #2488, opened 2022-10-04
Operational Keys Parallelization

SeanCheatham:
## Purpose
- At the start of each operational period, a staker generates several Ed25519 child linear keys.  This process takes a long time and could be multi-threaded
## Approach
- Use cats-effect's parTraverse to generate keys in parallel
## Testing
- NodeApp still runs normally
## Tickets
- BN-655


scasplte2 (approved):
🛤️

PR #2495, opened 2022-10-06
Tsdk 46 create orient db database 2

mgrand-topl:
## Purpose
This is another partial PR. This PR provides code to run the database, but without any configuration information.
## Approach
The remaining work will consist of providing the DB with configuration information and defining the schema.
## Testing
unit tests
## Tickets
* TSDK-46

### DIFF
--- a/genus-library/project/build.properties
+++ b/genus-library/project/build.properties
@@ -1,0 +1,2 @@
+# suppress inspection "UnusedProperty" for whole file
+sbt.version=1.6.2
--- a/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
@@ -3,7 +3,7 @@ package co.topl.genusLibrary
 import co.topl.genusLibrary.util.Log
 import com.typesafe.scalalogging.Logger
 
-import scala.util.{Success, Try}
+import scala.util.{Failure, Success, Try}
 
 /**
  * This is the main class of Genus. It captures blocks in an OrientDB database and provides ways to index and query the
@@ -12,7 +12,10 @@ import scala.util.{Success, Try}
 class Genus {
   import Genus.logger
 
-
+  private def shutDown(): Try[Unit] = {
+    logger.info("Genus shutting down")
+    Success()
+  }
 }
 
 object Genus {
@@ -24,7 +27,11 @@ object Genus {
    * This variable refers to the running instance of Genus when it is defined.
    * Call the startup method to define it.
    */
-  private val instance = Option.empty[Genus]
+  private var instance = Option.empty[Genus]
+
+  private def ensureActive[T](thunk: => T): Try[T] = {
+    instance.map( _ => Success(thunk)).getOrElse(Failure( GenusException("")))
+  }
 
   /**
    * This method returns the active instance of Genus. If one does not exist, it is created.
@@ -32,9 +39,22 @@ object Genus {
   def getGenus:Try[Genus] = {
     Log.debug("getGenus called: {}") {
       this.synchronized {
-        instance.map(Success(_)).getOrElse(Try(new Genus))
+        instance.map(Success(_)).getOrElse(Try{
+          instance = Some(new Genus)
+          instance.get
+        })
       }
     }
   }
 
+  def shutDown(): Try[Unit] = {
+    Log.info("shutDown() called") {
+      this.synchronized {
+        ensureActive {
+          instance.get.shutDown()
+          instance = None
+        }
+      }
+    }
+  }
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
@@ -13,9 +13,11 @@ import scala.util.{Failure, Success, Try}
  */
 class Genus {
   import Genus.logger
+  private var orientDB = OrientDBFacade()
 
   private def shutDown(): Try[Unit] = {
     logger.info("Genus shutting down")
+    orientDB.shutdown()
     Success(())
   }
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
@@ -1,13 +1,35 @@
 package co.topl.genusLibrary
 
+import com.typesafe.scalalogging.Logger
+
+import scala.util.{Success, Try}
+
 /**
  * This is the main class of Genus. It captures blocks in an OrientDB database and provides ways to index and query the
  * blocks, transactions and boxes.
  */
 class Genus {
+  import Genus.logger
+
 
 }
 
 object Genus {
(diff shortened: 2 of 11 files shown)


genus-library/project/build.properties
```diff
@@ -0,0 +1,2 @@
+# suppress inspection "UnusedProperty" for whole file
+sbt.version=1.6.2
```
SeanCheatham, line 2: A repository-wide `build.properties` already exists at `project/build.properties`

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
```diff
@@ -1,13 +1,65 @@
...
  * This is the main class of Genus. It captures blocks in an OrientDB database and provides ways to index and query the
  * blocks, transactions and boxes.
  */
 class Genus {
+  import Genus.logger
+  private var orientDB = OrientDBFacade()
 
+  private def shutDown(): Try[Unit] = {
+    logger.info("Genus shutting down")
+    orientDB.shutdown()
```
SeanCheatham, line 21: This method call should probably be wrapped in `Try(...)`

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
```diff
@@ -1,13 +1,65 @@
...
+    Success(())
+  }
 }
 
 object Genus {
+  private implicit val logger: Logger = Logger(classOf[Genus])
+
+  logger.info("Constructing Genus instance")
```
SeanCheatham, line 29: I think this statement would be logged even if the `Genus` construction is delayed.  It should probably be moved to the body of `getGenus`

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
```diff
@@ -1,13 +1,65 @@
...
+
+  /**
+   * This variable refers to the running instance of Genus when it is defined.
+   * Call the startup method to define it.
+   */
+  private var instance = Option.empty[Genus]
```
SeanCheatham, line 35: Global singleton patterns are risky and tend to lead to inconsistent unit test behavior.  I think consumers of the `genusLibrary` should be encouraged to construct their own instances.

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
```diff
@@ -1,13 +1,65 @@
...
+
+  private def ensureActive[T](thunk: => T): Try[T] = {
+    instance.map( _ => Success(thunk)).getOrElse(Failure( GenusException("")))
+  }
+
+  /**
+   * This method returns the active instance of Genus. If one does not exist, it is created.
+   */
+  def getGenus:Try[Genus] = {
+    Log.debug("getGenus called: {}") {
```
SeanCheatham, line 45: It may be simpler to just use `private lazy val instance = new Genus` (or something similar), instead of storing a var that is later expected to be initialized.

genus-library/src/main/scala/co/topl/genusLibrary/OrientDBFacade.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.File
+
+/**
+ * This is a class to hide the details of interacting with OrientDB.
+ */
+class OrientDBFacade {
+  import OrientDBFacade._
+
+  var dbDirectory = new File("genus_db")
```
SeanCheatham, line 15: Could these be `val`s?

genus-library/src/main/scala/co/topl/genusLibrary/OrientDBFacade.scala
```diff
@@ -0,0 +1,63 @@
...
+  private var RootPassword = "ae9fdaiwbfb9vu=vvau2#"
```
SeanCheatham, line 16: Just to confirm, this is only for temporary testing and will be removed in subsequent PRs?

(Actually I still suggest at least renaming this field.  I think the `GitGuardian` bot may complain once this goes into the `tetra` branch because of a field named Password)

genus-library/src/test/scala/co/topl/genusLibrary/OrientDBFacadeTest.scala
```diff
@@ -0,0 +1,33 @@
+package co.topl.genusLibrary
+
+import java.io.{File, FileOutputStream}
+
+class OrientDBFacadeTest extends munit.FunSuite {
+  def randomFile(): File = {
+    new File("TF" + Math.random().toString)
```
SeanCheatham, line 7: The `fs2` library has helpers for creating and deleting files/resources if you'd like.  But if not, I think this should at least create random files in /tmp.  (i.e. use `java.nio.Paths.createTempDirectory(...)`).  Otherwise, if any of the tests fail, there would be erroneous files in the current directory.

genus-library/project/build.properties
mgrand-topl: @SeanCheatham sbt keeps creating this. You said that I should not use a .gitignore at the project level, so the simplest thing seemed to be to commit it. If there is some way I can get sbt to stop generating this, I would be happy to do that. What do you suggest?

genus-server/src/main/scala/co/topl/genusServer/GenusServerApp.scala
```diff
@@ -21,5 +20,9 @@ object GenusServerApp extends IOApp {
     } yield ExitCode.Success
 
   def doIt(args: List[String]): IO[Unit] =
-    IO(Genus.getGenus)
+    IO {
```
SeanCheatham, line 23: Not necessary to do now, but this would be a good use-case for cats-effect's Resource.

i.e.
```scala
Resource.make[IO, Genus](IO.delay(Genus.getGenus))(genus => IO.delay(genus.shutdown()))
  .use(genus =>
    // Use initialized genus instance
    IO.unit
  )
  ```
  
  This will ensure proper cleanup of Genus even if the application fails or throws an exception.

File: genus-library/src/main/scala/co/topl/genusLibrary/util/Log.scala [hunk 10]
```diff
@@ -0,0 +1,53 @@
...
+object Log {
+  /**
+   * Log, at info level, the result of an expression like this:
+   * {{{
+   *   Log.info("f returned {}") {
+   *     f(oneArg, anotherArg)
+   *   }
+   * }}}
+   */
+  def info[T](message: String)( body: => T)(implicit logger: Logger): T = {
```
SeanCheatham, line 19: I noticed the spacing is slightly off in the `( body: => T)` parameter group.  I think you'll need to modify the `build.sbt` file to add `genusServer` and `genusLibrary` to the `.aggregate(...)` section near the top of the file.  This will enable scalafmt and scalafix for these modules when running `sbt preparePR`.

genus-library/project/build.properties
SeanCheatham: I'm not sure why it would generate project-level `build.properties` files.  Are you running `sbt` from the repository root?

SeanCheatham (approved):

PR #2496, opened 2022-10-07
ConsensusDataEventSourcedState Multi-Tx Unit Test

SeanCheatham:
## Purpose
- A bug existed previously in ConsensusDataEventSourcedState where the "UnapplyBlock" step did not reverse the transactions of a block before unapplying them.  The bug is already fixed, but this PR includes a unit test which includes multiple transactions in a block.
## Approach
- Updates the existing `ConsensusDataEventSourcedStateSpec` to verify behavior of a block with multiple transactions
## Testing
- Unit test passes
## Tickets
#2445


scasplte2 (approved):
👍

PR #2497, opened 2022-10-07
Chain Sync Beyond 2 Epochs

SeanCheatham:
## Purpose
- When synchronizing with a remote peer, the local node may be more than 2 epochs behind
- Header validation requires knowing block bodies from 2 epochs before the header being validated
## Approach
- Group tines by epoch, and fetch data one epoch at a time
## Testing
- Ran two nodes locally and was able to sync the 2nd one further than 2 epochs
## Tickets
- BN-659

### DIFF
--- a/blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
@@ -7,7 +7,8 @@ import cats.effect.Async
 import cats.effect.kernel.Sync
 import cats.implicits._
 import cats.{Applicative, Monad, MonadThrow, Monoid, Parallel, Show}
-import co.topl.algebras.{Store, StoreReader}
+import co.topl.algebras.ClockAlgebra.implicits.ClockOps
+import co.topl.algebras.{ClockAlgebra, Store, StoreReader}
 import co.topl.catsakka._
 import co.topl.codecs.bytes.tetra.instances._
 import co.topl.codecs.bytes.typeclasses.implicits._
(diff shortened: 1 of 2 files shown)


blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
```diff
@@ -127,8 +134,24 @@ object BlockchainPeerHandler {
                           ifTrue = for {
                             _ <- Logger[F]
                               .debug(show"Remote tine (head id=$id) appears to be better than the local chain")
-                            _ <- _fetchAndValidateMissingHeaders(id)
-                            _ <- _fetchAndValidateMissingBodies(id)
+                            // The remote tine may be large, extending beyond 2 epochs.  Header validation relies on
+                            // block bodies (lagging by 2 epochs), so group the tine by epoch step through each
+                            // group.
```
scasplte2, line 139: this description makes sense but I'll admit it is a bit unclear to me what is happening in the code, might just need to step through it in the morning

scasplte2 (approved):
🕙 ⬅️ 2️⃣

blockchain/src/main/scala/co/topl/blockchain/BlockchainPeerHandler.scala
scasplte2: would you like to merge this nd go over next week @SeanCheatham ? I wouldn't want to carry over a ticket just because we haven't chatted about it.

SeanCheatham: :+1: Yup that sounds good to me

PR #2501, opened 2022-10-12
CORE - BN-666 - avoid warning in pattern matching

nandotorterolo:
## Purpose
Avoid warning
> It would fail on the following inputs: HashLock(_, _), KesProduct(_, _, _), KesSum(_, _, _), Not(_), RequiredBoxState(), VrfEd25519(_)

## Approach
add last branch or pattern matching with Json.null

## Testing

unit test
## Tickets
BN-666

### DIFF
--- a/json-codecs/src/main/scala/co/topl/codecs/json/tetra/ModelsJsonCodecs.scala
+++ b/json-codecs/src/main/scala/co/topl/codecs/json/tetra/ModelsJsonCodecs.scala
@@ -286,10 +286,10 @@ trait ModelsJsonCodecs {
         "args"      -> io.circe.parser.parse(serializedArgs).getOrElse(serializedArgs.asJson)
       )
     case _ =>
-
       /**
        * It would fail on the following inputs: HashLock(_, _), KesProduct(_, _, _), KesSum(_, _, _), Not(_), RequiredBoxState(), VrfEd25519(_)
-       * TODO: ask if raising an error here is fine, never happen, or should we define specific encoders for the above items
+       * It may lead to odd behavior for users who hit this when decoding,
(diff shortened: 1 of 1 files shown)


json-codecs/src/main/scala/co/topl/codecs/json/tetra/ModelsJsonCodecs.scala
```diff
@@ -285,6 +285,13 @@ trait ModelsJsonCodecs {
         "proofType" -> "Script.JS".asJson,
         "args"      -> io.circe.parser.parse(serializedArgs).getOrElse(serializedArgs.asJson)
       )
+    case _ =>
+
+      /**
+       * It would fail on the following inputs: HashLock(_, _), KesProduct(_, _, _), KesSum(_, _, _), Not(_), RequiredBoxState(), VrfEd25519(_)
+       * TODO: ask if raising an error here is fine, never happen, or should we define specific encoders for the above items
+       */
+      Json.Null
```
SeanCheatham, line 294: I think this is suitable behavior for now.  It may lead to odd behavior for users who hit this when decoding, but only Dion users would hit this, and these Propositions aren't supported in Dion anyway.

SeanCheatham (approved):

json-codecs/src/main/scala/co/topl/codecs/json/tetra/ModelsJsonCodecs.scala
nandotorterolo: Thanks Sean, I will update te commend with your response, and delete todo message

scasplte2 (approved):
🔥

PR #2502, opened 2022-10-13
Network Delayer Application

SeanCheatham:
## Purpose
- Taktikos was designed to behave under real-world network constraints, such as limited latency or bandwidth
- Testing real-world network constraints in a global manner is difficult and expensive
- Simulating network constraints is relatively cheap, and allows for nodes running within the same data center to act as-if they were connected at long distances
- This PR introduces an application which intercepts TCP traffic and throttles the data to configurable amounts
## Approach
- Add new `commonApplication` library module which provides helpers for running cats-effect applications with config and argument parsing.  Split IOAkkaApp into IOBaseApp
- A new SBT Module + Application + Docker image for network-delayer
- Uses fs2-io to handle the TCP communications
- Uses fs2 to impose throttling
## Testing
- Unit testing of timing-related functionality is difficult
- This feature was manually tested in a local kubernetes cluster
## Tickets
- BN-652

### DIFF
--- a/common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
+++ b/common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
@@ -1,0 +1,118 @@
+package co.topl.common.application
+
+import cats.data.Chain
+import cats.effect._
+import cats.effect.unsafe.IORuntime
+import cats.implicits._
+import cats.kernel.Monoid
+import com.typesafe.config.{Config, ConfigFactory}
+import pureconfig.{ConfigObjectSource, ConfigSource}
+
+import java.nio.file.Paths
+
+/**
+ * Assists with constructing applications which use both cats-effect and akka.  Constructs an
+ * ActorSystem that will be installed as the cats-effect runtime.  Then runs the defined
+ * IO program before cleaning up.
+ * @param createArgs a function which turns stringified command-line args into a structured type
+ * @param createConfig a function which creates a HOCON config using the parsed command-line args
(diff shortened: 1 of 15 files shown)


common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  final def main(args: Array[String]): Unit = {
+    _args = createArgs(args.toList)
+    _config = createConfig(_args)
+    _appConfig = parseConfig(_args, _config)
+    initRuntime()
+    _ioApp = new IOApp {
+      override protected val runtime: IORuntime = _ioRuntime
+      def run(args: List[String]): IO[ExitCode] =
+        IOBaseApp.this.run.as(ExitCode.Success)
```
scasplte2, line 54: I'll admit I am not super familiar with what causes reflection but would the usage of `.this`  be problematic for the native image compilation?

Overall, really like this idea though 👍

scasplte2 (approved):
📦

common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
SeanCheatham: Nope, no reflection concerns with the usage of `this`.  And just as a note, normally `this` would suffice, but because there are two different scopes (`IOBaseApp` and `new IOApp`), both including a `this`, I need to disambiguate the one I need by using `IOBaseApp.this`

PR #2503, opened 2022-10-14
CORE - BN-665 AugmentedBoxState shall correctly adds newBoxes from transaction

o-aleksandrov-topl:
New boxes are correctly added to new boxes, the unit test had been added


SeanCheatham (approved):
Looks good to me!

Only feedback is to use the "PR Template" when creating the PR description (purpose, approach, testing, tickets).  We also generally omit the ticket number from the PR title.

PR #2524, opened 2022-10-15
Tsdk 46 create orient db database 3

mgrand-topl:
## Purpose
Add external configuration and random password generation

## Approach
This is the third incremental pull request for ticket TSDK-46.

Read configuration from a file. If the file does not exist, create it with default contents.

Read the root database password from a file. If it doesn't exist, create it.

The next, and presumably final, pull request will be to define the database schema.

## Testing
Unit tests

## Tickets
* TSDK-46


genus-library/src/main/scala/co/topl/genusLibrary/util/Log.scala
```diff
@@ -8,6 +8,7 @@ import scala.util.{Failure, Try}
  * Functions to make logging compatible with functional programming style
  */
 object Log {
```
scasplte2, line 10: we have typically put typeclasses like this under `co.topl.typeclass` . I would am curious if this is something that might already be standard in Cats though.

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
```diff
@@ -22,25 +22,33 @@ class Genus {
...
-  private implicit val logger: Logger = Logger(classOf[Genus])
+  implicit private val logger: Logger = Logger(classOf[Genus])
 
   /**
-   * This variable refers to the running instance of Genus.
+   * This variable refers to the running instance of Genus. It is an option so that we can call getGenus and shutDown in
+   * any order any number of times.
    */
-  private lazy val instance = new Genus
+  private var instance = Option.empty[Genus]
```
scasplte2, line 31: I would hope to not have mutable state like this in a singleton in production. Unless there is something I am misunderstanding, I believe there are better approaches to handling the application state within an IO app. For discovery this is fine.

scasplte2 (approved):
👀

scasplte2:
Please also run `sbt preparePR` to resolve https://github.com/Topl/Bifrost/actions/runs/3259819090/jobs/5352927100#step:8:272

genus-library/src/main/scala/co/topl/genusLibrary/Genus.scala
mgrand-topl: @scasplte2 The reason for this design decision is that this is a library that is built on top of OrientDB, which is inherently stateful. One of my design goals in designing a library is to avoid unnecessary failure modes, such as
- consecutive calls to the startup function
- consecutive calls to the shutdown function

Also, there is no particular reason that the user should be unable to call startup, shutdown and then call startup again. 
I consider these things that avoid surprising the developer using this library to be good features. If you disagree, I will take this out.

genus-library/src/main/scala/co/topl/genusLibrary/util/Log.scala
mgrand-topl: There is a cats logging package. It has different goals. I have not seen anything else that allows logging to be treated as an identity function with a side-effect.

I will move this to the typeclass package

PR #2527, opened 2022-10-18
BN-620_Extend_TransactionDataLimit

nandotorterolo:
## Purpose
The Transaction.Data field is currently defined as an Option[Sized.Max[Latin1Data, Lengths.`127`.type]].  Some use-cases may require more data than that, and some use-cases may require non-Latin1Data.  The field should be updated to allow for arbitrary byte data of greater length.

## Approach

we should consider making a Transaction Syntax Validation check to impose the 15kb limit, instead of using the Sized type

## Testing
unit test implemented for syntax validator.

## Tickets
https://topl.atlassian.net/browse/BN-620

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
@@ -26,7 +26,8 @@ object TransactionSyntaxValidation {
       scheduleValidation,
       positiveOutputValuesValidation,
       sufficientFundsValidation,
-      proofTypeValidation
+      proofTypeValidation,
+      dataLengthValidation
     )
 
   /**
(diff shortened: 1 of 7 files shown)


models/src/main/scala/co/topl/models/Transaction.scala
```diff
@@ -13,7 +11,8 @@ case class Transaction(
 
 object Transaction {
 
-  type Data = Sized.Max[Latin1Data, Lengths.`127`.type]
+  type Data = Bytes
+  val maxDataLength = 15360
```
SeanCheatham, line 15: Minor detail: Constants like this are usually capitalized i.e. `MaxDataLength`

ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
```diff
@@ -250,4 +251,11 @@ object TransactionSyntaxValidation {
           _ => (TransactionSyntaxErrors.InvalidProofType(proposition, proof): TransactionSyntaxError).invalidNec[Unit]
         )
 
+  private[interpreters] def dataLengthValidation(transaction: Transaction): ValidatedNec[TransactionSyntaxError, Unit] =
+    Validated.condNec(
+      transaction.data.isEmpty ||
```
SeanCheatham, line 256: This can technically be combined into `transaction.data.forall(...)`

ledger/src/main/scala/co/topl/ledger/interpreters/TransactionSyntaxValidation.scala
```diff
@@ -250,4 +251,11 @@ object TransactionSyntaxValidation {
...
+      transaction.data.exists(_.length < Transaction.maxDataLength),
```
SeanCheatham, line 257: Should this be `<=`?

scasplte2 (changes requested):
👍 well done but lets chat about the Dion comment this morning

nandotorterolo:
closing this, and opening v2 PR

PR #2529, opened 2022-10-21
BN-620_Extend_TransactionDataLimit_v2

### LINKED ISSUE #2436: BN-620 ⁃ Extend Transaction.Data Limit:
The Transaction.Data field is currently defined as an `Option[Sized.Max[Latin1Data, Lengths.`127`.type]]`.  Some use-cases may require more data than that, and some use-cases may require non-Latin1Data.  The field should be updated to allow for arbitrary byte data of greater length.

The chosen limit can never be decreased, so whatever new limit is selected, it must not be too large to potentially interfere with consensus.

┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/BN-620) by [Unito](https://www.unito.io)

nandotorterolo:
## Purpose
The Transaction.Data field is currently defined as an Option[Sized.Max[Latin1Data, Lengths.`127`.type]].  Some use-cases may require more data than that, and some use-cases may require non-Latin1Data.  The field should be updated to allow for arbitrary byte data of greater length.

## Approach

we should consider making a Transaction Syntax Validation check to impose the 15kb limit, instead of using the Sized type

## Testing
unit test implemented for syntax validator.

## Tickets
https://topl.atlassian.net/browse/BN-620

Fixes #2436

### DIFF
--- a/blockchain/src/main/scala/co/topl/blockchain/ToplRpcServer.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/ToplRpcServer.scala
@@ -27,7 +27,9 @@ object ToplRpcServer {
       show"NonPositiveOutputValue(value=${outputValue.toString})"
     case TransactionSyntaxErrors.InsufficientInputFunds(_, _) => "InsufficientInputFunds"
     case TransactionSyntaxErrors.InvalidProofType(_, _)       => "InvalidProofType"
-
+    case TransactionSyntaxErrors.InvalidSchedule(s) =>
+      show"InvalidSchedule(creation=${s.creation},maximumSlot=${s.maximumSlot},minimumSlot=${s.minimumSlot})"
+    case TransactionSyntaxErrors.InvalidDataLength => "InvalidDataLength"
   }
 
   /**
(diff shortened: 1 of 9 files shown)


scasplte2:
I would also like us to enforce a transaction wide size limit (not just of the Transaction.Data). This should likely also just be 15kB for not having to define too many numbers? @SeanCheatham thoughts?

nandotorterolo:
> I would also like us to enforce a transaction wide size limit (not just of the Transaction.Data). This should likely also just be 15kB for not having to define too many numbers? @SeanCheatham thoughts?

I will wait to merge for Sean's response, maybe this suggestion should be applied in this PR?

SeanCheatham:
> I would also like us to enforce a transaction wide size limit (not just of the Transaction.Data). This should likely also just be 15kB for not having to define too many numbers? @SeanCheatham thoughts?

While I like the idea of imposing a transaction-wide limit, I think there may be a couple considerations around how we achieve that.  In particular, which mechanism do we use to turn the Transaction into 15kb?  The Transmittable codec may result in varied length serialization once we move to protobuf.  The Signable and Identifiable would be good candidates, but they omit the proofs from the bytes.  So I think we would need a new one (probably just the "Immutable" one, from which Signable and Identifiable are derived).

I think we can merge this PR as-is and open a new ticket to explore the transaction-wide size limit.

PR #2531, opened 2022-10-21
Tetra DataStores Resource Fix

SeanCheatham:
## Purpose
- The underlying LevelDbStore.makeDb returns a Resource, but the consuming implementation uses and immediately closes the resource
- This causes the node to crash upon startup
## Approach
- Modify DataStores.init to return `Resource`
- Modify NodeApp to `.use` the Resource returned by DataStores
## Testing
- Ran node locally and observed proper launch
## Tickets
- N/A

### DIFF
--- a/node-tetra/src/main/scala/co/topl/node/DataStores.scala
+++ b/node-tetra/src/main/scala/co/topl/node/DataStores.scala
@@ -1,8 +1,8 @@
 package co.topl.node
 
-import cats.{Applicative, MonadThrow}
+import cats.{Applicative, Monad, MonadThrow}
 import cats.data.NonEmptySet
-import cats.effect.Async
+import cats.effect.{Async, Resource}
 import cats.implicits._
 import co.topl.algebras.Store
 import co.topl.codecs.bytes.typeclasses.Persistable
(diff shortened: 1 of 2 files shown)


node-tetra/src/main/scala/co/topl/node/NodeApp.scala
```diff
@@ -83,139 +83,144 @@ object NodeApp
 
       cryptoResources <- CryptoResources.make[F]
 
-      dataStores <- DataStores.init[F](appConfig)(bigBangBlock)
-      currentEventIdGetterSetters = new CurrentEventIdGetterSetters(dataStores.currentEventIds)
+      _ <- DataStores
+        .init[F](appConfig)(bigBangBlock)
+        .use(dataStores =>
```
SeanCheatham, line 88: This change simply wraps the rest of the method in a `.use(...)`.  The rest of the code is the same; just indentation changes.

node-tetra/src/main/scala/co/topl/node/DataStores.scala
```diff
@@ -138,25 +115,61 @@ object DataStores {
...
-
+      _ <- Resource.eval(initialize(dataStores, bigBangBlock))
     } yield dataStores
 
   private def makeDb[F[_]: Async, Key: Persistable, Value: Persistable](dataDir: Path)(
     name:                                                                        String
-  ): F[Store[F, Key, Value]] =
-    LevelDbStore.makeDb[F](dataDir / name).use(LevelDbStore.make[F, Key, Value])
+  ): Resource[F, Store[F, Key, Value]] =
+    LevelDbStore.makeDb[F](dataDir / name).evalMap(LevelDbStore.make[F, Key, Value])
```
SeanCheatham, line 124: Main change is here.  Previously, the resource was `.use`'d right here, which would allocate the DB, construct a LevelDbStore interpreter, and then release the DB.  The DB needs to live for the lifetime of the application, so the scope of the `.use` needed to be moved to `NodeApp`.

o-aleksandrov-topl (approved):

PR #2532, opened 2022-10-21
Consensus Testnet Helm Chart

SeanCheatham:
## Purpose
- To test various consensus configurations, we need a way to easily launch a multi-node setup
## Approach
- Added a new helm chart (copied from the existing one)
- Modified for Tetra
- Modified to generate separate resources for each instance of the setup (versus a StatefulSet like the other helm chart)
- Launches one pod per node, with each pod including a node container and a network-delayer container
- Enable node-level configuration, including throttling
## Testing
- Manual testing using local k8s cluster, and "remote" k8s cluster on a separate machine
- Have not tested on GCP k8s yet
## Tickets
- BN-651


scasplte2 (approved):
⛑️

PR #2533, opened 2022-10-23
Bn 675 re implement topl grpc using fs2

nandotorterolo:
## Purpose
Re-Implement ToplGrpc using FS2
## Approach
plugin substitution `sbt-akka-grpc `-> `sbt-fs2-grpc`, on module **topl-grpc**
## Testing
unit test.
## Tickets
https://topl.atlassian.net/browse/BN-675


SeanCheatham (changes requested):
Overall, this looks good!  Just need to move the usages of `.use(...)` up a few levels so that the scope of the Resource is properly captured.

topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
```diff
@@ -26,139 +23,138 @@ object ToplGrpc {
...
+                client
+                  .fetchBlockIdAtDepth(services.FetchBlockIdAtDepthReq(depth), new Metadata())
+                  .map(_.blockId)
+              )
+                .semiflatMap(_.toF[F, bifrostModels.TypedIdentifier])
+                .semiflatMap(EitherT.fromEither[F](_).leftMap(new IllegalArgumentException(_)).rethrowT)
+                .value
+          }
+        )
+        .use(client => Async[F].delay(client))
```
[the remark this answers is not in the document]
nandotorterolo: thanks for reviewing it and providing a detailed solution

scasplte2 (approved):
💧

brambl/src/main/scala/co/topl/client/BramblTetra.scala
```diff
@@ -0,0 +1,113 @@
...
+import co.topl.models.utility.Sized
+import co.topl.typeclasses.implicits._
+import com.typesafe.config.ConfigFactory
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+import scala.concurrent.duration._
+
+object BramblTetra
+    extends IOAkkaApp[Unit, Unit, Nothing](
```
nandotorterolo, line 27: this App was deleted on [Transaction Generator (Part 0)](https://github.com/Topl/Bifrost/pull/2530/files#top)

do we want to keep it? I was no sure during rebase

SeanCheatham: This particular file can be deleted.

node/src/test/scala/co/topl/api/RPCMockState.scala
```diff
@@ -50,7 +50,7 @@ trait RPCMockState
 
   implicit val timeout: Timeout = Timeout(10.seconds)
 
-  implicit protected val routeTestTimeout: RouteTestTimeout = RouteTestTimeout(5.seconds)
+  implicit protected val routeTestTimeout: RouteTestTimeout = RouteTestTimeout(10.seconds)
```
nandotorterolo, line 53: Maybe 6 seconds is enough, let me know if 10 is an aggressive value
failed job: https://github.com/Topl/Bifrost/actions/runs/3315505376/jobs/5476177791
```
[info] AssetTransferRPCSpec:
[info] AssetTransfer RPC should 
[info] - Return correct error response given a threshold transaction with invalid attestation type *** FAILED *** (5 seconds, 41 milliseconds)
[info]   Request was neither completed nor rejected within 5 seconds (AssetTransferRPCSpec.scala:11)
```

SeanCheatham (approved):

node/src/test/scala/co/topl/api/RPCMockState.scala
SeanCheatham: We've adjusted this number a few times, but it doesn't seem to fix the underlying issue.  I think the call is permanently blocking, so it won't work with any timeout.  Not a big deal though - I'm fine with this change

PR #2534, opened 2022-10-25
Fix Monocle Dependency Issue

SeanCheatham:
## Purpose
- A recent PR incorporated the `commonApplication` module into the `brambl` module.  `commonApplication` will only work on Scala 2.13 due to its dependency on a new version of Monocle.  `brambl` must compile to both 2.12 and 2.13
## Approach
- Remove `TetraMempoolReader` utility application from `brambl` and move it into `transactionGenerator`
- Remove `brambl` dependency on `commonApplication`
## Testing
- `preparePR` locally
## Tickets
- BN-671


topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
```diff
@@ -169,13 +169,13 @@ object ToplGrpc {
...
-        .flatMap { serverServiceDefinition =>
+        .flatMap(serverServiceDefinition =>
           NettyServerBuilder
             .forAddress(new InetSocketAddress(host, port))
             .addService(serverServiceDefinition)
-            .resourceWithShutdown[F](server => Async[F].delay(server.shutdown().awaitTermination()))
-            .map(_.start())
-        }
+            .resource[F]
+            .evalMap(server => Async[F].delay(server.start()))
```
nandotorterolo, line 177: thanks for the fix

transaction-generator/src/main/scala/co/topl/transactiongenerator/TransactionGeneratorApp.scala
```diff
@@ -39,37 +38,55 @@ object TransactionGeneratorApp
...
-        .traverse { case (host, port) =>
-          ToplGrpc.Client.make[F](host, port, tls = false)
-        }
-        .use { clients =>
+      clients = ClientAddresses.traverse { case (host, port) =>
+        ToplGrpc.Client.make[F](host, port, tls = false)
+      }
+      // Turn the list of clients into a single client (randomly chosen per-call)
+      _ <- clients
+        .evalMap(MultiToplRpc.make[F, NonEmptyChain])
```
nandotorterolo, line 46: thanks for the fix

nandotorterolo (approved):

topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
SeanCheatham: No problem - nothing was actually broken in this block of code.  I was debugging a separate issue but kept these changes anyway.  `.resource` does essentially what you did with `.resourceWithShutdown`, but instead runs the blocking "shutdown" on a separate thread pool.

transaction-generator/src/main/scala/co/topl/transactiongenerator/TransactionGeneratorApp.scala
SeanCheatham: No issues here, just a small refactor

Issue #680, opened 2020-10-15
Write a script to transition old key format to new key format


scasplte2:

scasplte2:
no longe rapplicable

PR #2535, opened 2022-10-25
Add block header verifications: mismatch timestamp-slot; block from far future

o-aleksandrov-topl:
Add next block header verifications: 
 - Reject block if block's timestamp is not matched slot for that timestamp
 - Reject block if block's slot is greater than current global slot + forward biased slot window

## Purpose
Do not allow to attack blockchain by propagating blocks from far future (which allow to switch to attacker chain, and stop minting block until attacker last block will be reach)

## Approach
Appropriate verification had been added to the block header validation

## Testing
tested by updated unit tests

## Tickets
[*](https://topl.atlassian.net/browse/BN-673)

### DIFF
--- a/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
+++ b/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
@@ -18,6 +18,8 @@ trait ClockAlgebra[F[_]] {
   def currentEpoch: F[Epoch]
   def globalSlot: F[Slot]
   def currentTimestamp: F[Timestamp]
+  def forwardBiasedSlotWindow: F[Slot]
+  def timestampToSlot(timestamp:       Timestamp): F[Slot]
   def delayedUntilSlot(slot:           Slot): F[Unit]
   def delayedUntilTimestamp(timestamp: Timestamp): F[Unit]
 }
--- a/consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
+++ b/consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
@@ -4,7 +4,7 @@ import cats._
 import cats.data._
 import cats.effect.kernel.Sync
 import cats.implicits._
-import co.topl.algebras.{Store, UnsafeResource}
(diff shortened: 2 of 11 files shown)


algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
```diff
@@ -18,6 +18,8 @@ trait ClockAlgebra[F[_]] {
   def currentEpoch: F[Epoch]
   def globalSlot: F[Slot]
   def currentTimestamp: F[Timestamp]
+  def forwardBiasedSlotWindow: F[Slot]
```
SeanCheatham, line 21: While I can see why you would put this value here, I think it should instead be provided as an argument to `BlockHeaderValidation.make` since the field only corresponds to consensus.

consensus/src/main/scala/co/topl/consensus/BlockHeaderValidation.scala
```diff
@@ -73,11 +77,21 @@ object BlockHeaderValidation {
       }.value
 
       private[consensus] def statelessVerification(child: BlockHeaderV2, parent: BlockHeaderV2) =
-        EitherT
-          .pure[F, BlockHeaderValidationFailure](child)
+        for {
```
SeanCheatham, line 80: I think this check should happen outside (after) `statelessVerification` since it requires knowing the state of the genesis block timestamp.

SeanCheatham (approved):
This looks good to me overall; just a couple minor suggestions.  I'd also like @scasplte2 and the research team to review as well

algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
o-aleksandrov-topl: I'm not sure. Looks like it could be considered as related to a consensus as well. Anyway, I'm fine with both solutions.  @scasplte2 what do you think?

hwbehrens (approved):
Looks good to me. We discussed potentially setting the forward bias size to a more carefully-chosen value but will leave that as a task in the backlog for now. The associated Jira ticket is #680. We have set the value to 50 slots for now.

PR #2536, opened 2022-10-25
Transaction Generator App Config

SeanCheatham:
## Purpose
- The TransactionGenerator application/utility currently hardcodes several settings
- This PR introduces some configurability to the node
## Approach
- Move app to `co.topl.transactiongenerator.app`
- Create ApplicationConfig, with corresponding application.conf
- Copy Args from network-delayer
## Testing
- Run node + utility locally
## Tickets
- BN-674


common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
```diff
@@ -64,13 +64,16 @@ object IOBaseApp {
...
     (
-      argsToDebugConfigs(cmdArgs) ++
-        Chain(ConfigSource.resources("environment.conf")) ++
-        argsToUserConfigs(cmdArgs) ++
+      argsToDebugConfigs(cmdArgs).map(_.config()) ++
         Chain(
-          ConfigSource.default
-        )
-    ).foldMapM(_.config()) match {
+          ConfigSource
```
SeanCheatham, line 69: This just makes `environment.conf` optional

scasplte2 (approved):
👍

SeanCheatham:
> Looks good to me,
> 
> maybe we could add another parameter in RPC config
> 
> > clients = ["localhost:9084:true"]
> 
> to allow `tls` support under local testing? maybe it is no possible and this flag in don't applicable

I think this is a good idea.  I updated the PR and implemented the change, but slightly differently.  If the address starts with `https://` (or if no protocol is provided), TLS is enabled.  If the address starts with `http://`, TLS is disabled.  (i.e. `clients = ["http://localhost:9084"]`)

PR #2578, opened 2022-11-09
TSDK-135 | Implement capture of Blocks in Genus [1]

ignaciofernandezsoto:
## Purpose
First iteration on how to implement capture of Blocks in Genus. No implementations, just a foundation.

## Approach
Created the necessary Algebras to get started with the implementations of the capture of blocks.

## Testing
No need to test it as there isn't an implementation yet.

## Tickets
* [TSDK-135](https://topl.atlassian.net/browse/TSDK-135)

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainHeightAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainHeightAlgebra.scala
@@ -1,0 +1,14 @@
+package co.topl.genusLibrary.algebras
+
+/**
+ * Represents the current height of the largest chain in the data store.
+ * @tparam F the effect-ful context to retrieve the value in
+ */
+trait ChainHeightAlgebra[F[_]] {
+
+  /**
+   * Retrieves the current height of the largest chain.
+   * @return the current height
+   */
+  def get: StoreResponse[F, Long]
+}
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainHeightAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainHeightAlgebra.scala
@@ -1,13 +1,13 @@
 package co.topl.genusLibrary.algebras
 
 /**
- * Represents the current height of the largest chain in the data store.
+ * Represents the current height of the chain in the data store.
  * @tparam F the effect-ful context to retrieve the value in
  */
 trait ChainHeightAlgebra[F[_]] {
 
   /**
-   * Retrieves the current height of the largest chain.
+   * Retrieves the current height of the chain in the data store.
    * @return the current height
    */
   def get: StoreResponse[F, Long]
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainUpdaterAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainUpdaterAlgebra.scala
@@ -1,0 +1,15 @@
+package co.topl.genusLibrary.algebras
+
+/**
+ * Updater of the largest chain in the data store.
+ * @tparam F the effect-ful context to retrieve the value in
+ */
+trait ChainUpdaterAlgebra[F[_]] {
+
+  /**
+   * Updates the largest chain in the data store.
+   * @return Unit
+   */
+  def update: F[Unit]
+
+}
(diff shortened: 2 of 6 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainHeightAlgebra.scala
```diff
@@ -0,0 +1,14 @@
+package co.topl.genusLibrary.algebras
+
+/**
+ * Represents the current height of the largest chain in the data store.
```
mgrand-topl, line 4: This is not needed. Genus only stores the canonical blockchain, no forks. What we will need is the ability to find the block at the top of the blockchain. Its height will be the height of the blockchain.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainUpdaterAlgebra.scala
```diff
@@ -0,0 +1,15 @@
+package co.topl.genusLibrary.algebras
```
mgrand-topl, line 1: Also not needed.  We update the height of the blockchain when we store the block at the top of the blockchain.

mgrand-topl (approved):
See my comments.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/ChainUpdaterAlgebra.scala
ignaciofernandezsoto: Also fixed it! Thanks.

This trait should be implemented with a `GenusChainUpdater` class or similar. Its purpose is to execute a background task where it fetches a block at a certain height, saves it, and goes to the next block with the previous height + 1. That's why its `update` method doesn't have any parameters.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
```diff
@@ -0,0 +1,8 @@
+package co.topl.genusLibrary
+
+import co.topl.genusLibrary.failure.Failure
+
+package object algebras {
+  type ServiceResponse[F[_], T] = F[Either[Failure, T]]
+  type StoreResponse[F[_], T] = ServiceResponse[F, T]
```
nandotorterolo, line 7: these two types are the same, which is the benefit of having a Store and Service Response?

nandotorterolo (approved):

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
nandotorterolo: I feel that there are a lot of Algebras with only 1 operation. ChainUpdater, Chain Height, BlockInsert, blockFetcher...
Why not just 1 algebra called "GenusAlgebra" with the 4 operations, and the response type in the same place?
I am not clear about the next Pr, which includes the implementation.
But the class that will implement these algebras, is just 1, 4 or a mix?
I mean: we will we have, a lot of implementation

- GenusImpl with ChainUpdater with  Chain Height with BlockInsert with blockFetcher...
- Genus2Impl  with ChainUpdater with  Chain Height 
- Genus3Impl  with BlockInsert

or just 1?

mgrand-topl: @nandotorterolo I suspect that the implementation will be a mix. The advantage to this approach is on the usage side. 

We will have code that is responsible for capturing blocks and so do mostly inserts, some updates (just for TxOs) and no reads. We will have code that queries only. By making the traits so fine grained we can provide greater certainty that someone will not use an operation in the wrong place

ignaciofernandezsoto: The idea of having both `ServiceResponse` and `StoreResponse` is to have a semantic distinction. A Service response means that it's a response that comes from the network and it's a service layer. It could be either a Node or a third party API. On the other hand, a Store response also means that it's a response that comes from the network, but it's a Store/Repository layer. I think it's a great way to show the developer who may use these methods that the response comes from a DB of some sorts.

About the amount of classes per trait. I wanted to enforce Single-responsibility principle for this implementation, making each trait having a single responsibility. I think it gives the developer much more flexibility on how to implement the traits. For our particular use case, I was thinking of creating a class for each trait. I also suspect concrete methods will be rather large and may be hard to read on a single file.

mgrand-topl: @ignaciofernandezsoto Please put this explanation in a comment. Fernando will not be the last person with this question.

ignaciofernandezsoto: Thank you for your suggestion Mark, I added the comments

PR #2579, opened 2022-11-09
Akka-Stream to FS2 Adapter Delay

SeanCheatham:
## Purpose
- The adapter from an akka-stream Source to an FS2 Stream materializes the Akka stream only once, even if the Stream is consumed multiple times
## Approach
- Delay the adaption from akka to FS2 until the Stream is consumed (instead of pre-adapting it and then consuming)
## Testing
- Verified manually when testing the testnet-simulation-orchestrator
## Tickets
N/A


nandotorterolo (approved):
Thanks for the fix

PR #2582, opened 2022-11-09
txRoot now correctly build transaction root by using Merkle Tree

o-aleksandrov-topl:
## Purpose
Using correct accumulator for transaction root in block header

## Approach
Importing existing implementation from Dion

## Testing
Manually tested that blocks header get correct Merkel tree root hash, additional tests will be added during implementation ticket BN-694

## Tickets
* closes https://topl.atlassian.net/browse/BN-596

### DIFF
--- a/typeclasses/src/main/scala/co/topl/typeclasses/ContainsTransactions.scala
+++ b/typeclasses/src/main/scala/co/topl/typeclasses/ContainsTransactions.scala
@@ -1,7 +1,6 @@
 package co.topl.typeclasses
 
 import cats.Foldable
-import cats.data.Chain
 import cats.implicits._
 import co.topl.codecs.bytes.tetra.TetraIdentifiableInstances._
 import co.topl.codecs.bytes.typeclasses.Identifiable.ops._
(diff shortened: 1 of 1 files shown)


SeanCheatham (approved):

typeclasses/src/main/scala/co/topl/typeclasses/ContainsTransactions.scala
```diff
@@ -1,22 +1,33 @@
...
+import co.topl.models.utility.HasLength.instances._
+import co.topl.models.utility.Lengths._
 import co.topl.models.utility.{Lengths, Sized}
 import co.topl.models.{BlockV1, BloomFilter, Bytes, Transaction}
 import simulacrum.{op, typeclass}
-import co.topl.models.utility.HasLength.instances._
-import Lengths._
-import cats.{Foldable, Traverse}
-import cats.implicits._
+import co.topl.typeclasses.implicits._
```
SeanCheatham, line 18: Cyclic dependency here.  I suggest changing this to `import co.topl.typeclasses.IdentityOps._` to give you access to `.asTypedBytes`

SeanCheatham (approved):

typeclasses/src/main/scala/co/topl/typeclasses/ContainsTransactions.scala
o-aleksandrov-topl: yeap, thanks, going to fix it

PR #2583, opened 2022-11-09
Simulation Orchestrator

SeanCheatham:
## Purpose
- The consensus testnet simulations require some form of central management to produce and collect data
- This manager is called the Orchestrator
## Approach
- A new SBT module with a new Docker application
- The Docker application exists as a Pod inside of a single testnet scenario.  Only one Orchestrator exists per scenario.
- Application reads adoptions from each node in the testnet until a certain height is satisfied on each
- Once the height is reached, the block bodies and transactions are collected
- All of the data is then uploaded to GCP
- Finally, the Orchestrator deletes its own Kubernetes namespace
- NOTE: The TransactionGenerator still needs to be integrated in a follow-up PR
## Testing
- Manual testing on local cluster, publishing data to GCP Storage
## Tickets
- #2554


testnet-simulation-orchestrator/src/main/scala/co/topl/testnetsimulationorchestrator/app/Orchestrator.scala
```diff
@@ -0,0 +1,155 @@
...
+    Logger[F].info(show"Node node=$name is ready")
+
+  private def runSimulation: F[Unit] =
+    resources.use { case (publisher, nodes) =>
+      for {
+        // Listen to the block adoptions of all nodes.
+        nodeBlockAdoptions <- nodes.toList.parTraverse { case (name, client) =>
+          for {
+            _          <- Logger[F].info(show"Fetching adoptions+headers from node=$name")
+            baseStream <- client.synchronizationTraversal()
```
nandotorterolo, line 76: I have a question here. synchronizationTraversal is a stream consumed by the Orchestrator.
This means, in the live of the node, the stream is available, and the orchestrator can use it or not.
Which makes me think that the current implementation is wrong, because we are exposing a trait on RPC, saying, hi this stream is available, (without saying) only 1 consumer is available.
I think we should improve the implementation. (No the simulator orchestrator implementation), the node implementation. (which is my mistake). Lets chat about it

nandotorterolo (approved):

o-aleksandrov-topl (approved):

testnet-simulation-orchestrator/src/main/scala/co/topl/testnetsimulationorchestrator/app/Orchestrator.scala
```diff
@@ -0,0 +1,155 @@
...
+            .map(_.toMap)
+        ).tupled
+      )
+
+  private def awaitNodeReady(name: String, client: ToplRpc[F, Stream[F, *]]) =
+    Logger[F].info(show"Awaiting readiness of node=$name") >>
+    Stream.retry(client.blockIdAtHeight(1), 250.milli, identity, 200).compile.drain >>
+    Logger[F].info(show"Node node=$name is ready")
+
+  private def runSimulation: F[Unit] =
```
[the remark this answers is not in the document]
SeanCheatham: Updated PR and split this method into several sub-methods.

testnet-simulation-orchestrator/src/main/scala/co/topl/testnetsimulationorchestrator/app/ApplicationConfig.scala
```diff
@@ -0,0 +1,46 @@
...
+
+@Lenses
+case class ApplicationConfig(
+  simulationOrchestrator: ApplicationConfig.SimulationOrchestrator
+)
+
+object ApplicationConfig {
+
+  @Lenses
+  case class SimulationOrchestrator(
```
scasplte2, line 19: this is very nice 👍

testnet-simulation-orchestrator/src/main/scala/co/topl/testnetsimulationorchestrator/app/Args.scala
```diff
@@ -0,0 +1,40 @@
...
+import co.topl.common.application.{ContainsDebugFlag, ContainsUserConfigs}
+import mainargs._
+
+@main
+case class Args(startup: Args.Startup)
+
+object Args {
+
+  @main
+  case class Startup(
```
scasplte2, line 13: this is so nice 🐙

scasplte2 (approved):
🧪 🤖

testnet-simulation-orchestrator/src/main/scala/co/topl/testnetsimulationorchestrator/app/Orchestrator.scala
SeanCheatham: I think I experienced strange behavior at one point with 0 consumers but have since fixed that.  With 1 consumer it should all be fine.  And with 2 consumers, I think we're still okay because of the deferred creation of the Stream (which also defers adding a new subscriber to the BroadcastHub).

PR #2595, opened 2022-11-16
TSDK-135 | Added NodeBlockFetcher implementation [2]

ignaciofernandezsoto:
## Purpose
Second iteration on how to implement capture of Blocks in Genus. `BlockFetcherAlgebra` was implemented with `NodeBlockFetcher`. A spec for it was added.

## Approach
The implementation fetches the possible block id that may exist at a given height. If no block id exists, then a Right(None) is returned. In case it exists, then the block header and body are fetched. If any of them don't exist, then a Left(Failure) is returned. If both exist, then the transactions are fetched. If one or more transactions are missing, then a Left(Failure) with all the missing transactions is returned. If all the transactions exist, then the full block is returned.

## Testing

Eight tests were created for this. The spec contemplates all the possible outcomes, to a certain degree. 

## Tickets
* TSDK-135

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/failure/Failure.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/failure/Failure.scala
@@ -1,5 +1,25 @@
 package co.topl.genusLibrary.failure
 
-sealed abstract case class Failure(message: String, exception: Option[Exception])
+import co.topl.models.TypedIdentifier
+import scodec.bits.ByteVector
 
-object Failures {}
+import scala.collection.immutable.ListSet
+
+sealed abstract class Failure(message: String, exception: Option[Exception] = None)
+
+object Failures {
+
+  case class NoPreviousHeaderVertexFailure(blockId: (Byte, ByteVector), parentHeaderId: TypedIdentifier)
+      extends Failure(
+        s"Block doesn't have a previous header vertex. blockId=[$blockId] parentHeaderId=[$parentHeaderId]"
+      )
+
+  case class NoBlockHeaderFoundOnNodeFailure(blockId: TypedIdentifier)
+      extends Failure(s"Block header wasn't found. BlockId=[$blockId]")
+
+  case class NoBlockBodyFoundOnNodeFailure(blockId: TypedIdentifier)
+      extends Failure(s"Block body wasn't found. BlockId=[$blockId]")
+
+  case class NonExistentTransactionsFailure(transactions: ListSet[TypedIdentifier])
+      extends Failure(s"Transactions weren't found. Transactions=[$transactions]")
+}
--- a/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
@@ -1,0 +1,57 @@
+package co.topl.genusLibrary.interpreter
+
+import cats.data.{Chain, EitherT}
+import cats.effect.kernel.Sync
+import cats.implicits._
+import co.topl.algebras.ToplRpc
+import co.topl.genusLibrary.algebras.{BlockFetcherAlgebra, ServiceResponse}
+import co.topl.genusLibrary.failure.{Failure, Failures}
+import co.topl.models.{BlockBodyV2, BlockHeaderV2, BlockV2, Transaction, TypedIdentifier}
+
+import scala.collection.immutable.ListSet
+
+class NodeBlockFetcher[F[_]: Sync](toplRpc: ToplRpc[F, Any]) extends BlockFetcherAlgebra[F] {
+
+  // TODO: TSDK-186 | Do calls concurrently.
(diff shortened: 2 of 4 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/failure/Failure.scala
```diff
@@ -1,5 +1,25 @@
 package co.topl.genusLibrary.failure
 
-sealed abstract case class Failure(message: String, exception: Option[Exception])
+import co.topl.models.TypedIdentifier
+import scodec.bits.ByteVector
 
-object Failures {}
+import scala.collection.immutable.ListSet
+
+sealed abstract class Failure(message: String, exception: Option[Exception] = None)
```
mgrand-topl, line 8: I don't understand this organization. Exceptions have a message. Why do you want to have failure that have a message and an exception which will have its own message? If you are trying to create instance of Failure that are subclasses of scala.util.Failure, that requires an exception. If you are not going for that, why are you allowing exceptions at all?

mgrand-topl (approved):

genus-library/src/test/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcherSpec.scala
```diff
@@ -0,0 +1,357 @@
...
+  TRANSACTION_ID_02,
+  TRANSACTION_ID_03
+}
+import co.topl.models.{BlockBodyV2, BlockHeaderV2, BlockV2, Bytes, Transaction, TypedBytes, TypedIdentifier}
+import munit.FunSuite
+import org.scalamock.munit.AsyncMockFactory
+
+import scala.collection.immutable.ListSet
+
+class NodeBlockFetcherSpec extends FunSuite with AsyncMockFactory {
```
nandotorterolo, line 25: (No blocking comment) would be great if you can switch to CatsEffectSuite with ScalaCheckEffectSuite

nandotorterolo (approved):

genus-library/src/main/scala/co/topl/genusLibrary/failure/Failure.scala
ignaciofernandezsoto: Even though an Exception contains a message, there are times when there's no need for a stack-trace, and [we may consider having errors with valuable information](https://blog.rockthejvm.com/idiomatic-error-handling-in-scala/). Also, if it's an expected failure, **I don't want to throw a raw exception**, doing so [may have a performance hinder](https://users.scala-lang.org/t/efficiency-creating-vs-throwing-an-exception/4226/4). It also provides better logging capabilities, better error display for end users, and it's easier to test with.

I include the possibility to add an exception because developer may want to wrap and bubble up exceptions in a nice to handle custom `Failure` abstract class. Said exceptions are more probable to come from a third party library. An example of it would be the following pseudo-code:

```scala
case class ExpensiveNumericOperationFailure(ex: Exception) extends Failure("Unexpected error happened while doing an expensive numeric operation", Some(ex))
case object ExpensiveNumericOperationWasCanceledFailure extends Failure("Expensive Numeric Operation was canceled") 

val expensiveNumericOperation: F[Int] = 1.pure[F]
val fiber: F[Fiber[F, Throwable, Int]] = Concurrent[F].start(expensiveNumericOperation)
val outcome: F[Outcome[F, Throwable, Int]] = fiber.flatMap(_.join)
val result: F[Either[Failure, Int]] = outcome.flatMap {
    case Succeeded(value) => value.map(_.asRight[Failure])
    case Errored(error) => ExpensiveNumericOperationFailure(error).asLeft[Int].pure[F]
    case Canceled() => ExpensiveNumericOperationWasCanceledFailure.asLeft[Int].pure[F]
}
```

genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
```diff
@@ -0,0 +1,57 @@
...
+import cats.effect.kernel.Sync
+import cats.implicits._
+import co.topl.algebras.ToplRpc
+import co.topl.genusLibrary.algebras.{BlockFetcherAlgebra, ServiceResponse}
+import co.topl.genusLibrary.failure.{Failure, Failures}
+import co.topl.models.{BlockBodyV2, BlockHeaderV2, BlockV2, Transaction, TypedIdentifier}
+
+import scala.collection.immutable.ListSet
+
+class NodeBlockFetcher[F[_]: Sync](toplRpc: ToplRpc[F, Any]) extends BlockFetcherAlgebra[F] {
```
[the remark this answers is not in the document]
ignaciofernandezsoto: Good catch! :-)

genus-library/src/test/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcherSpec.scala
ignaciofernandezsoto: Will check it out! What are the advantages that those trait/abstract class offer?

nandotorterolo: it offers support to create test suites using effects

PR #2614, opened 2022-11-21
Simulation Transactions

SeanCheatham:
## Purpose
- The existing Orchestrator manages a set of nodes in a cluster and reads data from them
- This PR updates the Orchestrator to also generate and broadcast transactions at a configurable frequency
## Approach
- Incorporate TransactionGeneratorApp logic into OrchestratorApp
- Update configs and helm charts to allow configuring the target transactions-per-second
## Testing
- Ran in local k8s cluster and observed data in GCP
## Tickets
- #2583


scasplte2 (approved):
🎫 per 🕐

PR #2619, opened 2022-11-24
TSDK-135 | Added fetchHeaders stream implementation [3]

ignaciofernandezsoto:
## Purpose
Third iteration on how to implement capture of Blocks in Genus. `BlockFetcherAlgebra` has a new method and `NodeBlockFetcher` implements it. It's the `fetchHeaders` method. It's part of an alternate way of returning a full block.

## Approach
The implementation creates a Stream from a given height to "positive infinity", then for each height the block id is fetched. If no block id exists, then a `Right(None)` element is returned. In case it exists, then the block header is fetched. If it exists, a `Right(Some(header))` is returned. If it doesn't then a `Left(Failure)` element is returned.

## Testing
No tests yet. Awaiting review of design.

## Tickets
* TSDK-135

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
@@ -1,15 +1,40 @@
 package co.topl.genusLibrary.algebras
 
-import co.topl.models.BlockV2
+import co.topl.models.{BlockHeaderV2, BlockV2}
 
 /**
  * Fetcher of blocks on the chain.
  * @tparam F the effect-ful context to retrieve the value in
+ * @tparam G sequence container. Ex: Stream, Seq
  */
-trait BlockFetcherAlgebra[F[_]] {
+trait BlockFetcherAlgebra[F[_], G[_]] {
+
+  /**
+   * Look-up block headers on the chain, starting from a given inclusive height.
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  def fetchHeaders(height: Long): SequenceResponse[F, G, Option[BlockHeaderV2]]
+
+  /**
+   * Look-up block bodies on the chain, starting from a given inclusive height.
+   *
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  // def fetchBodies(height: Long): SequenceResponse[F, G, Option[BlockBodyV2]]
+
+  /**
+   * Look-up block transactions on the chain, starting from a given inclusive height.
+   *
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  // def fetchTransactions(height: Long): SequenceResponse[Option[BlockBodyV2.Full]]
 
   /**
    * Look-up a block on the chain with a given height
+   *
    * @param height The height to lookup
    * @return the full block
    */
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
@@ -15,4 +15,6 @@ package object algebras {
    */
   type ServiceResponse[F[_], T] = F[Either[Failure, T]]
   type StoreResponse[F[_], T] = ServiceResponse[F, T]
+
+  type SequenceResponse[F[_], G[_], T] = F[G[Either[Failure, T]]]
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
@@ -4,13 +4,29 @@ import cats.data.{Chain, EitherT}
 import cats.effect.kernel.Async
 import cats.implicits._
 import co.topl.algebras.ToplRpc
-import co.topl.genusLibrary.algebras.{BlockFetcherAlgebra, ServiceResponse}
+import co.topl.genusLibrary.algebras.{BlockFetcherAlgebra, SequenceResponse, ServiceResponse}
 import co.topl.genusLibrary.failure.{Failure, Failures}
 import co.topl.models.{BlockBodyV2, BlockHeaderV2, BlockV2, Transaction, TypedIdentifier}
+import fs2.Stream
 
 import scala.collection.immutable.ListSet
 
-class NodeBlockFetcher[F[_]: Async](toplRpc: ToplRpc[F, Any]) extends BlockFetcherAlgebra[F] {
+class NodeBlockFetcher[F[_]: Async](toplRpc: ToplRpc[F, Any]) extends BlockFetcherAlgebra[F, Stream[F, *]] {
+
+  override def fetchHeaders(height: Long): SequenceResponse[F, Stream[F, *], Option[BlockHeaderV2]] = Async[F].delay {
+    Stream
+      // Range from given height to "positive infinity". If height is one, then the range would be [1, 2, 3, ...]
+      .range(height, Long.MaxValue)
+      .covary[F]
+      // Map each height to a block id. Range ends up being [blockId_01, blockId02, blockId03, ...]
+      .evalMap(toplRpc.blockIdAtHeight)
+      .evalMap {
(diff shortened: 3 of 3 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
```diff
@@ -1,15 +1,40 @@
...
+   */
+  // def fetchBodies(height: Long): SequenceResponse[F, G, Option[BlockBodyV2]]
+
+  /**
+   * Look-up block transactions on the chain, starting from a given inclusive height.
+   *
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  // def fetchTransactions(height: Long): SequenceResponse[Option[BlockBodyV2.Full]]
```
mundacho, line 33: remove commented code

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
```diff
@@ -1,15 +1,40 @@
...
-import co.topl.models.BlockV2
+import co.topl.models.{BlockHeaderV2, BlockV2}
 
 /**
  * Fetcher of blocks on the chain.
  * @tparam F the effect-ful context to retrieve the value in
+ * @tparam G sequence container. Ex: Stream, Seq
  */
-trait BlockFetcherAlgebra[F[_]] {
+trait BlockFetcherAlgebra[F[_], G[_]] {
```
mgrand-topl, line 10: Why add the complexity of G? You don't need to return a stream inside something else. If there is a problem creating a stream, log the problem and return an empty stream rather than make the failure something that the caller needs to deal with.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
```diff
@@ -1,15 +1,40 @@
...
+
+  /**
+   * Look-up block headers on the chain, starting from a given inclusive height.
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  def fetchHeaders(height: Long): SequenceResponse[F, G, Option[BlockHeaderV2]]
```
mgrand-topl, line 17: In what use case will you be using this? When do you want to fetch just headers?

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
```diff
@@ -15,4 +15,6 @@ package object algebras {
    */
   type ServiceResponse[F[_], T] = F[Either[Failure, T]]
   type StoreResponse[F[_], T] = ServiceResponse[F, T]
+
+  type SequenceResponse[F[_], G[_], T] = F[G[Either[Failure, T]]]
```
mgrand-topl, line 19: Again, why to you want to introduce the complication. If there is a failure in getting a sequence of results, what do you expect the caller to do about it? Why not just return an empty sequence?

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
nandotorterolo: I discuss the same concern with Sean a while ago, and he convinced me, you can see it on toplRpc algebra
For sure, Sean's input is would be appreciated again.

given 

- F – Effect type
- S –  Container, Ex: Stream, Seq

and an interface 
>   def yourInterface(): F[S[Body]]

allows the implementation to provide a context and a container, If you only provide a context, lets say
>   def yourInterface(): F[Body]

then in the application, you will create an application using (Resource, or IO effect), and then, when you want to consume this Container,  you will need to wrap the container  with the app context
```
yourApp[F: Async] {
  stream <- Async[F].delay(def yourInterface(): Stream[Body]) // without context
  stream <-  yourInterface(): F[Stream[Body]])  // with context
}
```

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
nandotorterolo: Maybe the failure is useful if it indicates the current height where the Exception was thrown.
So, if we log, inspect the error, and fix it, in the next call the consumer will start from there, and not from the beginning. I am not clear about the requirements, and retrying this call every time from scratch is not too bad

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
mgrand-topl: @nandotorterolo The difference is that for ToplRpc the caller does want to know about errors. If we are doing a gossip protocol the caller of the interpreter wants to know how well each of the peers it communicates with is carrying out their end of the conversation. This is so that it can adaptively decide which peers to choose for communication. Value is added by making the caller aware of errors or even the quality of a successful connection.

Genus does not care. The caller has no decisions to make about which nodes to communicate with. Each block needs to be fetched, but it does not matter which node it comes from. The caller does not need to get involved. If a connection to a node fails, the interpreter can just select an alternate node without any coordination from the caller. In the case of Genus, there is no added value by returning errors.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
mgrand-topl: For Genus, what matters is what blocks have made it into the database. Why they are in the database or not in the database does not matter.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
scasplte2: I think there is still value in returning Errors if the server is able to identify why a problem happened. It may also help to have a single standard approach to allow for the composition of multiple `F[G[_]]` functions without having to have middleware code between them.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
scasplte2: the node denormalizes the data, so I want to present that interface to the client. There is no reason to hide additional service logic in the API layer when the client may want the data and can process the separate pieces themselves

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
```diff
@@ -1,15 +1,40 @@
...
+
+  /**
+   * Look-up block bodies on the chain, starting from a given inclusive height.
+   *
+   * @param height The height to start the lookup. It's inclusive.
+   * @return a sequence container of headers
+   */
+  // def fetchBodies(height: Long): SequenceResponse[F, G, Option[BlockBodyV2]]
```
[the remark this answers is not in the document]
scasplte2: are we just not implementing them or do we not want to implement these?

genus-library/src/main/scala/co/topl/genusLibrary/algebras/package.scala
scasplte2: I don't understand why handling the error is considered a complication though @mgrand-topl

scasplte2 (approved):

genus-library/src/main/scala/co/topl/genusLibrary/algebras/BlockFetcherAlgebra.scala
SeanCheatham: I also still prefer `F[G[_]]`.  While we may intend to use `Stream` as the `G` in our main interpreter, the algebra doesn't dictate that.  It would be just as valid to have an interpreter use `List` as the `G` type parameter, in which case, having the additional `F[_]` context is helpful in "preparing" the List.  Adding the `F` type helps with uniformity and doesn't really impose much burden, so long as the rest of the algebras are also consistent in returning `F` for everything.

ignaciofernandezsoto:
Code will be reworked in a new branch. Implementation will be different as [the documentation](https://topl.atlassian.net/wiki/spaces/TSDK/pages/492273671/Capture+of+Historical+Blocks+in+Genus) has described

PR #2584, opened 2022-11-09
BN-690-Re-imlementLocalChainBroadCaster V1 (Do NOT Merge)

nandotorterolo:
## Purpose
BN-690- Re-implementLocalChainBroadCaster V1

## Approach

LocalChainBroadcaster returns:
- local chain algebra 
-  Akka Source of adaptions

it is used on: Akka source is used on 
 - BlockchainPeerServer,  
 -  mintedBlockStream
 - ToplGrpc (with an FS2 converter)

Changes adds a new Fs2 Stream in LocalChainBroadcaster, returning
- local chain algebra 
-  Akka Source of adaptions
- Fs2 Stream for ToplRpcServer  (replace the converter)
- Fs2 Stream for BlockchainPeerServer  (unused, it will be implemented in other PR)

## Tickets
BN-690

PR #2620, opened 2022-11-28
BN-690-Re-imlementLocalChainBroadCaster-V2

nandotorterolo:
## Purpose
BN-690- Re-implementLocalChainBroadCaster V2
This is other implementation of LocalChainBroadCaster, which does not create queues, (https://github.com/Topl/Bifrost/pull/2584)

## Approach

LocalChainBroadcaster :
- input: adoptions Topic, which can be subscribed on consumers later (grpc, BlockchainPeerServer)

## Testing
@SeanCheatham in your next run of transactions, orchestration run? is it possible that you can base on this branch to see if the behavior works as expected?

## Tickets
BN-690


scasplte2 (approved):
🔥

SeanCheatham (approved):
This looks good and I was able to run it successfully in a testnet simulator :+1: 

Just the one detail that we discussed in Slack about a potentially full queue/topic and a slow downstream subscriber.  Could you add a TODO (or error out the subscriber streams if they can't keep up)?

PR #2623, opened 2022-11-29
gRPC Web + Faucet

SeanCheatham:
## Purpose
- Connecting to a gRPC service from a browser requires special transformation called gRPC-web.  Envoy was the initial implementation approach, but it was unsuccessful in initial testing.  Eventually we want to use it, but in the meantime, embedding grpc-web directly into the service works
- For testing Ribn, testers need the ability to give funds to a wallet through a faucet
## Approach
- Incorporate akka-grpc-web into Genus
- Added a PrivateTestnetFaucet utility which accepts command-line arguments to fund a wallet
## Testing
- `docker compose up` followed by usage of the Faucet was successful
## Tickets
- #2588


node/src/main/scala/co/topl/PrivateTestnetFaucet.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.Success
+
+/**
+ * A simple application which serves as a "faucet" for a private blockchain testnet.  A private testnet is launched with
+ * funds available at pre-seeded addresses.  This utility creates a Transaction which spends those funds and moves a
+ * requested quantity to a requested address. It then broadcasts the transaction to a requested node (via json-rpc).
+ */
+object PrivateTestnetFaucet extends Logging {
```
scasplte2, line 30: oh my goodness, this is such a good idea. Thank you @SeanCheatham

scasplte2 (approved):
🐙

node/src/main/scala/co/topl/PrivateTestnetFaucet.scala
SeanCheatham: Born out of necessity :laughing: :potable_water:

PR #2627, opened 2022-12-01
Delete Tetra Modules

SeanCheatham:
## Purpose
- The `dion` branch does not need to keep any `tetra` code, except in cases where the models and codecs are used
## Approach
- Delete modules that are specific to Tetra.  Kept modules that are shared by Dion and Tetra.
## Testing
- N/A
## Tickets
- #2588


nandotorterolo (approved):

scasplte2 (approved):
💾 🤖 📚 🎒

PR #2630, opened 2022-12-02
BN-582-2313-submodule-integration

### LINKED ISSUE #2313: BN-582 ⁃ Modify the contents of the Bifrost repo to use the gRPC_specs repo
┆Issue is synchronized with this [Jira Task](https://topl.atlassian.net/browse/BN-582) by [Unito](https://www.unito.io)

nandotorterolo:
## Purpose
#2313 ⁃ Modify the contents of the Bifrost repo to use the gRPC_specs repo

## Approach
- A git module was versioned
- Old protobuf files are under protobuf_old directory,  keeping to compare version, I will remove later 
- add minting again in box.
- remove AssetV1BoxValue.Code.version in bifrost models

## Testing
unit test

## Tickets
closes #2313


nandotorterolo:
@admiraladmirable I need you help, adding a github action to setup a git-submodule

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "topl-grpc/src/main/protobuf"]
+	path = topl-grpc/src/main/protobuf
+	url = git@github.com:Topl/protobuf-specs.git
```
nandotorterolo, line 3: @admiraladmirable I need add a git submodule using the following command:
And I don't know how to setup it up the pipeline
Also, I am not sure, if this file should be versioned.
This was generated on my local

 > git submodule add -f git@github.com:Topl/protobuf-specs.git ./topl-grpc/src/main/protobuf

admiraladmirable: This is pretty easy to implement for the GitHub actions, just search for actions matching `- uses: actions/checkout` and add `submodules: true` (unless that submodule uses submodules, then we need to use `submodules: recursive`. https://github.com/marketplace/actions/checkout

An example of this: in [_sbt_build.yml](https://github.com/Topl/Bifrost/blob/dev/.github/workflows/_sbt_build.yml)

```
- name: Checkout current branch
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
          submodules: true
```

I am slightly worried that an update to the referenced repo can suddenly break this repo, and it doesn't appear that submodules support pinning to tags or commit refs.

We should probably look into a branching strategy for that repo since submodules do support targeting branches.

Also, it would be a good idea to update the build docs to make sure users know to run the `git submodule` commands so they can build.

nandotorterolo: Mark: we will need to take care of realeased version numbers. hash, tag,.... TBD

nandotorterolo: set up a meeting. TODO

SeanCheatham (approved):
:electric_plug:

.gitmodules
nandotorterolo: @admiraladmirable I will merge, we will discuss on a meet your concerns, and then implement changes in futures PRs

PR #2655, opened 2022-12-13
BN-714-2635-replace-models-with-protobuf v1

nandotorterolo:
## Purpose
Tetra defines a package models that serve as pure Scala ADTs and offer very minimal functionality.  We have since adopted protobuf to define the models of Topl’s ecosystem since it leads to accurate code generation.  Clients will use generated models from protobuf, so the protocol/server/node would benefit from referencing the same models.

## Approach

- [x] - Create new `protobuf` SBT module
- [x] - Move protobuf generation from `topl-grpc` to `protobuf` module. Update `topl-grpc` to depend on `protobuf`
- [ ] - Delete `models` SBT module
- [ ] - Update old references to `models` to use the generated models from `protobuf`

Pr v2 and v3 will address pendings items

## Testing
unit test checks

## Tickets

https://topl.atlassian.net/browse/BN-714

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "topl-grpc/src/main/protobuf"]
-	path = topl-grpc/src/main/protobuf
+[submodule "protobuf/src/main/protobuf"]
+	path = protobuf/src/main/protobuf
 	url = git@github.com:Topl/protobuf-specs.git
--- a/build.sbt
+++ b/build.sbt
@@ -207,7 +207,8 @@ lazy val bifrost = project
     crypto,
     catsAkka,
 //    brambl,
-    models,
+    models, // TODO remove BN-714 PR v2
+    protobuf,
     numerics,
     eventTree,
     algebras,
@@ -272,7 +273,7 @@ lazy val nodeTetra = project
   .configs(IntegrationTest)
   .settings(
     IntegrationTest / parallelExecution := false,
-    classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat //required for correct loading https://github.com/kamon-io/sbt-kanela-runner
+    classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat // required for correct loading https://github.com/kamon-io/sbt-kanela-runner
   )
   .dependsOn(
     models % "compile->compile;test->test",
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "topl-grpc/src/main/protobuf"]
-	path = topl-grpc/src/main/protobuf
+[submodule "protobuf/src/main/protobuf"]
```
nandotorterolo, line 1: change to https

project/Dependencies.scala
```diff
@@ -329,9 +329,13 @@ object Dependencies {
     Seq(akka("actor"), akka("actor-typed"), akka("stream")) ++
     Seq(fs2Core, fs2IO, fs2ReactiveStreams)
 
+  // TODO remove BN-714, PR v2
   lazy val models: Seq[ModuleID] =
     cats ++ simulacrum ++ newType ++ scodec
 
+  lazy val protobuf: Seq[ModuleID] =
```
o-aleksandrov-topl, line 336: do we really need simulacrum for protobuf?

build.sbt
```diff
@@ -417,6 +419,36 @@ lazy val models = project
   )
   .settings(libraryDependencies ++= Dependencies.test)
 
+// A task (meant to be run before compile) which ensures that the protobuf-specs git submodule is up-to-date
+// The protobuf specifications are defined in a remote repository and are downloaded locally for compilation.
+lazy val updateSubmodulesTask = TaskKey[Unit]("updateSubmodules", "Update git submodules")
+
+lazy val protobuf = project
+  .in(file("protobuf"))
+  .enablePlugins(BuildInfoPlugin, Fs2Grpc)
```
o-aleksandrov-topl, line 428: do we need Fs2 for protobuf here?

project/Dependencies.scala
nandotorterolo: yes and no. Right now, dependencies for protobuf new model are the same as models. Because following Prs, I will move progressively the utils and classes needed.  Then, in the final PR, I will clean up unused dependencies. thanks

build.sbt
nandotorterolo: Fs2 plugin is required for step 2: "Move protobuf generation from topl-grpc to protobuf module"

o-aleksandrov-topl (approved):

mgrand-topl (approved):

.gitmodules
nandotorterolo: pr v2

PR #2691, opened 2023-01-04
[Do Not Merge] Use BramblSc/crypto

SeanCheatham:
## Note
- This PR currently depends on a BramblSc/crypto via jitpack on a hard-coded commit hash.  This PR should not be merged until that particular commit hash makes its way into the `main` branch

## Purpose
- User-level cryptography is now defined in BramblSc/crypto, and Bifrost does not need to duplicate the implementation
- Bifrost now depends on BramblSc/crypto for the user-level crypto operations
## Approach
- Use jitpack to consume BramblSc as a library of Bifrost's `crypto` module
- Delete any code from Bifrost that now exists in BramblSc
- Update any references to the old type signatures
## Testing
- `preparePR` locally
## Tickets
- #2690


project/Dependencies.scala
```diff
@@ -184,6 +184,9 @@ object Dependencies {
   val circeYaml = "io.circe"               %% "circe-yaml"           % "0.14.1"
   val kubernetes = "io.kubernetes"          % "client-java"          % "16.0.1"
 
+  val bramblScCrypto =
+    "com.github.Topl" % "BramblSc" % "9034b442"
```
mgrand-topl, line 188:
```suggestion
    "com.github.Topl" % "BramblSc" % "v2.0.3"
```

mgrand-topl (approved):

nandotorterolo (approved):

crypto/src/main/scala/co/topl/crypto/crypto.scala
```diff
@@ -1,50 +1,18 @@
...
-
 import java.security.SecureRandom
-import scala.language.implicitConversions
 
 package object crypto {
 
   def defaultRandom: SecureRandom = defaultRandom(None)
 
-  def defaultRandom(seed: Option[Seed]): SecureRandom = {
+  def defaultRandom(seed: Option[Array[Byte]]): SecureRandom = {
```
nandotorterolo, line 9: maybe we can remove newType io.statico) from crypto Module Dependencies

project/Dependencies.scala
SeanCheatham: Updated

crypto/src/main/scala/co/topl/crypto/crypto.scala
SeanCheatham: newType is a transitive dependency through the `models` module at the moment.  But I was able to remove several other unused dependencies from `crypto` :+1:

PR #2728, opened 2023-01-11
TSDK-212 Modified algebras and added util methods

ignaciofernandezsoto:
## Purpose
Developed utils and tests for it. Updated algebras to reflect those changes.

## Approach
-> As the Graph Wrapper had a lot of implementation details and was hard to test because of the OrientGraph library field, I modified the wrapper to become a DAO that has a **wrapped** OrientGraph library field. That way, the responsabilities of the now DAO are more clearer and easier to test, as mocking a library class is really difficult and a **bad practice**.
-> Created WrappedVertex
-> Created WrappedEdge
-> Updated Algebras to reflect refactor

## Testing
-> Added tests for GraphTxDAO

## Tickets
* TSDK-212

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
@@ -1,5 +1,6 @@
 package co.topl.genusLibrary.algebras
 
+import co.topl.genusLibrary.failure.Failure
 import co.topl.genusLibrary.model.BlockData
 
 /**
@@ -14,6 +15,6 @@ trait HeaderInserter[F[_]] {
    * @param block  the full block data for context reasons
    * @return unit
    */
-  def insert(block: BlockData): F[Unit]
+  def insert(block: BlockData): F[Either[Failure, Unit]]
 
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
@@ -1,6 +1,5 @@
 package co.topl.genusLibrary.algebras
 
-import co.topl.genusLibrary.failure.Failure
 import co.topl.genusLibrary.model.BlockData
 
 /**
@@ -15,8 +14,6 @@ trait HeaderInserter[F[_]] {
    * @param block  the full block data for context reasons
    * @return unit
    */
-  def insert(
-    block: BlockData
-  ): F[Either[Failure, Unit]]
+  def insert(block: BlockData): F[Unit]
 
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
@@ -3,7 +3,6 @@ package co.topl.genusLibrary.algebras
 import co.topl.genusLibrary.Txo
 import co.topl.genusLibrary.failure.Failure
 import co.topl.genusLibrary.model.BlockData
-import co.topl.genusLibrary.orientDb.GraphTxDAO
 import co.topl.models.{BlockBody, Transaction}
 import co.topl.proto.models.TypedEvidence
 
@@ -21,7 +20,7 @@ trait Mediator[F[_]] {
    * @param graph graph in which to commit transaction
    * @return Unit
    */
-  def afterHeaderInserted(block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterHeaderInserted(block: BlockData): F[Either[Failure, Unit]]
 
   /**
    * Mediate after insertion of body.
@@ -31,7 +30,7 @@ trait Mediator[F[_]] {
    * @param graph  graph in which to commit transaction
    * @return Unit
    */
-  def afterBodyInserted(body: BlockBody, block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterBodyInserted(body: BlockBody, block: BlockData): F[Either[Failure, Unit]]
 
   /**
    * Mediate after insertion of transaction.
@@ -41,7 +40,7 @@ trait Mediator[F[_]] {
    * @param graph  graph in which to commit transaction
    * @return Unit
    */
-  def afterTxInserted(transaction: Transaction, block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterTxInserted(transaction: Transaction, block: BlockData): F[Either[Failure, Unit]]
 
   /**
    * Mediate after insertion of address.
@@ -51,7 +50,7 @@ trait Mediator[F[_]] {
    * @param graph  graph in which to commit transaction
    * @return Unit
    */
-  def afterAddressInserted(address: TypedEvidence, block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterAddressInserted(address: TypedEvidence, block: BlockData): F[Either[Failure, Unit]]
 
   /**
    * Mediate after insertion of transaction output.
@@ -61,7 +60,7 @@ trait Mediator[F[_]] {
    * @param graph  graph in which to commit transaction
    * @return Unit
    */
-  def afterTxoInserted(txo: Txo, block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterTxoInserted(txo: Txo, block: BlockData): F[Either[Failure, Unit]]
 
   /**
    * Mediate after insertion of address state.
@@ -70,6 +69,6 @@ trait Mediator[F[_]] {
    * @param graph in which to commit transaction
    * @return Unit
    */
-  def afterAddressStateInserted(block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
+  def afterAddressStateInserted(block: BlockData): F[Either[Failure, Unit]]
 
 }
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
@@ -17,12 +17,11 @@ trait Mediator[F[_]] {
   /**
    * Mediate after insertion of header. Commits DB transaction.
    *
-   * @param header inserted block header
    * @param block the full block data for context reasons
    * @param graph graph in which to commit transaction
    * @return Unit
    */
-  def afterHeaderInserted(header: BlockHeader, block: BlockData, graph: GraphTxWrapper): F[Either[Failure, Unit]]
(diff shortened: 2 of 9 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/orientDb/GraphTxDAO.scala
```diff
@@ -0,0 +1,104 @@
...
+import scala.util.{Success, Try}
+
+/**
+ * Even though Tinkerpop's API is very high level and great in design for a Java application,
+ * there are some use cases that feel a bit low level on our end, may induce duplicate code and/or bad practices.
+ * @param wrappedGraph Wrapped Tinkerpop's implementation of an TX Orient Graph.
+ */
+class GraphTxDAO[F[_]: Async](wrappedGraph: GraphTxWrapper) {
+
+  implicit private val logger: Logger[F] =
```
mgrand-topl, line 19: We don't need a logger instance for each instance of this class. The usual way to create the logger instance is to make it a field of the companion object and import it into the class.  There is an example in Genus.scala

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+  private val wrappedGraph = mock[GraphTxWrapper]
+
+  private val graphTxDao = new GraphTxDAO[F](wrappedGraph)
+
+  test("On a transaction with a failing function return, no methods on the graph are executed") {
+    withMock {
+
+      val request = Failures.NonExistentTransactionsFailure(ListSet.empty).asLeft[Any].pure[F]
+
+      graphTxDao.withEffectfulTransaction(request)
```
mgrand-topl, line 26: The ScalaMock documentation says, "By default, mocks expect exactly one call while stubs allow any number of calls." Don't you need something here to tell it to check that the graph methods are not called?

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    }
+  }
+
+  test("On a transaction with a successful function return and failing commit, a rollback should be executed") {
+    withMock {
+
+      val request = 2.asRight[Failure].pure[F]
+
+      (wrappedGraph.commit _)
```
mgrand-topl, line 36: Should there be something here to require the ordering that the commit is called before the rollback?

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+        .expects()
+        .returns(())
+        .once()
+
+      graphTxDao.withEffectfulTransaction(request)
```
mgrand-topl, line 60: How is this checking that rollback is not called?

mgrand-topl (approved):

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/GraphTxDAO.scala
```diff
@@ -0,0 +1,104 @@
...
+    inVertex:  WrappedVertex,
+    label:     Option[String] = None
+  ): WrappedEdge =
+    wrappedGraph.addEdge(null, outVertex, inVertex, label.orNull)
+
+  private def withTransaction[T <: Any](transactionalFun: Either[Failure, T]): F[Unit] =
+    transactionalFun match {
+      case Left(ex) => Logger[F].error(s"Something went wrong with the transaction function, won't commit. Error=[$ex]")
+      case Right(_) =>
+        Logger[F]
```
SeanCheatham, line 89: Suggested alternative:

```
Logger[F].info("Comitting transaction") >>
  EitherT(Sync[F].delay(Try(wrappedGraph.commit()).toEither))
    .foldF(
      ex => Logger[F].error(ex, "Something went wrong while trying to commit transaction.") >> Sync[F].delay(wrappedGraph.rollback()),
      _ => Logger[F].info("Transaction committed")
    )
```

- Logger[F].error accepts a Throwable as an argument
- The usage of "Try" can be wrapped up as an Either and used in an EitherT.  Alternatively, `ApplicativeError.catchNonFatal` could be considered

SeanCheatham: Also, it may be the case that `Sync[F].delay(...)` will handle catching exceptions, which omits the need for a `Try(...)`.  I don't remember exactly though.  If it does behave that way, then you could `handleErrorWith` to apply the logic you'd like (logging, rollback)

genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
```diff
@@ -15,8 +14,6 @@ trait HeaderInserter[F[_]] {
    * @param block  the full block data for context reasons
    * @return unit
    */
-  def insert(
-    block: BlockData
-  ): F[Either[Failure, Unit]]
+  def insert(block: BlockData): F[Unit]
```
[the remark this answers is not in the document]
SeanCheatham: Errors that would apply to _all_ interpretations of this algebra can be enumerated and surfaceced in the return type of this method.

But errors that are specific to one particular interpreter (i.e. a GenusHeaderInserter) can be encapsulated locally to that interpreter.  Failures can be logged and retried as needed, by that particular interpreter.

But there may also be an interpreter that just writes headers into an in-memory `Map` collection, in which case, failures wouldn't happen like they would in a `GenusHeaderInserter`.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
```diff
@@ -15,62 +15,61 @@ import co.topl.proto.models.TypedEvidence
...
-   * Mediate after insertion of header. Commits DB transaction.
+   * Mediate after insertion of header.
    *
-   * @param header inserted block header
    * @param block the full block data for context reasons
    * @param graph graph in which to commit transaction
    * @return Unit
    */
-  def afterHeaderInserted(header: BlockHeader, block: BlockData, graph: GraphTxWrapper): F[Either[Failure, Unit]]
+  def afterHeaderInserted(block: BlockData, graph: GraphTxDAO[F]): F[Either[Failure, Unit]]
```
SeanCheatham, line 24: Could the `graph` parameter be omitted from these methods and instead moved into the interpreter's `make` method?

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/GraphTxDAO.scala
SeanCheatham: When using effectful logging patterns, the `F` type needs to be encoded in `Logger` instance (which is different from when using the raw slf4j Logger API).  Because the effect type is encoded in the Logger, it can't be defined in the companion object.

SeanCheatham (approved):

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/GraphTxDAO.scala
SeanCheatham: I use a mixture of context-bound constraints (`F[_] : Logger`) and creating a new `implicit private val logger` depending on the circumstance.  For logical components (like an interpreter), I create a new `implicit private val logger`.  When calling utility functions, I use `F[_]: Logger`.

It's a matter of how you want it to show up in the logs.  If you want the log statement to show up under `GraphTxDao`'s name, you would use `implicit private val logger`.  If you want the log statement to show up under the parent context's name, then you can accept an implicit logger `F[_]: Logger`

mgrand-topl: I am a fan of the fluent logging style supported by co.topl.typeclasses.Log. My advice applies to that style of logging. If you prefer to use the effectful style of logging, then do as Sean says.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/HeaderInserter.scala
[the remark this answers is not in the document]
ignaciofernandezsoto: I changed the signature to `def insert(block: BlockData): F[Either[Failure, Unit]]` and the `Failure` abstract class is no longer sealed. That way, developers will be able to create different errors for their particular use case.

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
ignaciofernandezsoto: As I'm using an [expectation-first style for mocking](https://scalamock.org/user-guide/mocking_style/), Every method that has no expectations is _implicitly_ being verified that it isn't called.

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
ignaciofernandezsoto: Good catch! I added `inSequence` method execution on the spec. It's from [ScalaMock - Ordering](https://scalamock.org/user-guide/ordering/)

genus-library/src/test/scala/co/topl/genusLibrary/orientDb/GraphTxDAOSpec.scala
ignaciofernandezsoto: As I'm using an [expectation-first style for mocking](https://scalamock.org/user-guide/mocking_style/), the rollback method is _implicitly_ being verified that it should not be called.

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/GraphTxDAO.scala
ignaciofernandezsoto: Great suggestion @SeanCheatham! I had to modify it a bit so it contemplates error handling.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/Mediator.scala
ignaciofernandezsoto: I was thinking of the flow like this:

Element being Block, Header, whatever node we wish to insert. Pseudo-code ahead.

```scala
class ElementInserter(orientDb: OrientDb, mediator: Mediator) {
    def insert(element: Element) = {
        graph = orientDb.getGraph()
        graph.withTransaction {
            graph.createVertex(element)
            ...
            mediator.mediate(element, graph)
        }
    }
}
class Mediator() {
    def mediate(element: Element, graph: Graph) = {
        graph.createVertex(...)
        ...
        graph.createEdge(...)
    }
}
```

That way I can reuse the same graph to insert the element and its related elements through the mediator. It could also easily be two separate transactions.

I've given it a bit more thought and I think it'd be easier to test and develop if it were two separate transactions. Thanks @SeanCheatham

PR #2729, opened 2023-01-12
Mempool Rebroadcast Fix

SeanCheatham:
## Purpose
- An issue appears when multiple nodes try to exchange transactions due to an issue when converting between FS2 and akka-stream
  - The FS2 unicast stream was allocated to construct an akka-stream, but the FS2 stream was immediately closed
- Tracing the issue down led to some substantial refactoring of the Blockchain Peer Handler and Typed Protocol framework to rely less on akka-stream
- Unrelated, the Block Packer was slightly optimized to ignore transactions for which it did not locally possess its necessary ancestors.  This fixes a sporadic error log.
## Approach
- Updated rebroadcasters to return their own Topics instead of accepting as parameters
- Updated rebroadcasters to return Resources which close the Topics on exit
- Updated BlockchainPeerHandler to use FS2 instead of akka-stream
- Updated TypedProtocolSetFactory to (mostly) use FS2 instead of akka-stream
- Also added an FS2 helper which drops elements in the event of a slow downstream consumer
## Testing
- preparePR
- Local testnet simulation
## Tickets
- Relates to #2589

### DIFF
--- a/blockchain/src/main/scala/co/topl/blockchain/MempoolBroadcaster.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/MempoolBroadcaster.scala
@@ -8,20 +8,21 @@ import fs2.concurrent.Topic
 
 object MempoolBroadcaster {
 
-  def make[F[_]: Async](
-    mempool:           MempoolAlgebra[F],
-    txsAdoptionsTopic: Topic[F, TypedIdentifier]
-  ): F[MempoolAlgebra[F]] =
-    Async[F].delay {
-      new MempoolAlgebra[F] {
-        def read(blockId: TypedIdentifier): F[Set[TypedIdentifier]] = mempool.read(blockId)
+  def make[F[_]: Async](mempool: MempoolAlgebra[F]): F[(MempoolAlgebra[F], Topic[F, TypedIdentifier])] =
+    Topic[F, TypedIdentifier]
+      .map { txsAdoptionsTopic =>
+        val interpreter =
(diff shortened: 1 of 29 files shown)


blockchain/src/main/scala/co/topl/blockchain/MempoolBroadcaster.scala
```diff
@@ -8,20 +9,22 @@ import fs2.concurrent.Topic
...
+        val interpreter =
+          new MempoolAlgebra[F] {
+            def read(blockId: TypedIdentifier): F[Set[TypedIdentifier]] = mempool.read(blockId)
 
-        def add(transactionId: TypedIdentifier): F[Unit] =
-          mempool.add(transactionId) >>
-          txsAdoptionsTopic.publish1(transactionId).map(_.leftMap(_ => ())).map(_.merge)
+            def add(transactionId: TypedIdentifier): F[Unit] =
+              mempool.add(transactionId) >>
+              txsAdoptionsTopic.publish1(transactionId).map(_.leftMap(_ => ())).map(_.merge)
```
nandotorterolo, line 22: should we raise the exception in case topic is close, like the local chain broadcaster

nandotorterolo (approved):
Thanks for the fixes

blockchain/src/main/scala/co/topl/blockchain/MempoolBroadcaster.scala
SeanCheatham: :+1: Good idea

PR #2741, opened 2023-01-18
Repopulate Remote Mempool

SeanCheatham:
## Purpose
- A remote peer receives a stream of new mempool transaction IDs.  A remote peer is not notified of _existing_ mempool transactions.  This PR notifies the remote peer of the current mempool before streaming new transaction IDs
## Approach
- Move BlockchainPeerServer implementation to `blockchain` module
  - Add `mempool` a dependency.
  - When requesting the stream of transaction IDs, prepend it with a read from the current mempool.
- Semi-related, also prepend the "block adoption stream" with the _current_ head
- Semi-related, added a `DroppingTopic` helper which wraps an FS2 Topic with an implementation that drops messages if the downstream is slow
## Testing
- New unit tests for BlockchainPeerServer
## Tickets
- #2250


hwbehrens:
This is not necessarily directly related to this PR, but I'd like to better understand the behavior around the mechanism for the block adoption stream. I can think of at least four main scenarios:

1. Node A requests a block adoption stream with Head (depth) 5 from Node B, who has tip (depth) 5. Creates the stream but doesn't need to send any 'catch-up' data.
2. Node A requests Head 5 from Node B, who has tip 10. B sends 5 through 10 to catch A up.
3. Node A requests Head 5 from Node B, who has tip 2. B doesn't send anything until it's caught up (or maybe it asks A for head 2?).
4. Node A requests Head 5 from Node B, who has tip 5′ from a different fork. What does B do?

I assume that it uses isWorseThan to check which one is better, then... goes back to the common ancestor and sends the catch-up blocks from there?

nandotorterolo (approved):

PR #2742, opened 2023-01-18
TSDK-262 | Header Mediator in Genus

ignaciofernandezsoto:
## Purpose
Added functionality to apply after header inserted logic through the mediator. Had to do some refactors to get it working. Tests were added.

## Approach
- Created `GraphHeaderMediator` and its tests
- Created `BlockUtils` utils trait
- `GenusGraphMetadata` object was refactored to use methods from `BlockUtils`. It's now easier to mock and test.
- 

## Testing
Created tests for `GraphHeaderMediator`.

## Tickets
* TSDK-262

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphMediator.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphMediator.scala
@@ -1,0 +1,103 @@
+package co.topl.genusLibrary.interpreter
+
+import cats.data.EitherT
+import cats.effect.kernel.Async
+import cats.implicits._
+import co.topl.genusLibrary.algebras.Mediator
+import co.topl.genusLibrary.failure.{Failure, Failures}
+import co.topl.genusLibrary.model.BlockData
+import co.topl.genusLibrary.orientDb.DBFacade
+import co.topl.genusLibrary.orientDb.GenusGraphMetadata._
+import co.topl.genusLibrary.utils.BlockUtils
+import co.topl.models.{BlockBody, BlockHeader}
+import org.typelevel.log4cats.Logger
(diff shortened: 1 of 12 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphMediator.scala
```diff
@@ -0,0 +1,103 @@
...
+import co.topl.genusLibrary.model.BlockData
+import co.topl.genusLibrary.orientDb.DBFacade
+import co.topl.genusLibrary.orientDb.GenusGraphMetadata._
+import co.topl.genusLibrary.utils.BlockUtils
+import co.topl.models.{BlockBody, BlockHeader}
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import scodec.bits.ByteVector
+
+abstract class GraphMediator[F[_]: Async](
```
mundacho, line 17: I think it would be a good idea to add some documentation to this class. From the name it is not clear what exactly is its goal.

mundacho, line 17: Also, this is an abstract class, but there are no abstract methods. why?

mundacho (approved):

genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphMediator.scala
ignaciofernandezsoto: It's an abstract class because I didn't want to implement all the methods for now. Furthermore, the idea would be to split the Mediator trait into multiple ones. See https://topl.atlassian.net/browse/TSDK-278

DiademShoukralla (approved):

PR #2758, opened 2023-02-01
epochBoundary

nandotorterolo:
## Purpose
Epoch Boundary
## Approach

new clock ops, **epochBoundary** (related to Node Morning check-in discussion)

## Testing

preparePR
## Tickets
No tickets

### DIFF
--- a/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
+++ b/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
@@ -37,6 +37,7 @@ object ClockAlgebra {
       def epochRange(epoch: Epoch): F[EpochBoundary] =
         clock.slotsPerEpoch.map(slotsPerEpoch => (epoch * slotsPerEpoch) to (((epoch + 1) * slotsPerEpoch) - 1))
 
+      def epochBoundary(slot: Slot): F[Epoch] = clock.slotsPerEpoch.map(numberOfSlots => slot % numberOfSlots)
     }
   }
 
--- a/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
+++ b/algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
@@ -37,7 +37,7 @@ object ClockAlgebra {
       def epochRange(epoch: Epoch): F[EpochBoundary] =
         clock.slotsPerEpoch.map(slotsPerEpoch => (epoch * slotsPerEpoch) to (((epoch + 1) * slotsPerEpoch) - 1))
 
-      def epochBoundary(slot: Slot): F[Epoch] = clock.slotsPerEpoch.map(numberOfSlots => slot % numberOfSlots)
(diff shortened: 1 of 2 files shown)


scasplte2 (approved):
🔥

algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
```diff
@@ -37,6 +37,7 @@ object ClockAlgebra {
       def epochRange(epoch: Epoch): F[EpochBoundary] =
         clock.slotsPerEpoch.map(slotsPerEpoch => (epoch * slotsPerEpoch) to (((epoch + 1) * slotsPerEpoch) - 1))
 
+      def epochBoundary(slot: Slot): F[Epoch] = clock.slotsPerEpoch.map(numberOfSlots => slot % numberOfSlots)
```
o-aleksandrov-topl, line 40: what about `def isEpochBoundary(slot: Slot): F[Boolean] = clock.slotsPerEpoch.map(numberOfSlots => slot % numberOfSlots === 0L)` ?

o-aleksandrov-topl (approved):

algebras/src/main/scala/co/topl/algebras/ClockAlgebra.scala
nandotorterolo: could be, but this is a use case which uses predicates are not enough and we need the value.
- Eta calculation. : https://github.com/Topl/Bifrost/blob/tetra/consensus/src/main/scala/co/topl/consensus/interpreters/EtaCalculation.scala#L83
```scala
 private def isWithinTwoThirds(from: SlotData): Boolean =   
   from.slotId.slot % slotsPerEpoch <= twoThirdsLength
```

Which I don't if we can use it or no, because. sometimes we obtain _slotsPerEpoch_ from the clock, but in this case, we obtain from the constructor:
- https://github.com/Topl/Bifrost/blob/tetra/consensus/src/main/scala/co/topl/consensus/interpreters/EtaCalculation.scala#L50

```scala
private class Impl[F[_]: Sync: Parallel](
    fetchSlotData:      TypedIdentifier => F[SlotData],
    clock:              ClockAlgebra[F],
    genesisEta:         Eta,
    slotsPerEpoch:      Long,  <<<<
    blake2b256Resource: UnsafeResource[F, Blake2b256],
    blake2b512Resource: UnsafeResource[F, Blake2b512]
  )(implicit cache:     CaffeineCache[F, Bytes, Eta])
      extends EtaCalculationAlgebra[F] {
```

SeanCheatham: I think we should use @o-aleksandrov-topl 's suggestion, and perhaps call it `isEpochStart`.  Otherwise, I think this method should be renamed to something like `slotsSinceEpochStart` since that's technically what it's doing.

As for the `slotsPerEpoch` in the constructor arg, that's not really necessary to keep if it causes confusion.  The clock could be called directly instead; I only added it as a constructor arg to avoid the frequent lookup.

SeanCheatham (approved):

PR #2763, opened 2023-02-03
Mempool.make Return Resource, Reduce Min Success Tests

SeanCheatham:
## Purpose
- MempoolSpec occasionally fails in GitHub Actions due to a timeout
## Approach
- Update Mempool.make to return a Resource
  - Cancel any background expiration fibers during resource finalization
- Reduce the minimum successful unit test requirement for MempoolSpec (main fix)
## Testing
- `preparePR`
## Tickets
N/A

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
@@ -1,7 +1,9 @@
 package co.topl.ledger.interpreters
 
+import cats.{Applicative, MonadThrow}
+import cats.data.OptionT
 import cats.effect.kernel.Spawn
-import cats.effect.{Async, Fiber, Ref, Sync}
+import cats.effect.{Async, Deferred, Fiber, Ref, Resource, Sync}
 import cats.implicits._
 import co.topl.algebras.ClockAlgebra
 import co.topl.codecs.bytes.tetra.instances._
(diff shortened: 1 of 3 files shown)


o-aleksandrov-topl (approved):

ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
```diff
@@ -91,24 +95,46 @@ object Mempool {
...
+          fetchTransaction(transactionId).flatMap(addTransactionWithDefaultExpiration)
+        )
 
       def remove(transactionId: TypedIdentifier): F[Unit] =
-        state.update(_ - transactionId)
+        whenNotTerminated(
+          state
+            .getAndUpdate(_ - transactionId)
+            .flatTap(entries =>
+              OptionT.fromOption[F](entries.get(transactionId)).foldF(Applicative[F].unit)(_.expirationFiber.cancel)
```
nandotorterolo, line 127: LGTM, 1 question,
- fiber cancelation here, is because the resource release cancel, will be invoked at the end?

nandotorterolo (approved):

ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
SeanCheatham: When removing an entry from the mempool, it's not necessary to keep around the background fiber associated with the entry.  We can eagerly cancel it instead.  The previous behavior was to allow the callback to still take place; it would just be silently ignored.  This change is just a slight optimization to reduce the number of unnecessary background fibers.

PR #2767, opened 2023-02-06
Refine GitHub Workflows

SeanCheatham:
## Purpose
- Removes lingering Dion references in the github workflows
- Removes tetra branch references in the github workflows
## Approach
- Delete maven_release (Bifrost will no longer need to publish any libraries)
- Delete gjallarhorn_doc (No longer exists)
- Publish testnet-orchestrator and network-delayer to GCR
- Update release process to trigger on Tag to main or release- branches.  Workflow will publish docker images and create a GitHub draft release with the JAR file.
## Testing
- N/A (need to wait for this all to be on `main`)
## Tickets
- Relates to BN-804


.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: find JAR
+        run: echo "JAR=$(find ./node/target/scala-2.13 -maxdepth 1 -name "*.jar")" >> $GITHUB_ENV
+
+      - name: generate MD5 checksum
+        working-directory: ./node/target/scala-2.13
+        run: md5sum $(basename ${{ env.JAR }}) > "$(basename ${{ env.JAR }}).md5"
+
+      - name: upload JAR artifact
+        uses: actions/upload-release-asset@v1
```
hwbehrens, line 38: Did you have a specific reason for choosing `upload-release-asset` vs [`action-gh-release`](https://github.com/marketplace/actions/gh-release)? Since we're changing actions anyway, I'm a little nervous about adopting an unmaintained action vs a maintained one.

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: setup-scala
+        uses: olafurpg/setup-scala@v10
+
+      - name: assemble full JAR
+        run: sbt node/assembly
+
+      - name: find JAR
+        run: echo "JAR=$(find ./node/target/scala-2.13 -maxdepth 1 -name "*.jar")" >> $GITHUB_ENV
+
+      - name: generate MD5 checksum
```
hwbehrens, line 33: Might we consider using `sha256sum` here instead? I feel like the added length isn't that much more cumbersome, and the people who are actually checking build hashes might appreciate the stronger guarantee.

.github/workflows/_docker_publish_unofficial.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish Docker Image
+name: Publish Docker Images (Unofficial)
```
hwbehrens, line 1: If a user wants to use a Docker artifact, what does their decision-making process look like for choosing between the Official and Unofficial versions?

.github/workflows/_sbt_build.yml
```diff
@@ -70,11 +70,13 @@ jobs:
...
 
-      - name: Compile and Test
+      - name: Compile
+        run: sbt Test/compile
+
+      - name: Tests and Checks
         run: sbt checkPRTestQuick
 
       - name: Upload Test Results
-        if: always()
```
admiraladmirable, line 77: This if statement is needed to see failed test results, since steps do not run if previous ones fail

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,55 @@
...
+    with:
+      preserve-cache-between-runs: true
+  publish-docker-images-official:
+    uses: ./.github/workflows/_docker_publish_official.yml
+    needs: [sbt-build]
+  build-artifact:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout current branch (full)
+        uses: actions/checkout@v2
```
admiraladmirable, line 19: `actions/checkout@v2` for consistency

.github/workflows/prerelease.yml
SeanCheatham: No specific reason; it's just what was referenced in several (apparently outdated) docs.

Using `action-gh-release` would change up the approach a bit.  The goal of `upload-release-asset` was to allow a human to create a prerelease manually, and the workflow would be triggered and update it with the correct assets.  Using `action-gh-release`, I think we would instead have a human create and push a _tag_, and the workflow would be triggered and create a corresponding _release_ (attaching the assets in the process) (perhaps a draft).  A human could go back afterwards and update with proper release notes and finish the release.

I think I like the `action-gh-release` approach better (plus it's not deprecated)

.github/workflows/prerelease.yml
SeanCheatham: I don't have a particular preference here, just that the Dion artifacts used MD5 checksums

.github/workflows/_docker_publish_unofficial.yml
SeanCheatham: Maybe I should rename to `Public` and `Private` respectively.  The ones published by `Unnoficial` (private) go to our private Google Container Registry (which requires login credentials).  The ones published by `Official` (public) go to DockerHub and GitHub Container Registry (and don't require a login to read/use).

.github/workflows/push.yml
```diff
@@ -0,0 +1,16 @@
+name: Push
```
admiraladmirable, line 1: I am a little worried about removing the validation on the CI/Push action.

Even though it's pretty redundant, there are edge cases where validation can be missed after merging a PR.

.github/workflows/prerelease.yml
SeanCheatham: @scasplte2 Do you have a preference between MD5 and SHA256?

.github/workflows/push.yml
SeanCheatham: Good point - updated PR to do all the normal compile/test steps here

.github/workflows/_sbt_build.yml
SeanCheatham: :+1: Fixed

.github/workflows/prerelease.yml
SeanCheatham: Obsolete now

.github/workflows/prerelease.yml
SeanCheatham: Updated the implementation to use `action-gh-release` instead.  It'll now activate on any _tag_ push to `main` or `release-**`

.github/workflows/_docker_publish_unofficial.yml
SeanCheatham: Renamed to Public/Private

admiraladmirable (approved):

.github/workflows/prerelease.yml
hwbehrens: James would prefer to publish both MD5 and SHA256.

SeanCheatham: :+1: Updated to publish `.md5` and `.sha256` artifacts

hwbehrens (approved):
LGTM! Thanks for the changes.

PR #2775, opened 2023-02-07
BN-830_slotData_protobuf_replaceModel

nandotorterolo:
## Purpose
Replace  2 specific models, **SlotData, SlotId**

## Approach
Replace old Models for protobuf-specs
## Testing
preparePr
## Tickets
closes BN-830

### DIFF
--- a/blockchain/src/test/scala/co/topl/blockchain/interpreters/BlockchainPeerServerSpec.scala
+++ b/blockchain/src/test/scala/co/topl/blockchain/interpreters/BlockchainPeerServerSpec.scala
@@ -9,10 +9,11 @@ import org.scalamock.munit.AsyncMockFactory
 import co.topl.models.ModelGenerators._
 import co.topl.{models => legacyModels}
 import legacyModels._
-import co.topl.consensus.models.BlockHeader
+import legacyModels.utility._
+import co.topl.consensus.models.{BlockHeader, SlotData}
 import co.topl.node.models.BlockBody
 import co.topl.models.generators.node.ModelGenerators.arbitraryNodeBody
-import co.topl.models.generators.consensus.ModelGenerators.arbitraryHeader
+import co.topl.models.generators.consensus.ModelGenerators.{arbitraryHeader, arbitrarySlotData}
 import fs2._
 import fs2.concurrent.Topic
 import org.scalacheck.effect.PropF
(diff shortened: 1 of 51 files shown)


minting/src/test/scala/co/topl/minting/interpreters/OperationalKeyMakerSpec.scala
```diff
@@ -49,7 +51,8 @@ class OperationalKeyMakerSpec
       val vrfProof = mock[VrfCalculatorAlgebra[F]]
       val etaCalculation = mock[EtaCalculationAlgebra[F]]
       val consensusState = mock[ConsensusValidationStateAlgebra[F]]
-      val parentSlotId = SlotId(10L, TypedBytes(1: Byte, Bytes.fill(32)(0: Byte)))
+//      val parentSlotId = SlotId(10L, TypedBytes(1: Byte, Bytes.fill(32)(0: Byte)):BlockId) // remove
```
nandotorterolo, line 54: remove commented line

blockchain/src/test/scala/co/topl/blockchain/interpreters/BlockchainPeerServerSpec.scala
```diff
@@ -132,7 +135,7 @@ class BlockchainPeerServerSpec extends CatsEffectSuite with ScalaCheckEffectSuit
         withMock {
           val currentMempoolSet: Set[TypedIdentifier] = ListSet(mempoolTxA, mempoolTxB, mempoolTxC)
           val mempool = mock[MempoolAlgebra[F]]
-          (mempool.read _).expects(head.slotId.blockId).once().returning(currentMempoolSet.pure[F])
+          (mempool.read _).expects(head.slotId.blockId: TypedIdentifier).once().returning(currentMempoolSet.pure[F])
```
o-aleksandrov-topl, line 138: do we really need all such explicit type `TypedIdentifier`  here and everywhere else?

typeclasses/src/main/scala/co/topl/typeclasses/EqInstances.scala
```diff
@@ -15,6 +18,9 @@ trait EqInstances {
   implicit val bytesEq: Eq[Bytes] =
     (a, b) => a === b
 
+  implicit val bytesStringEq: Eq[ByteString] =
+    (a, b) => ByteString.unsignedLexicographicalComparator().compare(a, b) === 0
+
```
nandotorterolo, line 23: @SeanCheatham what do you think about this Eq implementation?

blockchain/src/test/scala/co/topl/blockchain/interpreters/BlockchainPeerServerSpec.scala
nandotorterolo: I don't like either :( 
We need to find a way to define TypedIdentifier on protobuf, for blockId, HeaderId, TransactionId.
Unfortunately protobuf does not allow to define constants

nandotorterolo: Maybe we can include a new field in blockId, with a new rule:

// Ecample for BlockId
// typePrefix must equal 1exactly
> float typePrefix = 1 [(validate.rules).byte.const = 1];
And then in the node, created the Identity type class for each of those

nandotorterolo: long term: remove TypedIdentifier in node

typeclasses/src/main/scala/co/topl/typeclasses/EqInstances.scala
SeanCheatham: I think you can safely use the normal `ByteString#equals` method (`a == b`).  The ByteString's equals method should compare the actual bytes, unlike an Array[Byte].

ignaciofernandezsoto (approved):

PR #2776, opened 2023-02-07
Fix Release Filters and Steps

SeanCheatham:
## Purpose
- The GitHub `on -> push` filters use OR operations, not AND.  Meaning, the the release.yml was trigger either on push to main/dev OR on a new tag push.  This caused the release to trigger on every push to dev/main.
  - See [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore)
- Luckily, the release.yml also had a broken step, so the releases currently fail
- See [this](https://github.com/Topl/Bifrost/actions/runs/4115872572) workflow run
## Approach
- Only run the release.yml steps when a `v**` tag is pushed
- Separate the MD5_LOCATION and SHA256_LOCATION steps
## Testing
- Needs to be pushed to dev first
## Tickets
- Relates to BN-804

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,10 +2,6 @@ name: Release
 
 on:
   push:
-    branches:
-      - 'main'
-      - 'dev'
-      - 'release-**'
     tags:
       - "v**"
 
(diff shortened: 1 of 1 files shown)


admiraladmirable (approved):
You can validate this by targeting your `BN-804-refine-workflows-2` branch in the on push: section if you want to make sure it's working

SeanCheatham:
@admiraladmirable
> You can validate this by targeting your BN-804-refine-workflows-2 branch in the on push: section if you want to make sure it's working

The `push.yml` workflow seems to behave as expected.  I just wanted to prevent `release.yml` from triggering on every main/dev push, and only have it trigger when there's a new tag.

PR #2778, opened 2023-02-07
Typed Protocol Pipelining (Part 1)

SeanCheatham:
## Purpose
- A TypedProtocol expects the "next message" to come from a particular party (depending on the previous message).  Receiving a message from the wrong party is a protocol violation.
- To enable better throughput, instead of a protocol violation, backpressure can instead be applied.  This would allow for multiple messages to be sent without waiting for a response for the first
## Approach
- Add queues to TypedProtocolInstance's "applier".  Add a background process to handle messages from the queues until failure or completion
## Testing
- preparePR
- Local Testnet Simulation
## Tickets
- BN-796
  - NOTE: There will be follow-up PRs for this ticket.  This PR only enables pipelined messages, but the rest of the code doesn't make use of it yet.


nandotorterolo (approved):
approving in case you need merge it, Networking module is a bit complex for :(

PR #2782, opened 2023-02-08
BN-797-VrfCalculator-Unit-Tests

nandotorterolo:
## Purpose
BN-797-VrfCalculator-Unit-Tests
## Approach

Implement unit test:

**Question:**
- BN-797-VrfCalculator-Unit-Tests

TODO:
- Missing scala doc: VrfCalculatorAlgebra, rhoForSlot, proofForSlot, getHit methods
- VrfCalculator in only used in NodeApp, where leaderElectionValidation and thresholdInterpreter are the same instance: 
    + _ineligibleSlots_ uses leaderElectionValidation
     + _getHit_ uses thresholdInterpreter

Ask: do we use this name to avoid confusion, maybe we can use only leaderElectionValidation?

## Testing
prepare pr

## Tickets
* closes #BN-797

### DIFF
--- a/minting/src/test/scala/co/topl/minting/interpreters/VrfCalculatorSpec.scala
+++ b/minting/src/test/scala/co/topl/minting/interpreters/VrfCalculatorSpec.scala
@@ -1,0 +1,212 @@
+package co.topl.minting.interpreters
+
+import cats.effect.IO
+import cats.effect.IO.asyncForIO
+import cats.implicits._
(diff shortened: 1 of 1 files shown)


SeanCheatham (approved):
I like this as a good starting point since it'll cover probably 95% of what we need.  Eventually if there's time, we should add some additional tests which use generators, just to make sure the proof construction behaves as expected under edge cases.

PR #2785, opened 2023-02-08
Typed Protocol Pipelines (Part 2)

SeanCheatham:
## Purpose
- Adds parallelism to the data requests in BlockchainPeerHandler
- Eagerly notify peers of transactions when adopting a block
- Adjust logging (names and level)
- Refactor BlockchainNetwork and AkkaP2PServer as Resources
## Approach
- In places where data is requested from the BlockchainPeerHandler, determine if the calls could be made in parallel
  - In some cases, data fetching is done in parallel while validation is done sequentially
- Rebroadcast block adoptions into the Transaction Adoptions topic (fetching Body and then Transactions along the way)
- Use `Bifrost.{Service}` naming convention in logging names
- Rework BlockchainNetwork and AkkaP2PServer to lean on FS2 and cats-effect a bit more, allowing for Resource allocation and cleanup
## Testing
- `preparePR`
- Testnet simulation
## Tickets
- BN-796


ignaciofernandezsoto (approved):

blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
```diff
@@ -198,7 +198,7 @@ object Blockchain {
           .compile
           .drain
       )
-      _ <- Resource.eval(p2pFiber.joinWithUnit)
+      _ <- Resource.never[F, Unit]
```
o-aleksandrov-topl, line 201: why we need it here?

o-aleksandrov-topl (approved):

PR #2790, opened 2023-02-08
BN-836_latestProtobufSpecs

nandotorterolo:
## Purpose
Update protobufSpecs with changes:

- https://github.com/Topl/protobuf-specs/pull/34  **BN-836-nodeRules**
- https://github.com/Topl/protobuf-specs/pull/33  **IoTransaction/Brambl PB Validation** 

## TODO
- discuss implementation with 2 failed test case

## Testing
preparePr

## Tickets
* BN-836

### DIFF
--- a/topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
+++ b/topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
@@ -196,13 +196,9 @@ object ToplGrpc {
         extends NodeRpcFs2Grpc[F, Metadata] {
 
       def broadcastTransaction(in: BroadcastTransactionReq, ctx: Metadata): F[BroadcastTransactionRes] =
-        in.transaction
-          .asRight[String]
-          .toEitherT[F]
-          .semiflatMap(interpreter.broadcastTransaction)
-          .leftMap(new IllegalArgumentException(_))
+        interpreter
+          .broadcastTransaction(in.transaction)
           .as(BroadcastTransactionRes())
-          .rethrowT
           .adaptErrorsToGrpc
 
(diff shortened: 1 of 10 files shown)


topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
```diff
@@ -199,7 +197,7 @@ object ToplGrpc {
 
       def broadcastTransaction(in: BroadcastTransactionReq, ctx: Metadata): F[BroadcastTransactionRes] =
         in.transaction
```
SeanCheatham, line 199: I think these implementations can be simplified a bit.  i.e. `interpreter.broadcastTransaction(in.transaction)`

I suspect we may not need `EitherT`s in most places now, unless an isomorphism is needed.

topl-grpc/src/main/scala/co/topl/grpc/package.scala
```diff
@@ -13,6 +14,8 @@ package object grpc {
         case e: StatusException => e
         case i: IllegalArgumentException =>
           Status.INVALID_ARGUMENT.withDescription(i.getMessage).asException()
+        case f: FieldValidationException =>
+          Status.INVALID_ARGUMENT.withDescription(f.getMessage).asException()
```
SeanCheatham, line 18: :+1:

topl-grpc/src/main/scala/co/topl/grpc/ToplGrpc.scala
nandotorterolo: thanks for reviewing it,  in all other cases we need to change the input-output,  to-for TypedIdentfier. but at least we can remove on this method broadcastTransaction is not

SeanCheatham (approved):

ignaciofernandezsoto (approved):

PR #2794, opened 2023-02-09
Timeout Testnet Orchestrator

SeanCheatham:
## Purpose
- Certain testnet conditions may result in a scenario that logically never terminates (i.e. a relay node with no connected peers)
- Semi-related: The K8s API may return an error, but the error message isn't fully logged
## Approach
- Add a timeout to the header adoptions stream, make it configurable
- Add logging to the K8sSimulationController error
## Testing
- Local testnet with a timeout of 5 seconds
  image
## Tickets
- BN-826


admiraladmirable (approved):
🐋

PR #2796, opened 2023-02-10
BN-774-2718-vrf-calculator-cache-eviction

nandotorterolo:
## Purpose
vrf-calculator-cache-eviction

## Approach

caches  (vrfProofsCache,  rhosCache)
entries will expire with this logic 
> operationalPeriodLength * 4

## Testing

The current implementation of VrfCalculator does not allow to test cache. 

Maybe we could change VrfCalculator implementation:

- use cache-store (common-interpreters) to create both caches
- use dataSores (node) to create both caches
- finally try to test cache

## Tickets
https://topl.atlassian.net/browse/BN-774


minting/src/main/scala/co/topl/minting/interpreters/VrfCalculator.scala
```diff
@@ -28,7 +29,8 @@ object VrfCalculator {
     clock:                    ClockAlgebra[F],
     leaderElectionValidation: LeaderElectionValidationAlgebra[F],
     ed25519VRFResource:       UnsafeResource[F, Ed25519VRF],
-    vrfConfig:                VrfConfig
+    vrfConfig:                VrfConfig,
+    vrfCacheTtl:              Long
```
SeanCheatham, line 33: I think instead of a TTL, we should go with a size-based cache.  A TTL is logically more suitable, but it would require proper conversion between `operationalPeriodLength` and a `FiniteDuration` (or Long, in this case).  So instead of providing a TTL for each entry, the cache could be configured with a maximum size upon creation.  The maximum size should be `4 * operationalPeriodLength`, so this parameter would just be renamed to `vrfCacheSize` but still accept the same argument.

For an example of a size-based cache, check out `co.topl.consensus.interpreters.EtaCalculation`.

nandotorterolo: Thanks for reviewing it, and provide an example implementation

SeanCheatham (approved):

PR #2808, opened 2023-02-12
Resdev 172: Create tooling to generate config files for testnet deployment

hwbehrens:
## Purpose

It can be tedious to hand-jam the experimental configurations when running the local testnet simulator. This tool makes that process easier by generating multiple configurations from a higher-level experimental descriptor language.

## Approach

Since most of the interaction will be with the experimental configuration file, a command line tool is sufficient to provide the input and output pathing. I chose Python since the encoding and formatting operations would be simple to implement and performance isn't an issue in this case. I included support for virtual environments to ensure that we don't influence the users' local state.

## Testing

No automated testing has been included.

## Tickets

* Closes RESDEV-172


docs/scripts/experiment-definitions/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+    $ sudo add-apt-repository ppa:deadsnakes/ppa
+    $ sudo apt update
+    $ sudo apt install python3.11 python3.11-venv
+
+Once you have Python 3.11 installed, you can create a virtual environment and install the dependencies:
+
+    $ python3.11 -m venv venv
+    $ source venv/bin/activate
+    $ pip install -r requirements.txt
```
nandotorterolo, line 28: steps work for me: Successfully installed PyYAML-6.0 click-8.1.3 networkx-3.0

nandotorterolo (approved):
Lgtm, but I was no able to run the localTestsimulation, I think my local setup crashed again. I need to fix it.

PR #2820, opened 2023-02-14
BN-838 Genus: new-protobuf-models

nandotorterolo:
## Purpose
Replace Old protobuf model, with new ones. **Genus** module

## Approach

Example: old Address
from
> import co.topl.proto.models.TypedEvidence
To
> import co.topl.brambl.models.Address
## Testing

TODO:  **Transaction**
Should we keep the old Transaction model, close the ticket, and complete the replacement later?

```
import co.topl.consensus.models.BlockHeader << ok
import co.topl.node.models.BlockBody   << ok
import co.topl.proto.models.Transaction   << old model
```

## Tickets
BN-838

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ docker/*.jar
 /node/project/build.properties
 /project/project/metals.sbt
 /project/project/project/metals.sbt
+/venv/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -39,3 +39,4 @@ docker/*.jar
 /node/project/build.properties
 /project/project/metals.sbt
 /project/project/project/metals.sbt
+/venv/
```
nandotorterolo, line 42: @hwbehrens Folder created when I test:  RESDEV-172
did I execute Python env command in the right place? (bifrost root), maybe some > cd biforst/node is missing in docs
thanks

ignaciofernandezsoto (approved):

PR #2822, opened 2023-02-16
BN-815-replace-models-with-protobuf-minting

nandotorterolo:
## Purpose
BN-815-replace-models-with-protobuf-minting
## Testing
preparePr
## Tickets
* closes #BN-815


minting/src/main/scala/co/topl/minting/interpreters/Staking.scala
```diff
@@ -55,27 +62,27 @@ object Staking {
...
             unsignedBlock = unsignedBlockBuilder(partialCertificate)
             messageToSign = unsignedBlock.unsignedHeader.signableBytes
-            signature <- ed25519Resource.use(_.sign(operationalKeyOut.childSK.bytes.data, messageToSign).pure[F])
+            signature <- ed25519Resource.use(_.sign(operationalKeyOut.childSK.value, messageToSign).pure[F])
             operationalCertificate = OperationalCertificate(
               operationalKeyOut.parentVK,
               operationalKeyOut.parentSignature,
               partialCertificate.childVK,
-              Proofs.Knowledge.Ed25519(Sized.strictUnsafe(signature))
+              SignatureEd25519.of(ByteString.copyFrom(signature.toArray))
```
SeanCheatham, line 85: If you import `co.topl.models.utility._`, then I think this could be shortened to `SignatureEd25519(signature)`.  It provides an implicit converter from `ByteVector` to `ByteString` (and vice-versa)

SeanCheatham: This would apply to most of the cases where you call `.toArray`

SeanCheatham (approved):

minting/src/main/scala/co/topl/minting/interpreters/Staking.scala
nandotorterolo: Many thanks for reviewing it, I forget again about the implicit converter.

PR #2827, opened 2023-02-27
Migrate helm charts to their own repo

admiraladmirable:
## Purpose
To follow conventions, we should store our various helm charts in their own repo.

This has a number of benefits, with the main one being hosting the helm repo in GitHub pages, and using GitHub actions to manage releasing.

I've already moved the charts there, so this is just cleaning up the old location and removing the validate helm action from our workflows.

https://github.com/Topl/helm-charts
https://topl.github.io/helm-charts/

## Approach
* Remove helm charts.
* Remove chart validation from workflows.

## Testing
This PR :)

## Tickets
*


hwbehrens (approved):
There was a small concern that changes to how the YAML files are processed might necessitate changes in both repos (and it might make automating testnets more complicated, too), but for now let's split them up and see how it goes. If we need them to be more integrated, we can always pull it in as a remote dependency.

SeanCheatham (approved):
You may need to re-delete the `helm` directory after rebasing `dev`

PR #2830, opened 2023-02-27
use implicit conversion byteString-Bytes

nandotorterolo:
## Purpose
use implicit conversion byteString-Bytes to cleanup code

## Testing
preparePr, Node/integration test
## Tickets
*


hwbehrens (approved):
Looks good, nice cleanup!

PR #2843, opened 2023-03-02
BN-841-remove-legacy-models-from-crypto-sub-modules

nandotorterolo:
## Purpose
remove-legacy-models-from-crypto-sub-modules
## Approach
- crypto module does not depend on any protobuf files
- a new model package was created, with the same names of protobuf model, plus util to help refactor
- new model will use ByteString, instead of Bytes.

## To discuss
- should we move co.topl.models.utility.`KesBinaryTree`, to new crypto.models package?
- should we stop using import co.topl.models.{Bytes, Proofs, SecretKeys, VerificationKeys}, and create it crypto.models package?
- what we want is that crypto module, does not depend on any external model?

## Testing
preparePr, Integration test
## Tickets
*BN-841

### DIFF
--- a/crypto/src/main/scala/co/topl/crypto/models/SignatureKesProduct.scala
+++ b/crypto/src/main/scala/co/topl/crypto/models/SignatureKesProduct.scala
@@ -1,9 +1,7 @@
 package co.topl.crypto.models
 
-import com.google.protobuf.ByteString
-
 case class SignatureKesProduct(
   superSignature: SignatureKesSum,
   subSignature:   SignatureKesSum,
-  subRoot:        ByteString
+  subRoot:        Array[Byte]
 )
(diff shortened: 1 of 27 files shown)


nandotorterolo:
- remove byteStrings -> Array[Byte]
- remove case classes with 1 field
- try to move KesBinaryTree to crypto module

crypto/src/main/scala/co/topl/crypto/models/SignatureKesProduct.scala
```diff
@@ -0,0 +1,7 @@
+package co.topl.crypto.models
+
+case class SignatureKesProduct(
+  superSignature: SignatureKesSum,
+  subSignature:   SignatureKesSum,
+  subRoot:        Array[Byte]
```
SeanCheatham, line 6: As an additional safety measure, would you mind overriding the `equals` and `hashCode` methods for these new case classes (to safely handle Array[Byte] equality)?

SeanCheatham (approved):

PR #2853, opened 2023-03-15
Bn 869 unit tests fs2 actors

o-aleksandrov-topl:
## Purpose
Adding unit tests for fs2-based actors

## Approach
New unit tests are added

## Testing
by `preparePR` command

## Tickets
closes BN-869

### DIFF
--- a/networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
+++ b/networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
@@ -17,7 +17,7 @@ import co.topl.models.generators.node.ModelGenerators
 import co.topl.networking.fsnetwork.BlockCheckerTest.F
 import co.topl.networking.fsnetwork.PeersManager.PeersManagerActor
 import co.topl.networking.fsnetwork.ReputationAggregator.ReputationAggregatorActor
-import co.topl.networking.fsnetwork.TestImplicits._
+import co.topl.networking.fsnetwork.TestHelper._
 import co.topl.node.models.{Block, BlockBody}
 import co.topl.quivr.runtime.DynamicContext
 import munit.{CatsEffectSuite, ScalaCheckEffectSuite}
@@ -40,7 +40,7 @@ class BlockCheckerTest extends CatsEffectSuite with ScalaCheckEffectSuite with A
   test("RemoteSlotData: Request no headers if new slot data is worse") {
     withMock {
       val slotData: NonEmptyChain[SlotData] =
-        arbitrarySlotDataChain.arbitrary.retryUntil(c => c.size < maxChainSize).sample.get
+        arbitraryLinkedSlotDataChain.arbitrary.retryUntil(c => c.size < maxChainSize).sample.get
 
       val reputationAggregator = mock[ReputationAggregatorActor[F]]
 
@@ -51,22 +51,15 @@ class BlockCheckerTest extends CatsEffectSuite with ScalaCheckEffectSuite with A
       (localChain.head _).expects().once().onCall(() => currentSlotDataHead.pure[F])
 
       val slotDataStore = mock[Store[F, BlockId, SlotData]]
-
       val headerStore = mock[Store[F, BlockId, BlockHeader]]
-
       val bodyStore = mock[Store[F, BlockId, BlockBody]]
-
       val headerValidation = mock[BlockHeaderValidationAlgebra[F]]
-
       val headerToBodyValidation = mock[BlockHeaderToBodyValidationAlgebra[F]]
-
       val bodySyntaxValidation = mock[BodySyntaxValidationAlgebra[F]]
-
       val bodySemanticValidation = mock[BodySemanticValidationAlgebra[F]]
-
       val bodyAuthorizationValidation = mock[BodyAuthorizationValidationAlgebra[F]]
-
       val chainSelectionAlgebra = mock[ChainSelectionAlgebra[F, SlotData]]
+
       (chainSelectionAlgebra.compare _).expects(slotData.last, currentSlotDataHead).onCall { case (_, _) =>
         (-1).pure[F]
       }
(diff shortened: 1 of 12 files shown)


networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
```diff
@@ -0,0 +1,948 @@
...
+      val bodyStore = mock[Store[F, BlockId, BlockBody]]
+      val headerValidation = mock[BlockHeaderValidationAlgebra[F]]
+      val headerToBodyValidation = mock[BlockHeaderToBodyValidationAlgebra[F]]
+      val bodySyntaxValidation = mock[BodySyntaxValidationAlgebra[F]]
+      val bodySemanticValidation = mock[BodySemanticValidationAlgebra[F]]
+      val bodyAuthorizationValidation = mock[BodyAuthorizationValidationAlgebra[F]]
+      val chainSelectionAlgebra = mock[ChainSelectionAlgebra[F, SlotData]]
+
+      val slotData: NonEmptyChain[SlotData] =
+        arbitraryLinkedSlotDataChain.arbitrary.retryUntil(c => c.size > 2 && c.size < maxChainSize).sample.get
```
SeanCheatham, line 108: `.sample.get` will only try the generator once, oftentimes resulting in `None` which unexpectedly fails the test.  There's a helper that adds a `.first` method to a Gen if you `import co.topl.models.ModelGenerators.GenHelper`

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
```diff
@@ -0,0 +1,948 @@
...
+      val idAndSlotData: NonEmptyChain[(BlockId, SlotData)] = slotData.map(s => (s.slotId.blockId, s))
+
+      // local is known data which are stored in stores
+      val (localId, localSlotData) = idAndSlotData.head
+      (slotDataStore
+        .getOrRaise(_: BlockId)(_: MonadThrow[F], _: Show[BlockId]))
+        .expects(localId, *, *)
+        .once()
+        .onCall { case (_: BlockId, _: MonadThrow[F], _: Show[BlockId]) => localSlotData.pure[F] }
```
SeanCheatham, line 117: I think these `onCall`s could be shortened to `.returning(localSlotData.pure[F])`.  I generally only use `onCall` if I need to either validate arguments or use arguments when constructing the result.

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
```diff
@@ -0,0 +1,948 @@
...
+              headerValidation,
+              headerToBodyValidation,
+              bodySyntaxValidation,
+              bodySemanticValidation,
+              bodyAuthorizationValidation,
+              chainSelectionAlgebra
+            )
+            .allocated
+        _             <- actor.sendNoWait(BlockChecker.Message.RemoteSlotData(hostId, slotData))
+        shutdownFiber <- actor.gracefulShutdown(shutdown)
```
SeanCheatham, line 86: Could these cleanups be handled automatically by the Resource, instead of using `.allocated` directly?

For example, the test would look like:
```
BlockChecker
    .makeActor(
      reputationAggregator,
      peersManager,
      localChain,
      slotDataStore,
      headerStore,
      bodyStore,
      headerValidation,
      headerToBodyValidation,
      bodySyntaxValidation,
      bodySemanticValidation,
      bodyAuthorizationValidation,
      chainSelectionAlgebra
    )
    .use(actor => actor.sendNoWait(BlockChecker.Message.RemoteSlotData(hostId, slotData))
```

SeanCheatham (approved):

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
o-aleksandrov-topl: thanks! I will change it

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
o-aleksandrov-topl: Yes it could be, but sometimes I want to check actor's state after processing a message, and it is a more natural way to check it in `for` statement

networking/src/test/scala/co/topl/networking/fsnetwork/TestHelper.scala
```diff
@@ -0,0 +1,58 @@
...
+import co.topl.models.generators.consensus.ModelGenerators
+import co.topl.node.models.BlockBody
+import org.scalacheck.{Arbitrary, Gen}
+import org.scalamock.handlers.{CallHandler1, CallHandler2, CallHandler3}
+
+import scala.annotation.tailrec
+
+object TestHelper extends TransactionGenerator {
+
+  implicit class CallHandler1Ops[T1, R](ch: CallHandler1[T1, R]) {
```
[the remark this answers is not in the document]
o-aleksandrov-topl: well, yes. if I try to check the exact number of calling for a function with `onCall` then for some reason the compiler does not recognize it correctly and doesn't compile. But if I use range then it compiles and works

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
o-aleksandrov-topl: I've changed to using Resource it make code cleaner, but according to documentation `allocated` will free the resource in any bad case anyway :) 
```
* If the outer `F` fails or is interrupted, `allocated` guarantees that the finalizers will
   * be called. However, if the outer `F` succeeds, it's up to the user to ensure the returned
   * `F[Unit]` is called once `A` needs to be released. If the returned `F[Unit]` is not called,
   * the finalizers will not be run.
 ```

PR #2854, opened 2023-03-16
Private Testnet Relative Stakes

SeanCheatham:
## Purpose
- Enables launching a Private Testnet using configurable relative stake values
- Updates ChainSelectionTest to use different stake values per staker
## Approach
- Add ApplicationConfig and application.conf settings for private testnet relative-stakes
- Use configuration settings when initializing a private testnet
- Update ChainSelectionTest to use 3/6, 2/6, and 1/6 relative stakes for each staker
- Update ChainSelectionTest to relaunch nodes in the order required by the P2P known peers
- Update ChainSelectionTest timeouts to be shorter (closer to expected durations)
## Testing
- Local test runs
## Tickets
N/A

### DIFF
--- a/node/src/main/resources/application.conf
+++ b/node/src/main/resources/application.conf
@@ -42,10 +42,14 @@ bifrost {
   big-bang {
     // Big-Bang configuration type (i.e. `private`)
     type = "private"
+    // The timestamp to use for the genesis block.  If not provided, will default to the current time (padded with a few seconds)
+    // timestamp = 12345667
     // The number of stakers to create and register in the big bang block
     staker-count = 1
     // The 0-based index of this node out of `staker-count` (i.e. if `10` total stakers, `local-staker-index` may be `3`)
     local-staker-index = 0
+    // The relative stake to assign to each of the stakers.  If not provided, will default to uniform stake distribution.
+    // relative-stakes = ["1/2", "1/4", "1/4"]
   }
   // Settings of the protocol (consensus)
   protocols {
--- a/node/src/main/resources/application.conf
+++ b/node/src/main/resources/application.conf
@@ -42,14 +42,14 @@ bifrost {
   big-bang {
     // Big-Bang configuration type (i.e. `private`)
     type = "private"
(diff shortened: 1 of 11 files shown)


node/src/main/resources/application.conf
```diff
@@ -42,10 +42,14 @@ bifrost {
...
     // Big-Bang configuration type (i.e. `private`)
     type = "private"
+    // The timestamp to use for the genesis block.  If not provided, will default to the current time (padded with a few seconds)
+    // timestamp = 12345667
     // The number of stakers to create and register in the big bang block
     staker-count = 1
     // The 0-based index of this node out of `staker-count` (i.e. if `10` total stakers, `local-staker-index` may be `3`)
     local-staker-index = 0
+    // The relative stake to assign to each of the stakers.  If not provided, will default to uniform stake distribution.
+    // relative-stakes = ["1/2", "1/4", "1/4"]
```
o-aleksandrov-topl, line 52: it looks strange, for all other configs we provide just an index of staker info, but here we provide whole relative stake distribution for all stakers. Maybe just give a relative stake for that particular staker?

o-aleksandrov-topl, line 52: and how about use some virtual amount instead of part? I.e. instead of 1/3 + 1/ 6 + 1/6 + 1/3 use some "virtual" amount. for example we could write "100, 500, 200, 200" and based on the total stake (100 + 500 + 200 + 200) relative stake will be calculated?

o-aleksandrov-topl (approved):

nandotorterolo (approved):

node/src/main/resources/application.conf
SeanCheatham: I think that makes sense.  I'll get started on it

SeanCheatham:
> Nice fire this will be awesome for testing! I wonder how deterministic using the same configuration will be over multiple runs thinking

I can only speak anecdotally, but every successful local run of this test takes exactly 332 or 333 seconds (out of perhaps 10 attempts).  Definitely curious to see if it's consistent in testnet simulations.

PR #2856, opened 2023-03-16
NtpClockSkewer + SchedulerClock

SeanCheatham:
## Purpose
- The local system clock can be incorrectly set by a user.  This may result in blocks with inaccurate/invalid timestamps
## Approach
- Skew the result of SchedulerClock's `currentTimestamp` result by the value of some lookup function
- The lookup function is implemented as a Resource which runs a background stream that polls the NTP Server for the offset (similar to Dion)
## Testing
- Logs indicate successful call 
  image
## Tickets
- #BN-844


common-interpreters/src/main/scala/co/topl/interpreters/NtpClockSkewer.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Constructs a function that returns the current difference between the local clock and the NTP server.
+   * @param ntpServer The server to call for NTP offset information
+   * @param refreshInterval The interval at which to refresh the offset
+   * @param timeout How long to wait for an NTP response
+   */
+  def make[F[_]: Async](
+    ntpServer:       String,
+    refreshInterval: FiniteDuration = 30.seconds,
```
o-aleksandrov-topl, line 24: do we need to do it so often? maybe set up it to big value like 1 hour? I think we will get no meaningful error in 30 seconds. But for example network shutdown for 300 seconds could be a case (in that case we will lost syncing at all)

o-aleksandrov-topl (approved):

common-interpreters/src/main/scala/co/topl/interpreters/NtpClockSkewer.scala
SeanCheatham: Good catch - I removed the default parameter values in the `.make` function.  The actual argument will pull from application.conf where the default is `30 minutes` (same as Dion's default)

PR #2858, opened 2023-03-17
Remove Morphism and Instances

SeanCheatham:
## Purpose
- The "Morphism" approach was a way of adapting between _similar_ types but with the expectation that the transformation may fail
- Our remaining "Morphism" instances are simple adapters between node-crypto and consensus types and are not expected to fail
## Approach
- Remove "Morphism" trait and remaining instances
- Convert the original node-crypto to consensus Isomorphisms into simple `implicit def`s in the `co.topl.consensus.models` package object.
## Testing
- `preparePR`
## Tickets
- Relates to #BN-839


nandotorterolo (approved):
Thanks for this cleanup, and apologies, this changes was introduced by me, maybe lack of experience on how to address the implementation

SeanCheatham:
> Thanks for this cleanup, and apologies, this changes was introduced by me, maybe lack of experience on how to address the implementation

No worries at all @nandotorterolo !  It made sense to use them while we still had other Isomorphism instances.  But now that we no longer need them in the other cases, these crypto-consensus instances can be simplified into basic functions to avoid the `F[_]` overhead.

PR #2859, opened 2023-03-17
BlockProducer Skip Operational Period If No Keys

SeanCheatham:
## Purpose
- A set of keys is held in-memory, and this set changes at the start of each operational period
- If a node restarts in the same operational period in which it previously stopped, it can't recreate the set of keys needed to make blocks for the current operational period
- The block producer should instead skip any slots/eligibilities within the current operational period
## Approach
- Add `slotsPerOperationalPeriod` to Clock
- When BlockProducer is unable to certify a block via the Staker, it will try again starting from the first slot of the next operational period
## Testing
- Byzantine Tests succeed locally, but unsure if this will fix the flakiness on GitHub
  - Ad-hoc test [success](https://github.com/Topl/Bifrost/actions/runs/4450533217)
## Tickets
- #BN-899

### DIFF
--- a/minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
+++ b/minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
@@ -55,52 +55,64 @@ object BlockProducer {
       Slf4jLogger.getLoggerFromName[F]("Bifrost.BlockProducer")
 
     val blocks: F[Stream[F, Block]] =
-      Sync[F].delay(
-        parentHeaders
-          .through(AbandonerPipe(makeChild))
-          .collect { case Some(block) => block }
-      )
+      Sync[F].delay(parentHeaders.through(AbandonerPipe(makeChild)))
 
     /**
      * Construct a new child Block of the given parent
      */
-    private def makeChild(parentSlotData: SlotData): F[Option[Block]] =
+    private def makeChild(parentSlotData: SlotData): F[Block] =
       Async[F].onCancel(
-        for {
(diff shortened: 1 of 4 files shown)


minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
```diff
@@ -55,52 +55,64 @@ object BlockProducer {
...
-          // restarts in the middle of an operational period.  The node must wait until the next operational period
-          // to have a set of corresponding linear keys to work with
-          maybeHeader <- staker.certifyBlock(parentSlotData.slotId, nextHit.slot, blockMaker)
-          body = BlockBody(fullBody.transaction.map(_.id))
-          _ <- OptionT
-            .fromOption[F](maybeHeader)
-            .semiflatTap(block => Logger[F].info(show"Minted header=$block body=$body"))
-            .value
-        } yield maybeHeader.map(Block(_, body)),
+        clock.globalSlot >>= attemptUntilCertified(parentSlotData),
```
scasplte2, line 65: Perhaps a silly question but what is that operator?

scasplte2 (approved):
Nice addition 👍

minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
SeanCheatham: It's a shorthand for [.flatMap](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/syntax/flatMap.scala#L45)

PR #2871, opened 2023-03-23
Revise CONTRIBUTING.md

abebeos:
* remove bold, ALL-CAPS, imperative sentence
  * Far too much emphasis, possibly alienating tenor (sidenote: CLA is missing)
* more: see commits


hwbehrens (approved):
Fair enough! I will look into the CLA situation and see if we can get that clarified.

PR #2873, opened 2023-03-23
Bn 879 block body sync error handling part 1

o-aleksandrov-topl:
## Purpose
Change way how downloading of block bodies is done:
Previous:
 -- Download the body for each downloaded header
Current:
 -- Request downloading of all bodies for the current better tip, to avoid duplication of block download requestsProxy had been created. That proxy sends download requests only for new blocks. If the proxy already contains requested blocks or blocks could be applied to the current chain then the proxy immediately returns block bodies

## Approach
By introducing requests proxy

## Testing
Unit tests + Multinode test
## Tickets
*Bn 879

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
@@ -286,6 +286,7 @@ object BlockChecker {
         newBlockBodies      <- EitherT.fromOptionF(newBodiesOpt, "Skip validation of known bodies")
         verifiedFullBlocks  <- verifyAndSaveBodies(state, newBlockBodies)
         appliedNewHeadBlock <- tryToApplyBestBlock(state, verifiedFullBlocks.last.header.id)
+        _                   <- EitherT.liftF(requestNextBodyBlocks(state))
         newState = updateState(state, appliedNewHeadBlock)
       } yield (newState, appliedNewHeadBlock)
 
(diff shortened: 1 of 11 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
```diff
@@ -257,12 +261,20 @@ object BlockChecker {
...
+    hostId:            HostId,
+    bestKnownHeaderId: BlockId
+  ): F[Unit] = {
+    val getMissedBodiesFun =
+      getFromChainUntil[F, BlockId](state.slotDataStore.getOrRaise, id => id.pure[F], state.bodyStore.contains) _
+
+    for {
+      missedBodies <- getMissedBodiesFun(bestKnownHeaderId)
+      _            <- Logger[F].info(show"Send request to get missed bodies for blockIds: $missedBodies")
+      message = RequestsProxy.Message.DownloadBlocksRequest(hostId, NonEmptyChain.fromSeq(missedBodies).get)
```
SeanCheatham, line 274: May need to handle the case where `missedBodies` is empty (i.e. switching from tine A to tine B, and then back to tine A) using an OptionT

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
o-aleksandrov-topl: Thanks! looks like I've missed that case

PR #2881, opened 2023-03-27
updateProtobufSpecs_ce3282d

nandotorterolo:
## Purpose
updateProtobufSpecs_ce3282d

### DIFF
--- a/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
+++ b/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
@@ -75,10 +75,10 @@ object BlockPacker {
                       queue.take
                         .flatMap { transaction =>
                           // Attempt to stuff that transaction into our current block
-                          val fullBody = current.addTransaction(transaction)
+                          val fullBody = current.addTransactions(transaction)
                           // If it's valid, hooray.  If not, return the previous value
                           val transactionValidationContext =
-                            StaticTransactionValidationContext(parentBlockId, fullBody.transaction, height, slot)
+                            StaticTransactionValidationContext(parentBlockId, fullBody.transactions, height, slot)
                           validateBody(transactionValidationContext).ifF(fullBody, current)
                         }
                         .logDuration("BlockPacker Iteration")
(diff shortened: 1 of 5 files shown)


minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
```diff
@@ -75,10 +75,10 @@ object BlockPacker {
                       queue.take
                         .flatMap { transaction =>
                           // Attempt to stuff that transaction into our current block
-                          val fullBody = current.addTransaction(transaction)
+                          val fullBody = current.addTransactions(transaction)
```
abebeos, line 78: https://github.com/Topl/Bifrost/blob/a8d8db95ca5536d1ad6ebd8d47fdcda097285892/ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala#L46

Use of plural seems invalid

nandotorterolo: Thanks for reviewing it.
We could use  these 2 options, methods are generated by scala protobuf compiler

```scala
> val fullBody = current.addTransactions(transaction)
> val fullBody = current.addTransactions(Seq(transaction))
``` 
Scala protobuf code,
```scala
    def addTransactions(__vs: co.topl.brambl.models.transaction.IoTransaction *): FullBlockBody = addAllTransactions(__vs)
    def addAllTransactions(__vs: Iterable[co.topl.brambl.models.transaction.IoTransaction]): FullBlockBody = copy(transactions = transactions ++ __vs)
```

SeanCheatham: We don't necessarily get direct control over the output of the generated protobuf-specs, so we can't really change the builder-pattern field naming.  With that said, `FullBlockBody#addTransactions` is variadic in its arguments, allowing you to do `fullBody.addTransactions(a, b, c, d)`.  It looks a little bit strange with only one argument, but it makes sense when dealing with multiple arguments.

SeanCheatham (approved):
Thanks @nandotorterolo !  The code looked funny before so I'm glad you fixed this.

PR #2893, opened 2023-03-30
getTransactionById implementation

nandotorterolo:
## Purpose
Genus:  TransactionServiceFs2Grpc, getTransactionById implementation
## Approach

- Rename server to Genus, and split the 2 services in different files
- getTransactionById implementation
- unit test for server call
- unit test for Transaction schema and vertex <-> Model conversion
## Tickets
subtask 1: of BN-958

- [x] def getTransactionById
- [ ] def getTransactionByAddressStream
- [ ] def getTxosByAddress
- [ ] def getTxosByAddressStream
- [ ] def getTxosByAssetLabel
- [ ] def createOnChainTransactionIndex
- [ ] def getExistingTransactionIndexes
- [ ] def getIndexedTransactions
- [ ] def dropIndex


genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphVertexFetcher.scala
```diff
@@ -20,17 +22,25 @@ object GraphVertexFetcher {
     Resource.pure {
       new VertexFetcherAlgebra[F] {
 
+//        override def fetchRId(oId: ORecordId): F[Either[GE, Option[Vertex]]] =
```
SeanCheatham, line 25: Will this be used later, or could it be removed?

SeanCheatham (approved):

genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphVertexFetcher.scala
nandotorterolo: yes, good catch, my mistake

PR #2899, opened 2023-03-31
BlockHeaderValidation Use EligibilityCache

SeanCheatham:
## Purpose
- The Eligibility Cache was implemented in a previous PR, but it is not yet integrated into the validation process
## Approach
- When a block header runs through BlockHeaderValidation, try to add the header's eligibility to the cache.  If unsuccessful, reject the header
## Testing
- New unit test for BlockHeaderValidation
- Also refactored BlockHeaderValidation to use mUnit
## Tickets
- BN-877
  - Another PR will follow which "repopulates" the cache upon node startup


nandotorterolo (approved):
Looks good to me:

I only would add a note at the begging(validate function) saying.
This validation step has a side effect, which is update the state of EligibilityCache
If you add more validations later, remember to remove the item added in the previous step, or run eligibilityVerification at the end
Maybe this prevent in the future if someone want to change the logic to parallel validation, or add post validations
No needs to add the comment in this Pr.

o-aleksandrov-topl (approved):

PR #2901, opened 2023-04-03
Repopulate Eligibility Cache

SeanCheatham:
## Purpose
- When a node launches, its eligibility cache is initially empty which would allow previous eligibilities along the canonical chain to be re-used
## Approach
- Adds a function which repopulates the Eligibility Cache with recent blocks on the canonical chain
## Testing
- New unit test
- preparePR
## Tickets
- #BN-877

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/interpreters/EligibilityCache.scala
+++ b/consensus/src/main/scala/co/topl/consensus/interpreters/EligibilityCache.scala
@@ -1,6 +1,5 @@
 package co.topl.consensus.interpreters
 
-import cats.{Applicative, Monad}
 import cats.effect._
 import cats.effect.implicits._
 import cats.implicits._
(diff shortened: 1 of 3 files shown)


consensus/src/main/scala/co/topl/consensus/interpreters/EligibilityCache.scala
```diff
@@ -30,6 +33,31 @@ object EligibilityCache {
         ref.modify(_.tryInclude(vrfVK, slot))
     }
 
+  /**
+   * When a node is launched, eligibilities from adopted blocks should be added to the cache.
+   *
+   * @param underlying    The underlying cache to populate
+   * @param maximumLength The maximum number of entries in the underlying cache
+   * @param canonicalHead The current head of the chain
```
nandotorterolo, line 41: This parameter confuse me a bit.
 Because the firs time, it will be the head of the chain, but then it will be other blocks. So not sure which is a good name here.
  Maybe scaladoc could say, 
  > @param canonicalHead The current head of the chain, or its parent block on in its recursive calls

nandotorterolo (approved):
LGTM

consensus/src/main/scala/co/topl/consensus/interpreters/EligibilityCache.scala
SeanCheatham: I think it was a fair point.  I updated the implementation to use an fs2 Stream instead of a recursive operation

o-aleksandrov-topl (approved):

PR #2902, opened 2023-04-04
EligibilityCache Allow BlockId Re-Use

SeanCheatham:
## Purpose
- After restarting a node, a new block will result in re-validating the chain (for now, PR to follow to change this a bit)
- Re-validating the chain of headers will attempt to re-deduplicate the eligibilities from those headers
- Because the EligibilityCache is repopulated with headers on launch, the re-validation of them will fail because the eligibility appears to have been used already
## Approach
- When running EligibilityCache.tryInclude, provide the block ID corresponding to the validation.  If it matches the existing entry, return true
## Testing
- Update unit test for EligibilityCache
- Run 2 local nodes.  Stop 2nd node.  Start 2nd node.  Observed it was able to sync without validation failure
## Tickets
- #BN-877

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/algebras/EligibilityCacheAlgebra.scala
+++ b/consensus/src/main/scala/co/topl/consensus/algebras/EligibilityCacheAlgebra.scala
@@ -1,5 +1,6 @@
 package co.topl.consensus.algebras
 
+import co.topl.consensus.models.BlockId
 import co.topl.models.Bytes
 import co.topl.models.Slot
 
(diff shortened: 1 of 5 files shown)


consensus/src/main/scala/co/topl/consensus/algebras/EligibilityCacheAlgebra.scala
```diff
@@ -14,6 +15,6 @@ trait EligibilityCacheAlgebra[F[_]] {
    * @param slot The slot for which the eligibility was generated
    * @return true if the eligibility was inserted, false if it was already in the cache
    */
-  def tryInclude(vrfVK: Bytes, slot: Slot): F[Boolean]
+  def tryInclude(blockId: BlockId, vrfVK: Bytes, slot: Slot): F[Boolean]
```
nandotorterolo, line 18: > @param blockId  the block ID provided corresponding to the validation

nandotorterolo (approved):
LGTM, maybe in next pr you can include docs for new added param

consensus/src/main/scala/co/topl/consensus/algebras/EligibilityCacheAlgebra.scala
SeanCheatham: :+1: Good catch - updated PR

PR #2903, opened 2023-04-04
Delay Block Producer

SeanCheatham:
## Purpose
- If a node launches and it is de-synced from the network, block production will fail
  - It is impossible to compute an eta value for an epoch if there are 0 blocks
## Approach
- OperationalKeyMaker remove "initialization" step and allow it to happen naturally from block production
- BlockProducer filter parent blocks that are too far in the past
## Testing
- Launch Node A.  Wait an epoch.  Launch Node B (same genesis timestamp).  Node B logs a warning and synchronizes with Node A before producing new blocks.
- Also updated ChainSelectionTest.  Only two (disconnected) nodes are launched at first.  After 2 epochs, the 2 nodes are restarted (connected) with a 3rd node also launched.  All 3 nodes achieve consensus.
## Tickets
- BN-854
  - Note: I'll make a follow-up PR with some changes to BlockHeaderValidation caching

### DIFF
--- a/minting/src/main/scala/co/topl/minting/interpreters/OperationalKeyMaker.scala
+++ b/minting/src/main/scala/co/topl/minting/interpreters/OperationalKeyMaker.scala
@@ -50,9 +50,7 @@ object OperationalKeyMaker {
     ed25519Resource:             UnsafeResource[F, Ed25519]
   ): Resource[F, OperationalKeyMakerAlgebra[F]] =
     for {
-      initialSlot              <- clock.globalSlot.map(_.max(0L)).toResource
-      initialOperationalPeriod <- Resource.pure[F, Long](initialSlot / operationalPeriodLength)
-      stateRef                 <- Ref.of((initialOperationalPeriod, none[Map[Long, OperationalKeyOut]])).toResource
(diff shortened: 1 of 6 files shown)


minting/src/main/scala/co/topl/minting/interpreters/OperationalKeyMaker.scala
```diff
@@ -90,15 +77,15 @@ object OperationalKeyMaker {
...
+    stateRef:                    Ref[F, Option[(Long, Map[Long, OperationalKeyOut])]]
   ) extends OperationalKeyMakerAlgebra[F] {
 
     def operationalKeyForSlot(slot: Slot, parentSlotId: SlotId): F[Option[OperationalKeyOut]] = {
       val operationalPeriod = slot / operationalPeriodLength
       MonadCancelThrow[F].uncancelable(_ =>
         stateRef.get.flatMap {
-          case (`operationalPeriod`, keysOpt) =>
-            keysOpt.flatMap(_.get(slot)).pure[F]
+          case Some((`operationalPeriod`, keys)) =>
```
nandotorterolo, line 87: LGTM, but not sure if we have any benefits with this named parameter.

PR #2904, opened 2023-04-04
OrientDb Address schema, Do not merge

nandotorterolo:
## Purpose
Address Schema implementation

## Approach
- each time we process a TxIO, a new address Vertex will be created and liked using edge

PROPERTIES

|#   |NAME     |TYPE   |LINKED-TYPE/CLASS|MANDATORY|READONLY|NOT-NULL|
|--------|--------|--------|--------|--------|--------|--------|
|0   |ledger   |INTEGER|                 |true     |true    |true    |   
|1   |index    |INTEGER|                 |false    |true    |false   |    
|2   |network  |INTEGER|                 |true     |true    |true    | 
|3   |addressId|BINARY |                 |true     |true    |true    | 

header_body_txio_address

## Questions
The genesis block contains (26), contains 1 Txio (42), and this Txio is related to 3 Addresses (LockAddress, or OutputAddress)
- why is connected with the same address twice, is it correct? if yes, should we represent this information, or if an edge exists between (Txio-Address), do not create a new one? 

## Tickets
* this work is required to implement `getTxosByAddress`  which is a Transaction service endpoint Api part-2 BN-958

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/TransactionFetcherAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/TransactionFetcherAlgebra.scala
@@ -1,8 +1,9 @@
 package co.topl.genusLibrary.algebras
 
+import co.topl.brambl.models.Address
 import co.topl.brambl.models.Identifier.IoTransaction32
 import co.topl.brambl.models.transaction.IoTransaction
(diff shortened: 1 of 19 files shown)


nandotorterolo:
Replaced by https://github.com/Topl/Bifrost/pull/2939

PR #2907, opened 2023-04-06
Remove BlockHeaderValidation Cached Recursion

SeanCheatham:
## Purpose
- BlockHeaderValidation.WithCache recursively validates all ancestor headers before validating the current headers
- This is an expensive and unnecessary computation, especially on node restart which would re-validate the entire chain
## Approach
- Remove the recursive call from BlockHeaderValidation.WithCache
- Add a cache size limit
## Testing
- preparePR and NodeAppTest
## Tickets
- #BN-854

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
+++ b/consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
@@ -335,7 +335,11 @@ object BlockHeaderValidation {
     /**
      * Wraps an existing BlockHeaderValidation with a cache.  Valid block IDs are saved in the cache to avoid recomputing
      * when switching branches.
-     * @param underlying The base cache
+     *
+     * Invalid block IDs are not saved, but this is subject to change.  This is to avoid an adversary flooding the
+     * cache with invalid block IDs, but this comes at the risk of the adversary flooding compute resources.
+     *
+     * @param underlying The base header validation implementation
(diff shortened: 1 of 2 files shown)


consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
```diff
@@ -330,31 +332,32 @@ object BlockHeaderValidation {
 
   object WithCache {
 
+    /**
+     * Wraps an existing BlockHeaderValidation with a cache.  Valid block IDs are saved in the cache to avoid recomputing
```
nandotorterolo, line 336: Question: if we switch branches, the validation of a blockHeader could change? why do we only save valid blocks, with Unit result, and not, blockId (false, true)
> maybe we can rename: @param underlying The base Block Header Validation Algebra

nandotorterolo (approved):

o-aleksandrov-topl (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
SeanCheatham: Good catch on the @param scaladoc - updated the PR.  Also updated with a comment on why only valid block IDs are stored.  We could change this behavior if we want to in the future.  Ticket BN-550 would cache invalid block IDs, but we can't do the same for transaction IDs.

PR #2909, opened 2023-04-06
dockerize GenusApp, do not merge

nandotorterolo:
## Purpose
dockerize Genus App
- v1 PR required to create Genus Integration Test
- next step, create a test, which includes these containers
  + Node
  + TransactionGenerator (which is not dockerized)
  + Genus     

## Testing
I was able to test the app using docker-compose from my host 
## Tickets
https://topl.atlassian.net/browse/BN-935

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -84,6 +84,12 @@ lazy val testnetSimulationOrchestratorDockerSettings =
     Docker / packageName := "testnet-simulation-orchestrator"
   )
 
+lazy val genusDockerSettings =
+  dockerSettings ++ Seq(
+    dockerExposedPorts := Seq(9091),
+    Docker / packageName := "genus-server"
+  )
+
 def assemblySettings(main: String) = Seq(
   assembly / mainClass := Some(main),
   assembly / test := {},
@@ -620,7 +626,7 @@ lazy val catsAkka = project
 
 lazy val genusServer = project
   .in(file("genus-server"))
-  .enablePlugins(BuildInfoPlugin)
+  .enablePlugins(BuildInfoPlugin, JavaAppPackaging, DockerPlugin)
   .settings(
     name := "genus-server",
     commonSettings,
@@ -628,7 +634,8 @@ lazy val genusServer = project
     crossScalaVersions := Seq(scala213),
     buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion),
     buildInfoPackage := "co.topl.buildinfo.genusServer",
-    libraryDependencies ++= Dependencies.genusServer
+    libraryDependencies ++= Dependencies.genusServer,
+    genusDockerSettings
   )
   .dependsOn(
     genusLibrary,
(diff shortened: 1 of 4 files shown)


genus-server/src/main/scala/co/topl/genusServer/GenusServerApp.scala
```diff
@@ -39,6 +39,8 @@ object GenusServerApp
 
       rpcInterpreter   <- ToplGrpc.Client.make[F](conf.rpcNodeHost, conf.rpcNodePort, conf.rpcNodeTls)
       nodeBlockFetcher <- NodeBlockFetcher.make(rpcInterpreter)
+      // sleep a little the service, helps on docker compose
+      _ <- Temporal[F].sleep(30.seconds).flatMap(_ => Logger[F].info(s"After sleep")).toResource
```
SeanCheatham, line 43: Is this meant to wait for the `node` to finish its startup process?  If so, the strategy I've used is to use FS2 to continually poll/retry the `blockIdAtHeight(1)` RPC until a successful result is returned.

build.sbt
```diff
@@ -84,6 +84,12 @@ lazy val testnetSimulationOrchestratorDockerSettings =
     Docker / packageName := "testnet-simulation-orchestrator"
   )
 
+lazy val genusDockerSettings =
+  dockerSettings ++ Seq(
+    dockerExposedPorts := Seq(9091),
+    Docker / packageName := "genus-server"
```
SeanCheatham, line 90: You may also need to add `genus-server` to the Docker publish workflows `_docker_publish_private.yml` and `_docker_publish_public.yml`

SeanCheatham (approved):

genus-server/src/main/scala/co/topl/genusServer/GenusServerApp.scala
nandotorterolo: Thanks for reviewing it, and suggestions.

build.sbt
nandotorterolo: Thanks for your suggestion, but I am unsure about it.
Maybe you can help me with this. I would like to create a readme  (workflows folder or docs) with some references.
Example:
- name, action,is used for, is triggered by
- _docker_publish_private, creates a container on GPC,  used by Integrations test, merging on main branch

I am a little confused about which required actions items we need to update when a new container is created on Bifrost ecosystem

nandotorterolo:
draft pr until define how we will containerize genus, 
Sean's suggestion:

```
Following up on the "integrate Genus into Node" discussion: A couple years ago I did an [experiment/proof-of-concept](https://github.com/Topl/Bifrost/tree/hybrid-db/graph-db/src/main/scala/co/topl/storage) which used OrientDB for the Node's internal storage (which eventually transitioned into using both OrientDB and LevelDB depending on the information being stored).  The reason we didn't move forward with it is because the Node didn't technically need the benefits of OrientDB, and we would sacrifice a little bit of performance to include it.  Back then, the goal was to build the most "performant" blockchain, but now the goal is to build the most impactful blockchain.  With some of the ideas involving TAMv2 and the Impact Ledger, I think we may finally have a need at the Node-level to incorporate OrientDB.
More concretely, I wonder if it's worth exploring embedding the Genus functionality directly in the Node, and not treat it as a separate service.  We would store Transaction and Block data in OrientDB, but other types of information like the event-sourced tree implementations (mempool, box state, canonical block heights) would continue to use LevelDB.
```

nandotorterolo:
Closing, the container is not required, Genus will be included into the Node modules

PR #2914, opened 2023-04-13
Density Chain Selection Tie-Breaker

SeanCheatham:
## Purpose
- Two blockchain peers may be very out-of-sync, in which case, selecting the "best" chain involves the "Density" Chain Selection Rule
- With the Density Chain Selection Rule, two branches may have equal density, in which case a tie will never be resolved
## Approach
- Modify Density Chain Selection to tie-break on the rhoTestHash value of the "last" slot
## Testing
- New unit test
## Tickets
- #BN-956

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/interpreters/ChainSelection.scala
+++ b/consensus/src/main/scala/co/topl/consensus/interpreters/ChainSelection.scala
@@ -26,7 +26,7 @@ object ChainSelection {
   import NonEmpty.instances._
   import Prepend.instances._
 
-  private val lengthOrder = Order.by[SlotData, Long](_.height)
+  private val heightOrder = Order.by[SlotData, Long](_.height)
   private val slotOrder = Order.by[SlotData, Slot](-_.slotId.slot)
 
   private def rhoTestHashOrderSlotData(implicit blake2b512: Blake2b512) =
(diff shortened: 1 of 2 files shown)


o-aleksandrov-topl (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/ChainSelection.scala
```diff
@@ -26,21 +26,21 @@ object ChainSelection {
   import NonEmpty.instances._
   import Prepend.instances._
 
+  private val lengthOrder = Order.by[SlotData, Long](_.height)
```
nandotorterolo, line 29: LGTM, maybe length does not apply anymore, and could be called SlotHeightOrder, or HeightOrder

nandotorterolo (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/ChainSelection.scala
nandotorterolo: no needs to rename if you think length is better

SeanCheatham: :+1: Updated

PR #2917, opened 2023-04-18
Background Operational Key Creation

SeanCheatham:
## Purpose
- At the turn of an operational period, several "child" keys must be constructed.  This is an expensive process which may come at the cost of a few slots near operational period boundaries.
## Approach
- OperationalKeyMaker state holds onto a Map of Deferred OperationalKeyOut instances
- The Deferred instances are fulfilled by a background fiber
- Getting the operational key for a slot does no longer semantically blocks until all keys are produced
- Remove VrfCalculator#ineligibleSlots and moved functionality into OperationalKeyMaker on a slot-by-slot basis
## Testing
- Updated unit tests
- NodeAppTest integration test still works
## Tickets
- #BN-964


minting/src/main/scala/co/topl/minting/interpreters/OperationalKeyMaker.scala
```diff
@@ -36,28 +38,31 @@ object OperationalKeyMaker {
    * @param etaCalculation An EtaCalculation interpreter is needed to determine the eta to use when determining VRF ineligibilities.
    * @param consensusState Used for the lookup of relative stake for VRF ineligibilities
    */
-  def make[F[_]: Async: Parallel: Logger](
-    parentSlotId:                SlotId,
+  def make[F[_]: Async: Logger](
```
nandotorterolo, line 41: no need to build again for this detail, in other pr, remove parentSlotId doc

nandotorterolo (approved):
Thanks for improve the code,

minting/src/main/scala/co/topl/minting/interpreters/OperationalKeyMaker.scala
SeanCheatham: :+1: Updated PR

PR #2925, opened 2023-04-22
AddressSchema

nandotorterolo:
## Purpose
Address Schema implementation
## Approach

- each time we process a TxIO, a new address Vertex will be created and liked using edges

## Testing
PROPERTIES

|#   |NAME     |TYPE   |LINKED-TYPE/CLASS|MANDATORY|READONLY|NOT-NULL|
|--------|--------|--------|--------|--------|--------|--------|
|0   |ledger   |INTEGER|                 |true     |true    |true    |   
|2   |network  |INTEGER|                 |true     |true    |true    | 
|3   |addressId|BINARY |                 |true     |true    |true    | 
|4   |addressEncodedId    |STRING|                 |false    |true    |false   |    

image

## Tickets
* this work is required to implement `getTxosByAddress`  which is a Transaction service endpoint Api part-2 BN-958

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/orientDb/instances/SchemaAddress.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/orientDb/instances/SchemaAddress.scala
@@ -1,0 +1,65 @@
+package co.topl.genusLibrary.orientDb.instances
+
+import co.topl.brambl.models.{Address, Identifier}
+import co.topl.genusLibrary.orientDb.schema.OIndexable.Instances._
+import co.topl.genusLibrary.orientDb.schema.OTyped.Instances._
+import co.topl.genusLibrary.orientDb.schema.{GraphDataEncoder, VertexSchema}
+
+object SchemaAddress {
+
+  /**
+   * Address model:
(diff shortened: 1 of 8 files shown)


SeanCheatham (approved):

SeanCheatham (approved):

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/instances/SchemaAddress.scala
```diff
@@ -0,0 +1,66 @@
...
+        .withProperty(
+          Field.AddressId,
+          _.id.toByteArray,
+          mandatory = true,
+          readOnly = true,
+          notNull = true
+        )
+        .withProperty(
+          Field.AddressEncodedId,
+          // TODO replace encoder for BramblSc implementation
```
nandotorterolo, line 52: @mundacho I need you help here

nandotorterolo: I don't know if I should wait for bamblSc release, or maybe you decide not to include it there, and use your source code, and implement an util method on the node

PR #2928, opened 2023-04-24
BN-973, append historical stream to live data

nandotorterolo:
## Purpose
Genus live data
## Approach
append live data to historical stream using syncronizationTraversal Rpc node
## Testing
- preparePr, all blocks are being inserted with no missing parents

## Tickets
BN-973

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
@@ -4,7 +4,7 @@ import cats.data.{Chain, EitherT, OptionT}
 import cats.effect.Resource
 import cats.effect.kernel.Async
 import cats.implicits._
-import co.topl.algebras.ToplRpc
+import co.topl.algebras.{SynchronizationTraversalSteps, ToplRpc}
 import co.topl.brambl.models.Identifier
 import co.topl.brambl.models.transaction._
 import co.topl.consensus.models.BlockId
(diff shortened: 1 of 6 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
```diff
@@ -97,6 +97,15 @@ object NodeBlockFetcher {
             blockHeader <- OptionT(toplRpc.fetchBlockHeader(headBlockId))
           } yield blockHeader.height).value
 
+        def fetchAdoptions(): F[Stream[F, BlockId]] =
+          for {
+            adoptions <- toplRpc
+              .synchronizationTraversal()
+              .map(_.collect { case SynchronizationTraversalSteps.Applied(blockId) =>
```
SeanCheatham, line 104: I think this is fine for now, but at some point, the `GraphBlockInserter` will need to support the ability to _remove_ blocks that have been un-adopted by the node

node/src/main/scala/co/topl/genus/GenusServer.scala
```diff
@@ -58,13 +60,27 @@ object GenusServer {
...
+              nodeBlockFetcher.fetch(startHeight = graphCurrentHeight + 1, endHeight = nodeLatestHeight + 3)
             )
             .evalTap(blockData => Logger[F].info(s"Inserting block data ${blockData.header.id.show}"))
             .evalMap(graphBlockInserter.insert)
+            .rethrow ++
+          fs2.Stream
+            .force[F, BlockId](
+              nodeBlockFetcher.fetchAdoptions()
+            )
+            .evalMap(nodeBlockFetcher.fetch)
```
SeanCheatham, line 78: In the future (not needed in this PR), it might help to add a step here which checks if the "parent" of the adopted block exists in the Genus DB.  If not, it would recursively fetch from the node until it finds a known block.

SeanCheatham (approved):

genus-library/src/main/scala/co/topl/genusLibrary/interpreter/NodeBlockFetcher.scala
nandotorterolo: thanks for the suggestion.

PR #2935, opened 2023-04-28
BN-996 Process get new best chain after error

o-aleksandrov-topl:
## Purpose
Correctly process state where block in current best chain became invalid, but new tip(s) from remote host is not received yet
 
## Approach
 * Extend typed protocol / blockchain client with remoteBestTip request
 * After block invalidation new best tips are requested from all remote hosts
 * Appropriate unit tests had been added as well

## Testing
Unit tests + multinode integration test

## Tickets
*

### DIFF
--- a/blockchain/src/main/scala/co/topl/blockchain/interpreters/BlockchainPeerServer.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/interpreters/BlockchainPeerServer.scala
@@ -88,7 +88,8 @@ object BlockchainPeerServer {
               blockIdOpt <- blockHeights.useStateAt(head.slotId.blockId)(_.apply(height))
             } yield blockIdOpt
 
-          def getCurrentTip: F[Option[SlotData]] = localChain.head.map(Option.apply)
+          def getLocalBlockAtDepth(depth: Long): F[Option[BlockId]] =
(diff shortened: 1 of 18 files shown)


blockchain/src/main/scala/co/topl/blockchain/interpreters/BlockchainPeerServer.scala
```diff
@@ -87,6 +87,8 @@ object BlockchainPeerServer {
               head       <- localChain.head
               blockIdOpt <- blockHeights.useStateAt(head.slotId.blockId)(_.apply(height))
             } yield blockIdOpt
+
+          def getCurrentTip: F[Option[SlotData]] = localChain.head.map(Option.apply)
```
SeanCheatham, line 91: I think it may be better to implement a `blockIdAtDepth` protocol instead of `getCurrentTip`.  It can still solve the same problem (depth=0 is the same as the tip) and keeps it more in line with the implementation of the gRPC layer.  It may also help us later on if we want to more easily detect how far out-of-sync two nodes are.

o-aleksandrov-topl: @SeanCheatham good idea! Changed!

SeanCheatham (approved):

PR #2939, opened 2023-05-04
BN-958-implement-the-genus-block-api-step-3-getTxosByAddress

nandotorterolo:
## Purpose
BN-958-implement-the-genus-block-api-step-3-getTxosByAddress
## Approach
- a new Schema Txo was created. 
  + fields: transactionOutput, state, outputAddress
  + TODO: Pending question for transactionId, and txoIndex
- the block inserter takes an `iotx`, and for each output creates Txo
  + TODO: Pending question for inputs, and changing the state

- Implement  Transaction Service-getTxosByAddress

## Testing
preparePR, some unit test,  local testing

## Tickets
BN-958

image

[BN-958.txt](https://github.com/Topl/Bifrost/files/11397852/BN-958.txt)
lockAddressTxo

### DIFF
--- a/genus-library/Examples/Queries.md
+++ b/genus-library/Examples/Queries.md
@@ -21,6 +21,12 @@ RETURN $pathelements
 Ex4: Relationship between headers, body, transaction and address, for an existing BlockHeader with id 26:0 (id:cluster)
 
 ```roomsql
-MATCH {Class: LockAddress}-hasAddress-{Class: Transaction}-hasTxIO-{Class: BlockHeader}
+MATCH {Class: LockAddress}-hasLockAddress-{Class: Transaction}-hasTxIO-{Class: BlockHeader}
+RETURN $pathelements
+```
+
+Ex5: Relationship between lockAddress and txo
+```roomsql
+MATCH {Class: LockAddress}-hasTxo-{Class: Txo}
 RETURN $pathelements
 ```
--- a/genus-library/src/main/scala/co/topl/genusLibrary/algebras/TransactionFetcherAlgebra.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/algebras/TransactionFetcherAlgebra.scala
@@ -1,8 +1,8 @@
 package co.topl.genusLibrary.algebras
 
-import co.topl.brambl.models.TransactionId
+import co.topl.brambl.models.{LockAddress, TransactionId}
 import co.topl.brambl.models.transaction.IoTransaction
-import co.topl.genus.services.TransactionReceipt
+import co.topl.genus.services.{TransactionReceipt, Txo}
 import co.topl.genusLibrary.model.GE
 
 /**
@@ -26,5 +26,13 @@ trait TransactionFetcherAlgebra[F[_]] {
    * @return Optional Transaction, None if it was not found
    */
   def fetchTransactionReceipt(transactionId: TransactionId): F[Either[GE, Option[TransactionReceipt]]]
+
+  /**
+   * Retrieve TxOs (spent or unspent) that are associated with any of the specified addresses
+   *
+   * @param lockAddress the lock address
+   * @return Optional LockAddress, None if it was not found
+   */
+  def fetchTransactionsByAddress(lockAddress: LockAddress): F[Either[GE, Seq[Txo]]]
 
 }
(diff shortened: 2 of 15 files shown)


genus-library/Examples/Queries.md
```diff
@@ -21,6 +21,12 @@ RETURN $pathelements
 Ex4: Relationship between headers, body, transaction and address, for an existing BlockHeader with id 26:0 (id:cluster)
 
 ```roomsql
-MATCH {Class: LockAddress}-hasAddress-{Class: Transaction}-hasTxIO-{Class: BlockHeader}
+MATCH {Class: LockAddress}-hasLockAddress-{Class: Transaction}-hasTxIO-{Class: BlockHeader}
```
scasplte2, line 24: Just pointing out that there are multiple addresses that can be assigned to a single TxO. This is probably understood but just wanted to point out.

genus-library/src/main/scala/co/topl/genusLibrary/algebras/TransactionFetcherAlgebra.scala
```diff
@@ -27,4 +27,12 @@ trait TransactionFetcherAlgebra[F[_]] {
    */
   def fetchTransactionReceipt(transactionId: TransactionId): F[Either[GE, Option[TransactionReceipt]]]
 
+  /**
+   * Retrieve TxOs (spent or unspent) that are associated with any of the specified addresses
+   *
+   * @param lockAddress the lock address
+   * @return Optional LockAddress, None if it was not found
+   */
+  def fetchTransactionsByAddress(lockAddress: LockAddress, state: TxoState): F[Either[GE, Seq[Txo]]]
```
scasplte2, line 36: Would be better named as `fetchTransactionByLockAddress` to differentiate from `fetchTransactionByTransactionOutputAddress`

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/instances/SchemaTxo.scala
```diff
@@ -0,0 +1,66 @@
...
+import co.topl.genusLibrary.orientDb.schema.OTyped.Instances._
+import co.topl.genusLibrary.orientDb.schema.{GraphDataEncoder, VertexSchema}
+
+object SchemaTxo {
+
+  /**
+   * Txo model:
+   * @see https://github.com/Topl/protobuf-specs/blob/main/proto/genus/genus_models.proto#L28
+   */
+  object Field {
```
scasplte2, line 16: I don't know enough here but would it help to include the LockAddress in this schema? Perhaps not as it is not a unique thing so Locks kind of live "on their own" and are associated with TxO's whereas an OutputAddress is a unique value that is only associated with a single object. Is that uniqueness property important?

node/src/test/scala/co/topl/genus/GrpcTransactionServiceTest.scala
```diff
@@ -89,4 +90,73 @@ class GrpcTransactionServiceSuite extends CatsEffectSuite with ScalaCheckEffectS
 
   }
 
+  test("getTxosByAddress: Exceptions") {
```
scasplte2, line 93: getTxosByLockAddress?

scasplte2 (approved):

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/instances/SchemaTxo.scala
nandotorterolo: Thanks for your comment. 
We can derive some lockAdress fields from the TransactionOutputAddress, but we are missing the LockId,  your point is valid. Let me pair with @mundacho who is testing the endpoint on the Sdk side
Uniqueness property is important on orientdb, if you define an index,  with  **UNIQUE** type which is our use case here.

```java
enum INDEX_TYPE {
    UNIQUE,   <<<
    NOTUNIQUE,   <<
    FULLTEXT(true),
 ...

```

genus-library/Examples/Queries.md
nandotorterolo: Yes, and your point makes me think that probably the endpoint implementation is not safe.
lockAddress_A -> Txo_A
lockAddress_B -> Txo_A

If we queryByLockAdress(`lockAddress_A`), should return or not Txo_A ?
Should the client provide all the locks to return Txo associated? 
>  queryByLockAdress(lockAddress_A, lockAddress_B)
Let me discuss this implementation

nandotorterolo:
image

SeanCheatham (approved):

node/src/test/scala/co/topl/genus/GrpcTransactionServiceTest.scala
nandotorterolo: I will update this change, in a future Pr, with protobuf-specs updated to the latest. Thanks for this suggestion

nandotorterolo:
@SeanCheatham thanks for reviewing it

PR #2940, opened 2023-05-05
BN-1017-P2P-RemoveRedundantBlockId

o-aleksandrov-topl:
## Purpose
Refactor code: do not use BlockId if blockId is available from the block header
## Approach
 * Use single BlockHeader instead (BlockId, BlockHeader) where it is possible
 * Unit tests refactoring/removing warnings
 
## Testing
Unit tests, Multinode integration test
## Tickets
*closes BN-1017

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
@@ -6,6 +6,7 @@ import cats.effect.{Async, Resource}
 import cats.implicits._
 import co.topl.actor.{Actor, Fsm}
 import co.topl.algebras.Store
+import co.topl.codecs.bytes.tetra.instances.blockHeaderAsBlockHeaderOps
 import co.topl.consensus.algebras._
 import co.topl.consensus.models.{BlockHeader, BlockId, SlotData}
 import co.topl.ledger.algebras._
@@ -41,17 +42,17 @@ object BlockChecker {
     /**
      * Check and adopt remote headers, if headers is valid then appropriate bodies will be requested
      * @param source source of headers, used as a hint from which peer bodies shall be requested
-     * @param idsAndHeaders headers to check and adopt
+     * @param headers headers to check and adopt
      */
-    case class RemoteBlockHeaders(source: HostId, idsAndHeaders: NonEmptyChain[(BlockId, BlockHeader)]) extends Message
+    case class RemoteBlockHeaders(source: HostId, headers: NonEmptyChain[BlockHeader]) extends Message
 
     /**
      * check and adopt block bodies, if adopted bodies is better than local chain
      * then remote bodies became new top block
      * @param source source of bodies, could be used as a hint from which peer next bodies shall be requested
-     * @param idsAndBodies bodies to check
+     * @param blocks blocks to check
      */
-    case class RemoteBlockBodies(source: HostId, idsAndBodies: NonEmptyChain[(BlockId, BlockBody)]) extends Message
+    case class RemoteBlockBodies(source: HostId, blocks: NonEmptyChain[Block]) extends Message
 
     /**
      * Invalidate block because block is invalid by some reason, for example no block body available at any peer or
@@ -86,7 +87,7 @@ object BlockChecker {
     Fsm {
       case (state, RemoteSlotData(hostId, slotData))         => processSlotData(state, hostId, slotData)
       case (state, RemoteBlockHeaders(hostId, blockHeaders)) => processRemoteHeaders(state, hostId, blockHeaders)
-      case (state, RemoteBlockBodies(hostId, blockBodies))   => processRemoteBodies(state, hostId, blockBodies)
+      case (state, RemoteBlockBodies(hostId, blocks))        => processRemoteBodies(state, hostId, blocks)
       case (state, InvalidateBlockId(invalidBlockId))        => processInvalidBlockId(state, invalidBlockId)
     }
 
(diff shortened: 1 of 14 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/PeersManager.scala
```diff
@@ -69,16 +69,16 @@ object PeersManager {
 
     /**
      * @param hostId use hostId as hint for now, later we could have some kind of map of available peer -> headers
-     * @param blockHeaders requested headers
+     * @param blockIds requested headers
```
scasplte2, line 72: minor but `blockIds` plural makes me think this is a list of some type but it seems it is singular since it is wrapped in a `NonEmptyChain`. Also should the comment instead be `Id of Block to be requested from Peer`?

scasplte2: hmm... it does make more sense to use the plural in the sense of the case class though so I'm conflicted.

scasplte2 (approved):
coming in quite unfamiliar with the motivation here but overall I agree with using already defined types and avoiding tuples for the most part. They always seem to be convenient but eventually become cumbersome so agreed with the refactor

networking/src/main/scala/co/topl/networking/fsnetwork/PeersManager.scala
o-aleksandrov-topl: @scasplte2 what about "list of block's id of headers to be requested from peer"?

scasplte2: that is more clear to me

o-aleksandrov-topl: @scasplte2 previously we were forced to have separate blockId and blockHeader, because we didn't have a block id field in the block header. Thus we use block id and block header together. Now, with the recently added blockId field to block header, (that value is checked then we got it from a remote peer) using a separate block id became redundant.

scasplte2: Oh I see. Can you point me to where the ID is defined or calculated? I tried looking through GH but was unsuccessful.

It is unclear to me how to have the block id in the block header since it is derived from the serialization of the block.

o-aleksandrov-topl: @scasplte2 https://github.com/Topl/Bifrost/blob/dev/tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraIdentifiableInstances.scala#L39 here. We just have an optional field in the block header which could be calculated and updated.

scasplte2: I see, thank you for the link 🎉

networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
```diff
@@ -276,30 +278,30 @@ object BlockChecker {
...
   ): F[Unit] = {
-    def takeWithKnownHeaders(ids: NonEmptyChain[BlockId]) =
+    def getKnownHeaderPrefix(ids: NonEmptyChain[BlockId]) =
       OptionT(
         fs2.Stream
           .emits(ids.toList)
           .evalMap(id => state.headerStore.get(id).map((id, _)))
           .takeWhile { case (_, headerOpt) => headerOpt.isDefined }
-          .map { case (id, headerOpt) => (id, headerOpt.get) }
+          .map(_._2.get.embedId)
```
SeanCheatham, line 292: We might benefit from a `embedIdIfMissing` method at some point

networking/src/main/scala/co/topl/networking/fsnetwork/BlockChecker.scala
```diff
@@ -312,12 +314,12 @@ object BlockChecker {
   private def processRemoteBodies[F[_]: Async: Logger](
     state:       State[F],
     hostId:      HostId,
-    blockBodies: NonEmptyChain[(BlockId, BlockBody)]
+    blockBodies: NonEmptyChain[Block]
```
SeanCheatham, line 317: `blocks: NonEmptyChain[Block]`

networking/src/test/scala/co/topl/networking/fsnetwork/BlockCheckerTest.scala
```diff
@@ -1146,7 +1155,7 @@ class BlockCheckerTest extends CatsEffectSuite with ScalaCheckEffectSuite with A
         .getOrRaise(_: BlockId)(_: MonadThrow[F], _: Show[BlockId]))
         .expects(*, *, *)
         .anyNumberOfTimes()
-        .onCall { case (id: BlockId, _: MonadThrow[F], _: Show[BlockId]) =>
+        .onCall { case (id: BlockId, _: MonadThrow[F] @unchecked, _: Show[BlockId] @unchecked) =>
```
SeanCheatham, line 1158: It might reduce the complexity of these tests if you use a `co.topl.algebras.testInterpreters.TestStore` (from the `test` directory of the `algebras` SBT module) instead of a mock for `StoreAlgebra`.

o-aleksandrov-topl: yeap, I have some ideas for refactoring those unit tests, so maybe I will change it as well

PR #2942, opened 2023-05-08
Implement MempoolGraph

SeanCheatham:
## Purpose
- Adds a `MempoolGraph` class which manages the dependencies of transactions within the Mempool
- Updates the old Mempool implementation to use a MempoolGraph internally
- Note: The underlying benefits of the graph structure aren't used yet
- Note: I removed the duplicate-spender detection from the mempool.  It introduced too much complexity, had too many gaps, and needs to be redesigned using the graph model anyway
## Approach
- Added MempoolGraph which tracks Transactions, the Transactions that spend them, and the Transactions that try to spend UTxOs that aren't im the mempool graph
## Testing
- Updated MempoolSpec
- Ran testnet simulator with 2 nodes
## Tickets
- #BN-904
  - Note: I would like to do a follow-up PR with unit tests on the MempoolGraph itself

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/MempoolGraph.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/MempoolGraph.scala
@@ -1,0 +1,94 @@
+package co.topl.ledger.interpreters
+
+import cats.data.NonEmptySet
+import cats.implicits._
+import co.topl.brambl.models.TransactionId
+import co.topl.brambl.models.transaction.IoTransaction
+import co.topl.brambl.syntax._
+
+/**
+ * @param transactions a collection of all transactions in the mempool
+ * @param spenders a mapping of UTxOs to the set of STxOs that spend them.  The key is the transaction ID and the value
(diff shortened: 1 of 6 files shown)


ledger/src/main/scala/co/topl/ledger/interpreters/MempoolGraph.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import cats.data.NonEmptySet
+import cats.implicits._
+import co.topl.brambl.models.TransactionId
+import co.topl.brambl.models.transaction.IoTransaction
+import co.topl.brambl.syntax._
+
+/**
+ * @param transactions a collection of all transactions in the mempool
+ * @param spenders a mapping of UTxOs to the set of STxOs that spend them.  The key is the transaction ID and the value
```
nandotorterolo, line 11: question to understand, when you said value, is value in protobuf specs the "quivr.models.Int128 quantity".
Maybe I am confused, if yes dismiss my comment, but if not, there is Value(Registration), which has no quantity.

nandotorterolo (approved):

PR #2947, opened 2023-05-12
add Genus genesis block to sanityCheck

nandotorterolo:
## Purpose
add Genus genesis block to sanityCheck
## Approach
- Create Genus Grpc client, with only 1 methd used by byzantine test, In futures PR, complete the implementation
- refactor on Test config, and not hard code ports or tls communications with Rpcs

## Testing
Byzantine test, preparePr

## Tickets
BN-1032

### DIFF
--- a/algebras/src/main/scala/co/topl/algebras/ToplGenusRpc.scala
+++ b/algebras/src/main/scala/co/topl/algebras/ToplGenusRpc.scala
@@ -1,0 +1,19 @@
+package co.topl.algebras
+
+import co.topl.genus.services.BlockResponse
+
+/**
+ * Topl Genus Rpc
+ * An interaction layer intended for users/clients of a blockchain node.
+ *
+ * @tparam F Effect type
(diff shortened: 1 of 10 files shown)


algebras/src/main/scala/co/topl/algebras/ToplGenusRpc.scala
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @tparam F Effect type
+ * @tparam S Canonical head changes Synchronization Traversal Container, Ex: Stream, Seq
+ */
+
+/**
+ * TODO Genus Rpc is used by GenusGrpc Client Implementation, which is used only right now by Byzantine test
+ * In future PRs, we should complete the Client Implementation, and this trait
+ */
+trait ToplGenusRpc[F[_]] {
```
SeanCheatham, line 17: I think this is fine for now to maintain parity with the NodeRpc algebra/trait, but it may be better in the long-run to directly use the generated gRPC client stubs.

SeanCheatham (approved):

PR #2948, opened 2023-05-12
P2P Network Use FS2

SeanCheatham:
## Purpose
- The P2P Networking layer still uses Akka (TCP Stream/IO) internally
- Mixing Akka with FS2 complicates Resource scoping when working with streams
- Resources are necessary for proper cleanup of networking actors
## Approach
- Replace AkkaP2PServer with FS2P2PServer
- Convert Multiplexer to use FS2
- Convert ConnectionLeader to use FS2
- Update BlockchainPeerHandler to use a Resource
## Testing
- preparePR, nodeAppTest, Ad-Hoc Test
## Tickets
- Part 1 of #BN-952


networking/src/main/scala/co/topl/networking/multiplexer/MessageParserFramer.scala
```diff
@@ -1,82 +1,41 @@
...
-            else {
-              val (b, remaining) = buffer.splitAt(size + 5)
-              val data = b.drop(5)
-              inner(remaining, acc :+ (typeByte, data))
-            }
-          }
-        inner(buffer, Nil)
+  def apply[F[_]](): Pipe[F, Byte, (Byte, Chunk[Byte])] = {
+    def start(s: Stream[F, Byte]): Pull[F, (Byte, Chunk[Byte]), Unit] =
+      s.pull.unconsN(5).flatMap {
```
nandotorterolo, line 14: question, what does it mean 5 here? could we create a val with some documentation, and in case this value is the same on `parseWhole` method use a variable?

genus-library/src/main/scala/co/topl/genusLibrary/orientDb/OrientDBMetadataFactory.scala
```diff
@@ -28,7 +28,12 @@ object OrientDBMetadataFactory {
   ): Resource[F, Unit] =
     for {
       db <- Resource
-        .make(Sync[F].blocking(orientGraphFactory.getDatabase))(db => OrientThread[F].delay(db.close()))
+        .make(Sync[F].blocking(orientGraphFactory.getDatabase))(db =>
+          OrientThread[F].delay {
+            db.activateOnCurrentThread()
```
nandotorterolo, line 33: thanks for this fix

networking/src/main/scala/co/topl/networking/multiplexer/MessageSerializerFramer.scala
```diff
@@ -1,19 +1,22 @@
...
-  def apply(): Flow[(Byte, ByteString), ByteString, NotUsed] =
-    Flow[(Byte, ByteString)].map(functionTupled)
+  def apply[F[_]](): Pipe[F, (Byte, Chunk[Byte]), Chunk[Byte]] =
+    _.map((function _).tupled)
 
-  val functionTupled: Function1[(Byte, ByteString), ByteString] =
-    t => ByteString(t._1) ++ intToBytestring(t._2.length) ++ t._2
+  def function(typeByte: Byte, chunk: Chunk[Byte]): Chunk[Byte] =
+    Chunk.singleton(typeByte) ++
+    Chunk.array(ByteBuffer.allocate(4).putInt(chunk.size).array()) ++
```
nandotorterolo, line 20: when we allocate 4, does it mean that we are trying to a create an array of byte for Int values, could we use the function created on Connection Leader

networking/src/test/scala/co/topl/networking/p2p/ConnectionLeaderFlowSpec.scala
```diff
@@ -1,89 +1,79 @@
...
-        BigInt(
-          new Blake2b256()
-            .hash(intToBytestring(remoteInt).toArray ++ intToBytestring(expectedPublishedInt).toArray)
-        )
-      ) ConnectionLeaders.Local
-      else ConnectionLeaders.Remote
+import java.nio.ByteBuffer
 
-    val subFlowF = mockFunction[ConnectionLeader, Flow[ByteString, ByteString, NotUsed]]
+class ConnectionLeaderFlowSpec extends CatsEffectSuite with AsyncMockFactory {
```
nandotorterolo, line 13: maybe we can remove the word Flow in the name of the Spec

networking/src/test/scala/co/topl/networking/p2p/ConnectionLeaderFlowSpec.scala
```diff
@@ -1,89 +1,79 @@
...
+        .returning(().pure[F])
+
+      for {
+        implicit0(random1: Random[F]) <- Random.scalaUtilRandomSeedInt[F](0)
+        _ <- ConnectionLeader.fromSocket(reader, writer).assertEquals(expectedConnectionLeader)
+      } yield ()
+    }
   }
 
+  private def intToBytestring(i: Int): Chunk[Byte] =
```
nandotorterolo, line 74: maybe we canuse the ConnecteionLeader private function intToBytestring, "ByteString" name confuses me a little because googleProtobuf.ByteString is not involved here

nandotorterolo (approved):
Great work, honestly I don't know how to test all these changes, just left a couple o minor comments.

networking/src/main/scala/co/topl/networking/multiplexer/MessageParserFramer.scala
SeanCheatham: Updated PR with some additional documentation.  The 5 comes from 1 + 4 (1 for the type byte, 4 for the integer length).  I also updated the implementation a bit to separate reading the type prefix, the data size, and the data.  Because of that, it wouldn't really add significant benefit to create separate variables for those numbers.

networking/src/main/scala/co/topl/networking/multiplexer/MessageSerializerFramer.scala
SeanCheatham: Updated PR.  I moved the helper function to a package object so that it's accessible from multiple places in networking.

networking/src/test/scala/co/topl/networking/p2p/ConnectionLeaderFlowSpec.scala
SeanCheatham: Updated to `ConnectionLeaderPipeSpec`

networking/src/test/scala/co/topl/networking/p2p/ConnectionLeaderFlowSpec.scala
SeanCheatham: Renamed to `encodeInt`

PR #2954, opened 2023-05-18
BN-800 StakerInitializers Spec

nandotorterolo:
## Purpose
StakerInitializers
## Approach
add unit test
## Testing
- TODO: more testing is needed, ask suggestions for specifications
## Tickets
* BN-800


nandotorterolo:
> My suggestion would be to explore the limits of different configurations.
> 
> If the three variables are `(Network, time, num stakers)` then I think the options would be things like
> 
> * Network -> Main, Testnet, Private
> * Time -> past, current future
> * Num stakers -> 1, 2, 3, 10
> 
> * Network is an odd one here since I doubt you want to connect the unit tests to an actual public network. You likely just want to check that the network selections work correctly.
> * Time only has three options but for future time I would just ensure that the program can handle idling for a second or two as we definitely need to be able to spin up the nodes and have them communicating before the big bang goes off
> * Num stakers seems like 1, 2, & 3 are most important but if you can handle a case with many then I would go as large as possible. I suggest 10 since it seems reasonably big but Ideally would be something like 100 or 1000 but that seems too large for a unit test.
> 
> Just my two cents though

Thanks for all your suggestions @scasplte2

SeanCheatham (approved):

nandotorterolo:
@SeanCheatham @scasplte2 Thanks for approval/review

PR #2957, opened 2023-05-22
Block Body Score

SeanCheatham:
## Purpose
- The BlockPacker needs a scoring algorithm for candidate blocks
## Approach
- Add a BlockPackerAlgebra
- Added a BlockPacker interpreter which uses a TransactionCostCalculator and TransactionRewardCalculator
## Testing
- New BlockPacker unit tests
## Tickets
- #BN-905

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
@@ -10,7 +10,6 @@ import co.topl.brambl.syntax.ioTransactionAsTransactionSyntaxOps
 import co.topl.brambl.validation.algebras.TransactionCostCalculator
 import co.topl.ledger.algebras.BlockBodyScoreAlgebra
 import co.topl.ledger.algebras.TransactionRewardCalculatorAlgebra
-import co.topl.models.utility.Ratio
 import com.github.benmanes.caffeine.cache.Caffeine
 import scalacache.Entry
 import scalacache.caffeine.CaffeineCache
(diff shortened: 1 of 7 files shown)


ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
```diff
@@ -0,0 +1,85 @@
...
+      /**
+       * Returns the aggregate benefit-to-cost ratio of the given Transactions
+       * @param candidate A collection of Transactions that may be used in a block body
+       * @return a rational, with higher values indicating a better score
+       */
+      def scoreOf(candidate: Seq[IoTransaction]): F[Ratio] =
+        (
+          candidate.foldMapM(benefitOf),
+          // Associate a base cost of 1 to the list of transactions, to avoid division by zero
+          candidate.foldMapM(costOf).map(_ + 1)
```
nandotorterolo, line 71: minor question, 
- why do you choose to avoid division by zero if Ratio handles it using modules?

nandotorterolo (approved):

ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
scasplte2: I think it is reasonable to have a default cost of 1 associated with any transaction, even if the math can be handled correctly but to that logic perhaps such a "base cost" should be accounted for in the cost calculator so that it will always return a positive-definite value. Unless there is reason to have 0 be a valid return of the cost calculator.

scasplte2 (commented):
very cool 🎯

ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
SeanCheatham: I added the `+ 1` change to avoid potential division-by-zero issues, but after thinking about it more, that's not a foolproof approach.  The denominator of the Ratio is the "cost".  It's theoretically possible that a Transaction has a negative cost if the protocol is configured in a way to support it.  In particular, the cost of each Transaction Input is `-1` right now, with the idea being that consuming a UTxO is a _good thing_ because it results in less memory for node operators.  That gets offset by other logic of the cost estimation (like proof verification), but if we adjusted that `-1` value, we could end up with Transactions that have a net-negative "cost".

This discussion makes me wonder if the logic is wrong; perhaps instead of returning a ratio `benefit / cost`, it should just return a BigInt `benefit - cost` with each side normalized.

Tagging @scasplte2 since his comment is about the same thing.

ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
```diff
@@ -0,0 +1,85 @@
...
+      implicit0(transactionCostCache: CaffeineCache[F, TransactionId, BigInt])   <- createCache(cacheSize)
+      implicit0(transactionRewardCache: CaffeineCache[F, TransactionId, BigInt]) <- createCache(cacheSize)
+    } yield new BlockBodyScoreAlgebra[F] {
+
+      /**
+       * Calculates-and-caches the "cost" of the Transaction, using the `transactionCostCalculator`
+       * @param transaction The transaction to consider
+       * @return a BigInt representing the aggregate "cost" of the transaction
+       */
+      private def costOf(transaction: IoTransaction) =
```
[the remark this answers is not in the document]
SeanCheatham: Updated with type annotation

ledger/src/main/scala/co/topl/ledger/interpreters/BlockBodyScore.scala
SeanCheatham: Updated the PR with my suggested change.  We can change this in the future if we'd like once we integrate it.

PR #2959, opened 2023-05-24
Combine Node+Genus gRPC Bindings

SeanCheatham:
## Purpose
- Serves Node + Genus gRPC Services from a single binding/port
## Approach
- Split GenusServer into Genus and Replicator
- GenusGrpc remove `serve` and replace with `services`
- Rename ToplGrpc to NodeGrpc
- Create new ToplGrpc which acts as a multi-server
## Testing
- preparePR, NodeAppTest
## Tickets
- Part 1/2 of #BN-244
  - Note, rather than add command-line arguments for Genus RPC host/port, I opted to remove the Genus binding altogether
  - Part 2 will add actual command-line arguments for Genus enabled, data directory, and password


nandotorterolo (approved):
Looks good! Thanks for improving Grpc implementation.

PR #2960, opened 2023-05-25
Add Genus command-line args for disable, DB dir, and DB password.  Sa…

SeanCheatham:
## Purpose
- Adds new command-line arguments to NodeApp for Genus
  - `disableGenus`
  - `orientDbDir`
  - `orientDbPassword`
- Also remove the password from Args and Config "show" implementations
## Approach
- Added new definitions to `Args`
- Updated `show` for Args and ApplicationConfig
## Testing
- Launched with and without OrientDB using arguments.  Also verified that passwords are not printed.
## Tickets
- #2198

### DIFF
--- a/genus-library/src/main/scala/co/topl/genusLibrary/orientDb/OrientDBFactory.scala
+++ b/genus-library/src/main/scala/co/topl/genusLibrary/orientDb/OrientDBFactory.scala
@@ -38,10 +38,7 @@ object OrientDBFactory {
         else Resource.eval(Files[F].createDirectory(directory)).attempt.map(r => (r.isRight, false))
 
       _ <- Logger[F]
-        .info(s"""
-          |Genus directoryPath: $directoryPath
-          |Is Genus using a previous database: $previousDb
-          |""".stripMargin)
+        .info(s"Genus directoryPath=$directoryPath previousDatabaseExists=$previousDb")
         .toResource
 
       _ <- Either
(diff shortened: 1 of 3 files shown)


genus-library/src/main/scala/co/topl/genusLibrary/orientDb/OrientDBFactory.scala
```diff
@@ -38,10 +38,7 @@ object OrientDBFactory {
         else Resource.eval(Files[F].createDirectory(directory)).attempt.map(r => (r.isRight, false))
 
       _ <- Logger[F]
-        .info(s"""
-          |Genus directoryPath: $directoryPath
-          |Is Genus using a previous database: $previousDb
-          |""".stripMargin)
+        .info(s"Genus directoryPath=$directoryPath previousDatabaseExists=$previousDb")
```
nandotorterolo, line 41: thanks for this fix

node/src/main/scala/co/topl/node/Args.scala
```diff
@@ -78,12 +81,35 @@ object Args {
     testnetStakerIndex: Option[Int] = None
   )
 
-  implicit val showArgs: Show[Args] =
-    Show.fromToString
+  @main @Lenses
+  case class GenusArgs(
+    @arg(
+      doc = "Disables the Genus server and gRPC services"
```
nandotorterolo, line 87: only gRPC related to Genus,  will be disabled right?  I feel like someone could confuse about all gRPC services(node + genus)

nandotorterolo (approved):

node/src/main/scala/co/topl/node/Args.scala
SeanCheatham: :+1: Updated to `Disables the Genus server and Genus gRPC services`

PR #2969, opened 2023-06-05
Graph Block Packer

SeanCheatham:
## Purpose
- Implements a new "Block Packer" which selects Transactions using the MempoolGraph structure
  - Duplicate/double spends are eliminated based on Transaction Score
- Moves the Node IntegrationTest into a separate module `node-it` which is implemented _like_ a unit test so that it's easier to run from an IDE
- Migrate RPC helpers from byzantine-tests to common-interpreters
- Removes BlockBodyScore
- Fixes a few issues with MempoolGraph
- Update NodeAppTest to send more transactions
- Move CryptoResources, DataStores, and Validators to `blockchain` module
- Modify TransactionGenerator to accept a "fee" function
## Approach
- Implement new BlockPacker based on graph structure
  - Iterative instance performs operations in stages
    1. Fetch mempool.  If empty, retry in 5 seconds.
    2.  Perform authorization validation on all transactions in mempool.  Any invalid transactions are removed (as well as their subtrees).  Note: Transactions aren't permanently removed from the mempool; just from the local BlockPacker attempt
    3. Remove any transaction sub-trees that aren't spendable on the canonical chain
    4. De-double-spend the remaining trunk transactions
    5. De-double-spend the rest of the tree (recursively) until no double-spends remain
    6. Walk the remaining graph to form a block body
## Testing
- NodeAppTest now uses the Transaction Generator to create two competing graphs of transactions.  The "good" graph is confirmed while the "bad" graph is checked for non-existence
- Adds some basic BlockPackerSpec tests (but it could probably use more in the future)
## Tickets
- #BN-906


ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
```diff
@@ -56,13 +57,9 @@ object Mempool {
...
-          .modify(_.removeSubtree(transaction))
-          .map(_.map(_.id))
-          .flatTap(removed =>
-            expirationsState
-              .getAndUpdate(_.removedAll(removed))
-              .flatTap(expirations => removed.flatMap(expirations.get).toList.traverse(_.cancel))
-          )
+          .update(_.removeSingle(transaction)) *> expirationsState
+          .getAndUpdate(_.removed(transaction.id))
+          .flatTap(_.get(transaction.id).traverse(_.cancel))
```
nandotorterolo, line 62: I would like to understand why sometimes the method `removeSingle` is called, and sometimes `removeSubtree`
The mempool graph is well documented but I am not familiar with this logic and confuse me a bit.
Great work!

nandotorterolo (approved):

ledger/src/main/scala/co/topl/ledger/interpreters/Mempool.scala
SeanCheatham: `removeSingle` is used if you want to delete just the specified transaction.  `removeSubtree` is used if you want to delete the specified transaction, as well as any transactions that spend the specified transaction (in a recursive process).  In this particular case, `removeSingle` is used primarily because of the `applyBlock` call.  When applying a block, its transactions should be removed from the mempool.  But the transactions that spend those transactions may still be valid for the next block, so we need to retain them in the mempool.

PR #2977, opened 2023-06-13
BN-1003

o-aleksandrov-topl:
Adding network quality actor, fixing actor's resources releasing

## Purpose
For network discovery we need to get remote peer network performance, thus special actor is introduced for such purpose

## Approach
Separate sub-actor for PeerActor for measuring ping message

## Testing
Unit tests, Multi Node test

## Tickets
*closes BN-1046

### DIFF
--- a/networking/src/test/scala/co/topl/networking/fsnetwork/PeerNetworkQualityTest.scala
+++ b/networking/src/test/scala/co/topl/networking/fsnetwork/PeerNetworkQualityTest.scala
@@ -1,11 +1,12 @@
 package co.topl.networking.fsnetwork
 
 import cats.effect.IO
+import cats.effect.kernel.Async
 import cats.implicits._
 import co.topl.brambl.generators.TransactionGenerator
 import co.topl.networking.blockchain.BlockchainPeerClient
 import co.topl.networking.fsnetwork.NetworkQualityError.{IncorrectPongMessage, NoPongMessage}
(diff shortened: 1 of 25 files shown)


networking/src/test/scala/co/topl/networking/fsnetwork/PeerNetworkQualityTest.scala
```diff
@@ -0,0 +1,243 @@
...
+  val hostId: HostId = "127.0.0.1"
+
+  test("Ping shall be started and result shall be sent to reputation aggregator") {
+    withMock {
+      val pingPongInterval = FiniteDuration(100, MILLISECONDS)
+      val pingDelay = 10
+
+      val client = mock[BlockchainPeerClient[F]]
+      (client.getPongMessage _).expects(*).atLeastOnce().onCall { ping: PingMessage =>
+        Thread.sleep(pingDelay)
```
SeanCheatham, line 39: Even though these are just unit tests, I think it would be best to use Async[F].delayBy instead of Thread.sleep

SeanCheatham (approved):

networking/src/test/scala/co/topl/networking/fsnetwork/PeerNetworkQualityTest.scala
o-aleksandrov-topl: @SeanCheatham yeap good idea. Removed it from tests. Thanks

PR #2978, opened 2023-06-14
ConsensusData Track ActiveStaker

SeanCheatham:
## Purpose
- A protobuf-specs update modifies the Topl token to embed an optional StakingRegistration, and removes the Registration token
- This modification prevents cases where stake is assigned to a non-registered address or a registration is associated with 0 stake.
## Approach
- Change ConsensusDataEventSourcedState to only track changes to Topl box values
## Testing
- `preparePR` locally
## Tickets
- #2712 

## Notes
- There is still a remaining edge case where a user may attempt to hold multiple Topl tokens under the same address.  This needs to be prevented in a future ticket at the ledger level since it needs to be validated in a non-lagging manner
- Depends on https://github.com/Topl/protobuf-specs/pull/73 and https://github.com/Topl/BramblSc/pull/68

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/interpreters/ConsensusDataEventSourcedState.scala
+++ b/consensus/src/main/scala/co/topl/consensus/interpreters/ConsensusDataEventSourcedState.scala
@@ -70,20 +70,6 @@ object ConsensusDataEventSourcedState {
         _ <- addedRegistrations.traverseTap(r => state.stakers.put(r.registration.address, r))
       } yield state
 
-    private def activeQuantityOf(values: Seq[Value]) =
-      values
-        .flatMap(_.value.topl)
-        .filter(_.registration.nonEmpty)
-        .map(_.quantity.value.toByteArray)
-        .foldMap(BigInt(_))
-
-    private def inactiveQuantityOf(values: Seq[Value]) =
-      values
-        .flatMap(_.value.topl)
-        .filter(_.registration.isEmpty)
-        .map(_.quantity.value.toByteArray)
-        .foldMap(BigInt(_))
-
     private def removedStakersOf(transaction: IoTransaction): List[ActiveStaker] =
       transaction.inputs
         .flatMap(_.value.value.topl)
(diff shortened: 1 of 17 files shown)


consensus/src/main/scala/co/topl/consensus/interpreters/ConsensusDataEventSourcedState.scala
```diff
@@ -60,59 +54,47 @@ object ConsensusDataEventSourcedState {
...
-          .map(_.stakingAddress)
-          .tupleRight(none[SignatureKesProduct])
-          .toMap
-          .pure[F]
-        registrations = transaction.outputs
-          .flatMap(_.value.value.registration)
-          .map(r => r.stakingAddress -> r.registration.some)
-          .toMap
-      } yield deregistrations ++ registrations
+    private def activeQuantityOf(values: Seq[Value]) =
```
nandotorterolo, line 73: Is it possible to create shared functions for Apply and UnapplyBlock, and reference them?

models/src/test/scala/co/topl/models/generators/consensus/ModelGenerators.scala
```diff
@@ -68,6 +68,14 @@ trait ModelGenerators {
       )
     )
 
+  implicit val arbitraryStakingRegistration: Arbitrary[StakingRegistration] =
```
nandotorterolo, line 71: maybe we can add later an ActiveStaker generator,  don't have to do it in this pr if you don't need it on test

nandotorterolo (approved):
LGTM

o-aleksandrov-topl (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/ConsensusDataEventSourcedState.scala
SeanCheatham: :+1: I was able to move `activeQuantityOf` and `inactiveQuantityOf` to the companion object.  The other two methods need to remain with the Apply/Unapply since there are slight differences between the implementations.

models/src/test/scala/co/topl/models/generators/consensus/ModelGenerators.scala
SeanCheatham: It might come in handy later - I didn't need it for this ticket but if other work needs it in the future it would be helpful to have

PR #2984, opened 2023-06-22
BN -  leveldbjni from other source

nandotorterolo:
## Purpose
test alternative solution if we are not able to reach out this dependency: 
> sbt.librarymanagement.ResolveException: Error downloading org.ethereum:leveldbjni-all:1.18.3
## Approach
- https://repo1.maven.org/maven2/io/github/tronprotocol/leveldbjni-all/1.18.3/

### DIFF
--- a/node-it/src/test/scala/co/topl/node/NodeAppTest.scala
+++ b/node-it/src/test/scala/co/topl/node/NodeAppTest.scala
@@ -205,7 +205,7 @@ class NodeAppTest extends CatsEffectSuite {
           .flatMap(filterTransactions(_)(ids))
           .map(_.isEmpty)
           .assert,
-        1000.milli,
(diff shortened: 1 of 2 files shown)


node-it/src/test/scala/co/topl/node/NodeAppTest.scala
```diff
@@ -205,7 +205,7 @@ class NodeAppTest extends CatsEffectSuite {
           .flatMap(filterTransactions(_)(ids))
           .map(_.isEmpty)
           .assert,
-        1000.milli,
+        1500.milli,
```
nandotorterolo, line 208: this change is related to timeout issues on the test.

SeanCheatham (approved):

PR #2987, opened 2023-06-27
BN-1072 Added reputation performance calculation for P2P network

o-aleksandrov-topl:
## Purpose
Adding simple performance calculation for P2P remote peer for future Network discovery

## Approach
Receiving data from a remote host triggers a performance reputation update

## Testing
Unit tests + multinode test

## Tickets
*closes #BN-1072

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockBodyFetcher.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockBodyFetcher.scala
@@ -36,7 +36,6 @@ object PeerBlockBodyFetcher {
      * @param blockHeaders bodies block header to download
      */
     case class DownloadBlocks(blockHeaders: NonEmptyChain[BlockHeader]) extends Message
-
   }
 
   case class State[F[_]](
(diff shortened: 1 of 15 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockBodyFetcher.scala
```diff
@@ -79,22 +78,25 @@ object PeerBlockBodyFetcher {
...
-      } yield body
+        _                 <- Logger[F].info(show"Fetching remote body id=$blockId")
+        bodyStart         <- System.currentTimeMillis().pure[F]
+        body              <- downloadBlockBody(state, blockId)
+        bodyEnd           <- System.currentTimeMillis().pure[F]
+        _                 <- Logger[F].info(show"Fetched remote body id=$blockId")
+        _                 <- checkBody(state, Block(blockHeader, body))
+        txAndDownloadTime <- downloadingMissingTransactions(state, body)
+        allTxDownloadTime = txAndDownloadTime.collect { case (_, Some(time)) => time }
+      } yield UnverifiedBlockBody(state.hostId, body, bodyEnd - bodyStart, allTxDownloadTime)
```
SeanCheatham, line 94: Would it be beneficial to make the `delay` proportional to the size of the header/body/transactions?  Otherwise, bigger blocks impose a larger penalty on reputation scores since they naturally take longer to transfer.

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockBodyFetcher.scala
o-aleksandrov-topl: I've created an additional ticket for it BN-1084 . Additional research/check could be required here.

PR #2989, opened 2023-06-29
BN-1075 Switch to actor based P2P system by default

o-aleksandrov-topl:
## Purpose
By default actor based system will be used for P2P interaction

## Approach
Change config file

## Testing
Unit test + Integration tests

## Tickets
*closdes #BN-1075


SeanCheatham (approved):
:partying_face:

PR #2991, opened 2023-06-30
 BN-719-2641-recreate-transaction-test

nandotorterolo:
## Purpose
BN-719-2641-recreate-transaction-test

## Approach
- [x] A Byzantine Test exists which submits many transactions to a group of running nodes and verifies the transactions are confirmed (In progress, only tested using 1 node)
- [ ] - Staking/Registration Transactions (and launch an additional node after 2 epochs which uses the registration) 0%
- [x] - Various Quivr transactions    60 %. Status Pending (threshold)   Not implemented (exact, less, greater, equal)

## TODO (BN-1128) in other Pr, Question  Staking/Registration Transactions:
 -  If a running node, processes a valid transaction from the genesis iotx, which is the configuration that the 2nd node should have to recognize the node 1 registration? (For example, know peers is required?)

## Testing
```
- co.topl.byzantine.TransactionTest:
  + Submits many transaction to one node and verifies the transaction is confirmed 59.701s
```

## Tickets
https://topl.atlassian.net/browse/BN-719


SeanCheatham (approved):
I know there's still more you'd like to cover, but this looks great so far!

nandotorterolo:
> I know there's still more you'd like to cover, but this looks great so far!
Thanks for reviewing Sean.  I won't proceed to merge until test the registration process, but I would like to understand which is the flow that we expect to create the right assertions.

PR #2992, opened 2023-07-06
Private StakerInitializer LockAddress.  Genesis Eta Timestamp

SeanCheatham:
## Purpose
- The private testnet Staker UTxOs use Ed25519 Locks which aren't supported at this time
- To simplify, the 
## Approach
- Staker UTxOs use the same HeightLock as the rest of the genesis transactions
- Add genesis timestamp into genesis Eta calculation
## Testing
- preparePR
## Tickets
- Part 1 of #BN-908


nandotorterolo (approved):
Thanks for this change!

PR #2993, opened 2023-07-07
BN-1124-block-packer-subgraph-score-throws-exception

nandotorterolo:
## Purpose
Block Packer subgraphScore throws exception

## Tickets
 closes: https://topl.atlassian.net/browse/BN-1124

### DIFF
--- a/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
+++ b/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
@@ -285,7 +285,11 @@ object BlockPacker {
                 .toList
                 .map(_._2)
                 .foldMapM(spenders =>
-                  spenders.toList.map(_._1).map(graph.transactions).traverse(subgraphScore(graph)).map(_.max)
+                  spenders.toList
+                    .map(_._1)
+                    .map(graph.transactions)
+                    .traverse(subgraphScore(graph))
+                    .map(_.appended(BigInt(0)).max)
                 )
             ).mapN(_ + _)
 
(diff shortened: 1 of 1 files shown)


minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
```diff
@@ -285,7 +285,11 @@ object BlockPacker {
                 .toList
                 .map(_._2)
                 .foldMapM(spenders =>
-                  spenders.toList.map(_._1).map(graph.transactions).traverse(subgraphScore(graph)).map(_.max)
+                  spenders.toList
```
SeanCheatham, line 288: Good catch!  I think you may be able to simplify this with an earlier match statement.  For example:

```
.foldMapM {
  case spenders if spenders.isEmpty =>
    BigInt(0).pure[F]
  case spenders =>
    spenders.toList.map(_._1) ...
}
```

SeanCheatham (approved):

minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
```diff
@@ -285,7 +285,11 @@ object BlockPacker {
...
+                    .map(_._1)
+                    .map(graph.transactions)
+                    .traverse(subgraphScore(graph))
+                    .map(_.appended(BigInt(0)).max)
```
o-aleksandrov-topl, line 292: maybe better use `maxOption` and then `getOrElse(0)`?

o-aleksandrov-topl (approved):

minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
nandotorterolo: thanks for the suggestion!

minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
nandotorterolo: thanks for the suggestion, going with Sean's review if you don't mind

PR #2995, opened 2023-07-10
Bifrost CLI/Shell

SeanCheatham:
## Purpose
- To participate in staking & block production, a user must generate several secret keys and a Transaction which gets included in the chain
## Approach
- Create a CLI/shell that guides a user through the key generation and transaction creation process
## Testing
- image
## Tickets
- #BN-908
## Usage
- `sbt "node/run cli"`
## Notes
- Follow-up work is still needed to load these keys + transaction from disk for _public_ networks

### DIFF
--- a/node/src/main/scala/co/topl/node/cli/CliApp.scala
+++ b/node/src/main/scala/co/topl/node/cli/CliApp.scala
@@ -1,0 +1,251 @@
+package co.topl.node.cli
+
+import cats.MonadThrow
+import cats.data.EitherT
+import cats.effect.{IO, Resource, Sync}
+import cats.implicits._
+import co.topl.blockchain.StakerInitializers
+import co.topl.brambl.models.transaction.IoTransaction
+import co.topl.brambl.models.{Datum, Event, LockAddress}
+import co.topl.config.ApplicationConfig
+import co.topl.crypto.generation.EntropyToSeed
+import co.topl.crypto.generation.mnemonic.Entropy
+import co.topl.crypto.signing.{Ed25519, Ed25519VRF, KesProduct}
+import co.topl.node.Args
+import com.google.protobuf.ByteString
+import fs2.Chunk
+import fs2.io.file.{Files, Path}
+import quivr.models.Int128
+
+class ConfiguredCliApp(args: Args, appConfig: ApplicationConfig) {
+  import cats.effect.std.Console
+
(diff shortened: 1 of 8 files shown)


node/src/main/scala/co/topl/node/cli/CliApp.scala
```diff
@@ -0,0 +1,291 @@
...
+  /**
+   * Generate and save a VRF key.
+   */
+  private val createVrfKey =
+    StageResultT.liftF[F, Unit](
+      c.println("Generating a VRF Key.  This key establishes your \"randomness\" within the blockchain.")
+    ) >>
+    createSeed
+      .map(Ed25519VRF.precomputed().deriveKeyPairFromSeed)
+      .flatTap(key => write(key._1)("VRF SK", "operator-key.ed25519.sk"))
```
nandotorterolo, line 200: question, at the end of the process, only 1file should be created?
example the operator key will be overridden by the vrf key?

node/src/main/scala/co/topl/node/cli/CliApp.scala
```diff
@@ -0,0 +1,291 @@
...
+      Sync[F]
+        .delay(EntropyToSeed.instances.pbkdf2Sha512(32).toSeed(Entropy.generate(), password = None))
+    )
+
+  /**
+   * Generate and save an Ed25519 Operator key.
+   */
+  private val createOperatorKey =
+    StageResultT.liftF[F, Unit](
+      c.println("Generating an Ed25519 Operator SK.  This key determines your StakingAddress.")
```
nandotorterolo, line 185: Does this require support on Ed25519 on brambl side?

nandotorterolo (approved):
Looks good, great PR, just a couple of questions.

node/src/main/scala/co/topl/node/cli/CliApp.scala
SeanCheatham: Good catch, thanks!  This should have been a separate file.  I updated the PR.

node/src/main/scala/co/topl/node/cli/CliApp.scala
SeanCheatham: The "Operator Key" isn't used for Locks on UTxOs; at this time it is only used to prove/demonstrate that you own a particular StakingAddress

PR #2997, opened 2023-07-11
BN-1101 Send information about block novelty to reputation aggregator

o-aleksandrov-topl:
Send information about block novelty to the reputation aggregator
Slot data is no longer requested and saved if remote slot data can't be new best tip after the sanity check

## Purpose
For novelty reputation information about newly received remote slot data shall be sent to the reputation aggregator

## Approach
No longer decline any already known slot data, accept it and send information about it if block source could be useful for us

## Testing
Unit test + Multinode integration test

## Tickets
closes #BN-1101

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/RequestsProxy.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/RequestsProxy.scala
@@ -6,7 +6,7 @@ import cats.implicits._
 import co.topl.actor.{Actor, Fsm}
 import co.topl.algebras.Store
 import co.topl.codecs.bytes.tetra.instances.blockHeaderAsBlockHeaderOps
-import co.topl.consensus.models.{BlockHeader, BlockId}
+import co.topl.consensus.models.{BlockHeader, BlockId, SlotData}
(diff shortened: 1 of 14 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/RequestsProxy.scala
```diff
@@ -25,6 +25,12 @@ object RequestsProxy {
 
     case object GetCurrentTips extends Message
 
+    case class RemoteSlotData(hostId: HostId, slotData: NonEmptyChain[SlotData]) extends Message
+
+    case class BlocksSource(sources: NonEmptyChain[(HostId, BlockId)]) extends Message
+
+    case class BadKLoopbackSlotData(hostId: HostId) extends Message
```
SeanCheatham, line 32: Is this meant to be "Lookback" instead of "Loopback"?

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/fsnetwork/RequestsProxy.scala
o-aleksandrov-topl: LOL, yes it shall :)

PR #2998, opened 2023-07-12
Tip NN - do not merge this PR

nandotorterolo:
## Purpose
alternative repo to discuss Tip 3, which should live on Tip repo (I don't have access)

for reviewers, Tip 3 defines: protocol version and update proposal, maybe they should be 2 Tips?

I would like to take this PR to discuss items on TODOs list and create an implementation plan.


nandotorterolo:
it will be moved to Tip repo eventually

PR #3001, opened 2023-07-20
BN-1131 Add notifier actor

o-aleksandrov-topl:
## Purpose
Have all P2P time-related actions in one place

## Approach
Creating separate Notification actor for sending notifications about actions to be done

## Testing
Unit test + Integration tests

## Tickets
closes #BN-1131

### DIFF
--- a/config/src/main/scala/co/topl/config/ApplicationConfig.scala
+++ b/config/src/main/scala/co/topl/config/ApplicationConfig.scala
@@ -58,7 +58,7 @@ object ApplicationConfig {
       minimumBlockProvidingReputationPeers: Int = 2,
       minimumRequiredReputation:            Double = 0.66,
       minimumHotConnections:                Int = 3,
-      warmHostsUpdateEveryNSlots:           Double = 4.0
+      warmHostsUpdateEveryNBlock:           Double = 4.0
     )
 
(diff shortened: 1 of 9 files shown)


config/src/main/scala/co/topl/config/ApplicationConfig.scala
```diff
@@ -57,7 +57,8 @@ object ApplicationConfig {
       remotePeerNoveltyInExpectedBlocks:    Double = 2.0,
       minimumBlockProvidingReputationPeers: Int = 2,
       minimumRequiredReputation:            Double = 0.66,
-      minimumHotConnections:                Int = 3
+      minimumHotConnections:                Int = 3,
+      warmHostsUpdateEveryNSlots:           Double = 4.0
```
nandotorterolo, line 61: question: Nslot should be a Integer?

nandotorterolo (approved):

config/src/main/scala/co/topl/config/ApplicationConfig.scala
o-aleksandrov-topl: wrong naming, changed, But it shall be double yes because we could select non integer value here because end unit will be milliseconds

PR #3002, opened 2023-07-24
Bn 1126 block version

nandotorterolo:
## Purpose
- Block Version implementation based on Tip3-draft spec.

## Approach
- node application will run with `app-version`
- Each protocol setting config contains a `min-app-version`
- there are no logic changes 

- depends on:  https://github.com/Topl/protobuf-specs/pull/77

## example 

```
// upgrade example, todo remove this config
//   a staker wants to introduce a change in the protocol, it will send a tx with this information
//   a local storage will collect this configuration file, and also the incoming transaction(the porposal updated)  from other peers
    100 {
      // if App version is 2.0.0, this config will be not considered as compatible
      // new field ideas:
      // status = enabled, disabled, invalid, quorum incomplete,
      // voting = slotProccessBegin = < 100 , slotProccessEnd < 100 - 2
      // votingQuorum = 75%, 90% of total staker?
      // votes = [10 yes, 5 No, 3 abstention]
      min-app-version = "2.1.0"
      f-effective = "15/100"
      vrf-ldd-cutoff = 50
      vrf-precision = 40
      vrf-baseline-difficulty = "1/20"
      vrf-amplitude = "1/2"
      chain-selection-k-lookback = 50
      slot-duration = 2000 milli // this will be the change
      forward-biased-slot-window = 50
      operational-periods-per-epoch = 2
      kes-key-hours = 9
      kes-key-minutes = 9
    }
```

## Testing
- preparePr, node run

## Tickets
* https://topl.atlassian.net/browse/BN-1126

### DIFF
--- a/node/src/main/scala/co/topl/node/ProtocolVersioner.scala
+++ b/node/src/main/scala/co/topl/node/ProtocolVersioner.scala
@@ -1,0 +1,94 @@
+package co.topl.node
+
+import cats.{Eq, Order}
+import scala.collection.SortedSet
+import cats.syntax.all._
+import co.topl.config.ApplicationConfig.Bifrost
+import co.topl.consensus.models.ProtocolVersion
+import co.topl.models.Slot
+import scala.util.Try
+
+/**
+ * This class provides functionality for managing the backwards compatibility with previous protocol definitions
+ * @param appVersion application version
+ * @param protocolVersions all previous rule sets (that must be available in the configuration file)
+ */
+case class ProtocolVersioner(appVersion: Version, protocolVersions: SortedSet[ProtocolSettings]) {
+
+  /** this is the set of protocol settings a particular version of the software can utilize */
+  lazy val compatibleProtocolVersions: SortedSet[ProtocolSettings] =
+    protocolVersions.filter(appVersion >= _.minAppVersion)
+
+  /**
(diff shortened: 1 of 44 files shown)


o-aleksandrov-topl (approved):

node/src/main/scala/co/topl/node/ProtocolVersioner.scala
```diff
@@ -0,0 +1,94 @@
...
+      if (x.firstDigit != y.firstDigit) {
+        x.firstDigit - y.firstDigit
+      } else if (x.secondDigit != y.secondDigit) {
+        x.secondDigit - y.secondDigit
+      } else {
+        x.thirdDigit - y.thirdDigit
+      }
+    }
+
+  def apply(value: String): Version =
```
SeanCheatham, line 69: I think it may be better to return an `Option[Version]` instead of using a default fallback.  (And also rename the method to `parse` instead of `apply`)

SeanCheatham (approved):

node/src/main/scala/co/topl/node/ProtocolVersioner.scala
nandotorterolo: Thanks for review it! Changed to Option, and I will use the first appVersion configured on protocol settings, as the initial app Version.

PR #3014, opened 2023-08-14
Support Loading Staker From Disk

SeanCheatham:
## Purpose
- In real world blockchain networks, the staking keys are owned by a human rather than generated at runtime
- We need the ability to load staking information from disk and use it during the staking process
## Approach
- Create a `StakingInit` helper object which can generate private testnet staking (existing behavior) or load information from disk (new behavior)
## Testing
- Modify MultiNodeTest to register a new staker and launch a 4th node with the new staking information
## Tickets
- #BN-1128
## Note
- I would like to do follow-up work which searches the chain for the staker's registration transaction and delays staker initialization until the registration is active

### DIFF
--- a/node/src/main/scala/co/topl/node/StakingInit.scala
+++ b/node/src/main/scala/co/topl/node/StakingInit.scala
@@ -35,6 +35,21 @@ object StakingInit {
   final val RegistrationTxName = "registration.transaction.pbuf"
 
   /**
+   * Inspects the given stakingDir for the expected keys/files.  If the expected files exist, `true` is returned.
+   */
+  def stakingIsInitialized[F[_]: Async](stakingDir: Path): F[Boolean] =
+    Files
+      .forAsync[F]
+      .list(stakingDir)
+      .compile
+      .toList
+      .map(files =>
+        files.exists(_.endsWith(KesDirectoryName)) &&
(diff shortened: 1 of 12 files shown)


node/src/main/scala/co/topl/node/StakingInit.scala
```diff
@@ -0,0 +1,209 @@
...
+import co.topl.minting.algebras.StakingAlgebra
+import co.topl.minting.interpreters.{OperationalKeyMaker, Staking, VrfCalculator}
+import com.google.protobuf.ByteString
+import fs2.Chunk
+import fs2.io.file.{Files, Path}
+import org.typelevel.log4cats.Logger
+
+object StakingInit {
+
+  final val KesDirectoryName = "kes"
```
o-aleksandrov-topl, line 32: maybe add all filenames to config instead of having it hardcoded?

o-aleksandrov-topl (approved):

node/src/main/scala/co/topl/node/StakingInit.scala
SeanCheatham: I think for now, I'd like to keep these hardcoded.  The directory for staking is configurable, but it may confuse users if we allow them to change the structure of it.  We can always change it later though.

nandotorterolo (approved):

PR #3015, opened 2023-08-15
BN-1071 Open new remote connections to neighbours to hot peers if it is siutable by config

o-aleksandrov-topl:
## Purpose
Dynamically open connection to new remote peers;
resolve peers hostname to IP address to avoid double connection to remote peer;
request and save remote peer server port, for creating correct outbound connection;

## Approach
Add required requests/responses to client, open new connections if it required

## Testing
Unit testing + integration tests

## Tickets
closes #BN-1071


networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
```diff
@@ -21,7 +22,7 @@ import scala.util.Random
 
 package object fsnetwork {
 
-  type HostId = RemoteAddress // IP address? IP address could be changed and bad for identify good peer
+  type HostId = String // IP address? IP address could be changed and bad for identify good peer
```
SeanCheatham, line 25: Could this be a Verification Key of some sort?  The peer's VK could be used as an identifier in the network, and to confirm authenticity, a random message could be exchanged that must be signed by the SK.

blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
```diff
@@ -65,9 +66,10 @@ object Blockchain {
   ): Resource[F, Unit] = {
     implicit val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLoggerFromName[F]("Bifrost.Blockchain")
     for {
-      remotePeers <- Resource.liftK(Queue.unbounded[F, DisconnectedPeer])
-      closedPeers <- Resource.liftK(Queue.unbounded[F, ConnectedPeer])
-      _           <- Resource.liftK(Logger[F].info(s"Received known peers from config: $knownPeers"))
+      remotePeers  <- Resource.liftK(Queue.unbounded[F, DisconnectedPeer])
+      closedPeers  <- Resource.liftK(Queue.unbounded[F, ConnectedPeer])
+      _            <- Resource.liftK(Logger[F].info(s"Received known peers from config: $knownPeers"))
+      currentPeers <- Resource.pure(new CurrentPeersHolder())
```
SeanCheatham, line 72: I think it would be safer to implement this as a `Ref[F, Set[RemoteAddress]]`

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
o-aleksandrov-topl: @SeanCheatham I think about SSL certificate in the future as well

blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
o-aleksandrov-topl: @SeanCheatham fixed! :)

PR #3017, opened 2023-08-15
MultiNodeTest Reduce Nodes to 3

SeanCheatham:
## Purpose
- The MultiNodeTest byzantine test has issues running 4 nodes on GitHub Actions (garbage collection accounts for excessive CPU usage)
## Approach
- Remove one of the nodes in MultiNodeTest
## Testing
- Success: https://github.com/Topl/Bifrost/actions/runs/5872370762/job/15923794654
## Tickets
N/A


nandotorterolo (approved):
Thanks for fixing it

o-aleksandrov-topl:
Is it possible to save initial test? And just disable it? I use that test for checking new nodes discovery and I need at least 3 nodes for it :)

PR #3018, opened 2023-08-16
BN-1143: Improve Testnet Orchestrator Docs

admiraladmirable:
## Purpose
Recent usage of the testnet orchestrator demo has shown the docs to be a bit lacking, This PR aims to improve the documentation. 

## Approach
Update the markdown docs to add:
* More information about the purpose of devcontainers and what they achieve
* Information related to a common issue

## Testing
n/a

## Tickets
*

### DIFF
--- a/docs/DevContainerSimulatorGuide.md
+++ b/docs/DevContainerSimulatorGuide.md
@@ -1,6 +1,25 @@
 # DevContainer Testnet Simulator Guide
 Instructions for running a testnet simulation in a Devcontainer environment (i.e. GitHub Codespaces).
 
+## Devcontainer
+Devcontainers are a way to package your dev enviornment and run it from a Docker container (or remotely via GitHub Codespaces).
+
+The configuration is stored in the `.devcontainer` folder which contains a `devcontainer.json` file as well as a few install scripts. 
+
+Our devcontainer setup installs:
+* Docker (in Docker)
+* Minikube
+* SDKMan
+
+### Use Devcontainer locally
+To use devcontainers locally, you will need:
+
+* Docker
+* VSCode
+* Devcontainer plugin for VSCode
+
+Once these are installed, you can open the folder in VSCode, and it will prompt you to open the devcontainer environment. From there, you will be able to run the commands below.
+
 ## Installation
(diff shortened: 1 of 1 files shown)


admiraladmirable:
@scasplte2

docs/DevContainerSimulatorGuide.md
```diff
@@ -27,3 +46,9 @@ All pre-requisites should be automatically installed, including: Java, SBT, Dock
     ```
     Note: You can specify a different configuration file by changing the `-f` argument.
 1. View the [results](https://console.cloud.google.com/storage/browser/bifrost-topl-labs-testnet-scenario-results/%2Fsimulation/results)
+
+### Issues
```
hwbehrens, line 50: I might call this "Troubleshooting", since it seems to provide potential error messages you might encounter along with their remedies. I think of "Issues" as more like, hey, here are some bugs we know about and will come back and fix later.

docs/DevContainerSimulatorGuide.md
```diff
@@ -10,7 +29,7 @@ All pre-requisites should be automatically installed, including: Java, SBT, Dock
 
 ### Helm Charts
 1. Run `cd /workspaces`
-1. Run `git clone https://github.com/Topl/helm-charts.git`
+1. Run `sudo git clone https://github.com/Topl/helm-charts.git`
```
hwbehrens, line 32: Would changing this to `sudo git clone https://github.com/Topl/helm-charts.git && sudo chown -R $(stat -c '%u:%g' .) helm-charts` help with making it compatible with a wider range of configurations? That way, if the directory is already owned by the user, it'll stay that way -- my concern is that having root-owned dirs in a place where they're not expected might cause other issues later.

hwbehrens (approved):
Seems reasonable! I guess it's somewhat subjective based on whether people get confused, though.

docs/DevContainerSimulatorGuide.md
admiraladmirable: I ended up changing it to `$(stat -c '%u:%g' ./Bifrost)` because the command was was being run from the `/workspaces` directory and getting the wrong user/group. Updating to `./Bifrost` got the correct user/group.

PR #3020, opened 2023-08-17
BN-1132 Node now could save/reload known hosts to/from the disk

o-aleksandrov-topl:
## Purpose
Node shall be able to save/load known peers to disk, which allows us to not bootstrap network connections every launch from scratch

## Approach
Use data storage to save known peers after peers manager actor shutdown

## Testing
Unit test + multinode test + manual changed multinode test

## Tickets
closes #BN-1132

### DIFF
--- a/node/src/main/scala/co/topl/node/DataStoresInit.scala
+++ b/node/src/main/scala/co/topl/node/DataStoresInit.scala
@@ -9,7 +9,6 @@ import co.topl.blockchain.{CurrentEventIdGetterSetters, DataStores}
 import co.topl.brambl.models.TransactionId
 import co.topl.brambl.models.transaction.IoTransaction
 import co.topl.brambl.syntax._
-import co.topl.codecs.bytes.tetra.TetraScodecCodecs
 import co.topl.codecs.bytes.tetra.instances._
 import co.topl.codecs.bytes.typeclasses.Persistable
 import co.topl.config.ApplicationConfig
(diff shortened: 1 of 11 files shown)


node/src/main/scala/co/topl/node/DataStoresInit.scala
```diff
@@ -23,7 +24,7 @@ import com.google.protobuf.ByteString
 import fs2.io.file.{Files, Path}
 import org.typelevel.log4cats.Logger
 
-object DataStoresInit {
+object DataStoresInit extends TetraScodecCodecs {
```
nandotorterolo, line 27: could you add an import instead of extends it?

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
```diff
@@ -154,4 +152,10 @@ trait TetraScodecCodecs {
       byteStringCodec :: // metadata
       stakingAddressCodec // address
   ).as[UnsignedBlockHeader]
+
+  implicit val knownHostCodec: Codec[KnownHost] = (
+    utf8_32 ::
+      intCodec ::
+      unknownFieldSetCodec
```
nandotorterolo, line 159: I think you don't need to unknownFieldSetCodec

nandotorterolo (approved):
Lgtm, just 2 minnor questions

node/src/main/scala/co/topl/node/DataStoresInit.scala
o-aleksandrov-topl: yeap did it

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraScodecCodecs.scala
o-aleksandrov-topl: we need it, otherwise magic of codec creation doesn't work here :)

PR #3023, opened 2023-08-18
Bn 1159 proposal update

nandotorterolo:
## Purpose

- This PR, create a new step on node-cli for a Transaction Update proposal
- These will use protobuf models described here: https://github.com/Topl/protobuf-specs/pull/83
- There are more task that we should add, but I like to create the first iteration.

## Note
- This PR was not tested on with Brambl-CLI
- This PR is **not** ready to merge

## Tickets
* Bn 1159 

```shell
Welcome to the Bifrost CLI
Please enter a command. [ QUIT | register | proposal]
> proposal
This tool will guide you through the process of preparing a Proposal Update. Press <Enter> to skip.
Please enter label parameter. Ex: Update Slot duration
> aa
Please enter f-effective parameter. Ex:18/2, 9/1, 9
> 
Please enter vrf-ldd-cutoff parameter. Ex:50
> 
Please enter vrf-precision parameter. Ex:40
> 
Please enter vrf-baseline-difficulty parameter. Ex:1/20
> 
Please enter vrf-amplitude parameter. Ex:1/2
> 1
Please enter chain-selection-k-lookback parameter. Ex:50
> 1
Please enter slot-duration parameter. Ex: 1000 milli
> 1 s
Please enter forward-biased-slot-window parameter. Ex:50
> 1
Please enter operational-periods-per-epoch parameter. Ex:2
> 3
Please enter kes-key-hours parameter. Ex:2
> 4
Please enter kes-key-minutes parameter. Ex:9
> 1
Please enter Address parameter. Ex: ptetP7jshHVrEKqDRdKAZtuybPZoMWTKKM2ngaJ7L5iZnxP5BprDB3hGJEFr
> ptetP7jshHVrEKqDRdKAZtuybPZoMWTKKM2ngaJ7L5iZnxP5BprDB3hGJEFr
Using a wallet (i.e. Brambl CLI), create or provide a new LockAddress.
Writing Proposal Update to /tmp/proposalUpdate.transaction.pbuf
This process will save the Proposal update in the current /tmp path
Your Update proposal has been saved. The Transaction that was saved should be imported into Brambl for input selection and broadcast.
Please enter a command. [ QUIT | register | proposal]
> 
Mischief Managed.

```


node/src/main/scala/co/topl/node/cli/ProposalCommand.scala
```diff
@@ -0,0 +1,195 @@
...
+          s.split("/") match {
+            case Array(numerator) => co.topl.node.models.Ratio(numerator.toInt, 1)
+            case Array(numerator, denominator) =>
+              co.topl.node.models.Ratio(numerator.toInt, denominator.toInt)
+            case _ => throw new IllegalArgumentException("Ratio parse error")
+          }
+        }
+
+    implicit private[cli] val parseDuration: String => Try[Duration] = (s: String) =>
+      Try(Duration(scala.concurrent.duration.Duration(s).toSeconds, 0))
```
SeanCheatham, line 176: I think this can just be simplified to `Try(Duration(s))`

SeanCheatham (approved):

node/src/main/scala/co/topl/node/cli/ProposalCommand.scala
nandotorterolo: Thanks for reviewing it, Maybe adding the package name is clear let me explain: 
The first Duration, is not the same as the second one.

> Try(protobuf.duration.Duration(scala.concurrent.duration.Duration(s).toSeconds, 0))

- The input of the cli will be a string, which could be, 1 sec, 100 millis, 1 minute..
- Using the first Library, I convert the string to seconds, and then assign it to protobuf

nandotorterolo: I'm going to merge and then continue with the testing, these changes do not compromise any logic on the node, but for sure more things on protobuf-specs and the cli are required. I will implement it later when I understand what is happening in my testing session. Thanks for reviewing it.

PR #3024, opened 2023-08-22
Support Public Genesis

SeanCheatham:
## Purpose
- Currently, only private testnets are supported which deterministically generate a genesis block from a timestamp
- We will soon need to support public testnets in addition to the main public network
## Approach
- Add the ability to read a FullBlock from a remote destination (either a URL or a local directory)
- Add configuration support for a public network
## Testing
- Private testnets still work, but more testing will be needed for public network
## Tickets
- #BN-1118


nandotorterolo (approved):
LGTM, great work, 
only 1 question, is Htt4s required to read a file from Url, or we can use fs2, to stream this content? I'm fine with both, just curios. (benefits of fs2 is not depend in other library)

SeanCheatham:
> LGTM, great work, only 1 question, is Htt4s required to read a file from Url, or we can use fs2, to stream this content? I'm fine with both, just curios. (benefits of fs2 is not depend in other library)

Thanks for the review Fernando!

I don't think it's possible to make HTTP requests using FS2 alone.  FS2 only seems to support TCP, not HTTP.  http4s is built on cats/fs2 though.

PR #3026, opened 2023-08-25
BN-1127 PeerActor refactoring / added tests

o-aleksandrov-topl:
## Purpose
Improving code quality

## Approach
PeerActor refactoring / added unit tests

## Testing
Unit tests + Integration tests

## Tickets
closes #BN-1127


networking/src/test/scala/co/topl/networking/fsnetwork/TestHelper.scala
```diff
@@ -99,4 +100,44 @@ object TestHelper extends TransactionGenerator {
...
+          case (
+                expectedMessage: RequestsProxy.Message.DownloadHeadersResponse,
+                actualMessage: RequestsProxy.Message.DownloadHeadersResponse
+              ) =>
+            val newResp =
+              actualMessage.response.map { case (header, res) =>
+                (header, res.map(b => b.copy(downloadTimeMs = 0)))
+              }
+            expectedMessage == actualMessage.copy(response = newResp)
+          case (_, _) => throw new IllegalStateException()
```
nandotorterolo, line 119: should contain the message: "Unexpected case"? like the other, function, I am not sure which is the benefit of throwing an Ilegal State, versus a default Match Error Execption.

nandotorterolo (approved):

networking/src/test/scala/co/topl/networking/fsnetwork/TestHelper.scala
o-aleksandrov-topl: yeap I will add message :) I prefer to cover all cases so we will not have compilation warning

PR #3029, opened 2023-08-30
Adding ActorPeerHandlerBridgeAlgebra test

o-aleksandrov-topl:
## Purpose
Adding basic integration test of P2P actor system without running the whole node;
Actor no longer throws "Actor dead" exception for "sendNoWait" messages.

## Approach
Created test for `ActorPeerHandlerBridgeAlgebra` 

## Testing
Unit test itself + integration tests

### DIFF
--- a/actor/src/main/scala/co/topl/actor/Actor.scala
+++ b/actor/src/main/scala/co/topl/actor/Actor.scala
@@ -8,6 +8,7 @@ import cats.syntax.all._
 import co.topl.actor.Actor.ActorId
 import fs2.Stream
 import fs2.concurrent.SignallingRef
+import org.typelevel.log4cats.Logger
 
 import scala.collection.immutable.ListMap
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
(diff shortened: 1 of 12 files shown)


actor/src/main/scala/co/topl/actor/Actor.scala
```diff
@@ -143,10 +143,20 @@ object Actor {
         override def mailboxSize: F[Int] = mailbox.size
 
         override def sendNoWait(input: I): F[Unit] =
-          throwIfDead *> Deferred[F, O].flatMap(promise => mailbox.offer((input, promise).some)).void
+          isAlive.get
+            .ifF(
+              ifTrue = Deferred[F, O].flatMap(promise => mailbox.offer((input, promise).some)).void,
+              ifFalse = Logger[F].error(s"Actor: $actorName; is dead")
```
SeanCheatham, line 149: Would there be drawbacks to raising an ActorDeadException here as well?

networking/src/test/scala/co/topl/networking/fsnetwork/ActorPeerHandlerBridgeAlgebraTest.scala
```diff
@@ -0,0 +1,254 @@
...
+    override def adopt(newHead: Validated.Valid[SlotData]): F[Unit] =
+      (currentHead = newHead.a).pure[F]
+
+    override def head: F[SlotData] =
+      currentHead.copy().pure[F]
+
+    override def genesis: F[SlotData] = genesisSlotData.pure[F]
+  }
+
+  def createClockAlgebra: ClockAlgebra[F] = new ClockAlgebra[F] {
```
SeanCheatham, line 113: Was there a reason you didn't want to use `SchedulerClock.make` here?

SeanCheatham (approved):

actor/src/main/scala/co/topl/actor/Actor.scala
o-aleksandrov-topl: @SeanCheatham In some rare cases it could fail our tests by trying to stop an already dead actor. So I think it is safer to just log it.

networking/src/test/scala/co/topl/networking/fsnetwork/ActorPeerHandlerBridgeAlgebraTest.scala
o-aleksandrov-topl: good point! changed it

PR #3032, opened 2023-09-01
Delay Staking Construction Until Active

SeanCheatham:
## Purpose
- After registering, a staker must wait N+2 epochs before the registration becomes active
- Previously, the staker would immediately attempt to calculate eligibilities but would be unsuccessful
- Now, the staking initialization is delayed until the registration becomes active
## Approach
- Implement functionality to search a stream of block IDs for a particular transaction ID
- Implement a calculation which determines the epoch in which a staker becomes active
- Delay staking initialization until active
- Wrap staking initialization in a resource that gets consumed in the minted block stream/fiber
## Testing
- Updated MultiNodeTest to immediately launch the delayed staker
## Tickets
- #BN-1156


consensus/src/main/scala/co/topl/consensus/interpreters/LocalChain.scala
```diff
@@ -1,60 +1,66 @@
...
+          head.flatMap(chainSelection.compare(_, newHead).map(_ < 0))
 
-          val genesis: F[SlotData] =
-            _g.pure[F]
+        def adopt(newHead: Validated.Valid[SlotData]): F[Unit] = {
+          val slotData = newHead.a
+          Sync[F].uncancelable(_ =>
+            onAdopted(slotData.slotId.blockId) >>
+            headRef.set(slotData) >>
+            EitherT(adoptionsTopic.publish1(newHead.a.slotId.blockId))
```
nandotorterolo, line 43: maybe you can use:
```suggestion
            EitherT(adoptionsTopic.publish1(slotData.slotId.blockId))
```

nandotorterolo (approved):
Looks good to me, just a minor question

blockchain/src/main/scala/co/topl/blockchain/StakingInit.scala
```diff
@@ -111,16 +119,41 @@ object StakingInit {
...
+      registrationHeader <- blockFinder(registrationTransaction.id).toResource
+      _ <- Logger[F]
+        .info(
+          s"Registration transactionId=${registrationTransaction.id.show} found in blockId=${registrationHeader.id.show}"
+        )
+        .toResource
+      registrationEpoch <- clock.epochOf(registrationHeader.slot).toResource
+      // Stakers who are registered in the genesis block have an activation epoch of 0.  Everyone else has an
+      // activation epoch = registration epoch + 2
+      activationEpoch = if (registrationHeader.height == 1) 0L else registrationEpoch + 2
```
o-aleksandrov-topl, line 143: maybe use some constant instead of "1"? something like genesisBlockHeight? But it shall be used everywhere then

o-aleksandrov-topl (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/LocalChain.scala
SeanCheatham: :+1: Updated, thanks!

blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
```diff
@@ -181,7 +180,7 @@ object Blockchain {
           localChain,
           blockHeights,
           blockIdTree,
-          droppingBlockAdoptionsTopic.subscribeUnbounded,
+          Stream.force(localChain.adoptions).dropOldest(10),
```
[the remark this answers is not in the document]
SeanCheatham: DroppingTopic is still fine to use; I just no longer needed it in this particular case.

blockchain/src/main/scala/co/topl/blockchain/StakingInit.scala
SeanCheatham: :+1: Updated to reference `BigBang.Height`

PR #3035, opened 2023-09-08
CLI: Testnet Generator

SeanCheatham:
## Purpose
- To support launching a public testnet, we need a utility which can construct the genesis block and the initial staker configs
## Approach
- Add a new bifrost-cli command `init-testnet`
  - Prompts for initializing 1-or-more stakers
  - Prompts for initializing unstaked TOPLs
  - Prompts for initializing LVLs
  - Prompts for a timestamp
## Testing
- image
## Tickets
- #BN-1163


node/src/main/scala/co/topl/node/cli/InitTestnetCommand.scala
```diff
@@ -0,0 +1,274 @@
...
+    } yield StakerInitializerWithQuantity(initializer, quantity, initializer.registrationTransaction(quantity))
+
+  private val readStakers: StageResultT[F, List[StakerInitializerWithQuantity]] =
+    for {
+      _       <- writeMessage[F]("Please initialize the first staker.")
+      staker0 <- readStaker
+      otherStakers <- List
+        .empty[StakerInitializerWithQuantity]
+        .tailRecM(current =>
+          writeMessage[F]("Add another staker? [ y | N ]") >>
```
o-aleksandrov-topl, line 108: N -- maybe in lowercase? or Y in upper case

o-aleksandrov-topl (approved):

node/src/main/scala/co/topl/node/cli/InitTestnetCommand.scala
SeanCheatham: At this point in the for-comprehension, the user is adding the 2nd (or more) staker.  Since only 1 staker is required, I figured it best to leave this step optional.  This enables the user to smash the "enter" button over-and-over without needing to type anything.  Otherwise, there's no exit-case on the enter-smashing approach.

PR #3040, opened 2023-09-12
[BN-018] Implement gRPC Health Check

admiraladmirable:
## Purpose
Implement gRPC health check for Node.

## Approach
Build and implement the proto based on: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

The implementation uses a map to store the service statuses for Bifrost, Genus, and the overall status noted by an empty string `""`.

This is a first pass as there is no logic for determining an unhealthy status for other services, but currently the Kubernetes gRPC liveliness and readiness probes only use the overall health for the health check.

I believe this should cover most cases initially as long as Bifrost and Genus will throw errors (causing the Kubernetes pod to automatically restart itself), or if the gRPC server hosting goes down, the health check will get an unknown status and restart anyways.

## Testing
* Deployed to GKE, verified that the gRPC health check normally reported healthy.
* Verified the health check restarted the pods when an unhealthy status was returned.
* Verified the health check restarted when nothing (an UKNOWN) status was returned.

~~## TODO~~
~~Merge and use this PR's protobuf specs: https://github.com/Topl/protobuf-specs/pull/89~~ Done!

## Tickets
*

### DIFF
--- a/node/src/main/scala/co/topl/healthcheck/HealthChecker.scala
+++ b/node/src/main/scala/co/topl/healthcheck/HealthChecker.scala
@@ -4,21 +4,23 @@ import cats.{Eq, MonadThrow}
 import cats.effect.{Async, Ref, Resource}
 import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxEq, toFlatMapOps, toFunctorOps}
 import co.topl.algebras.HealthCheckAlgebra
+import co.topl.models.ServiceStatus
 import fs2.Stream
 import fs2.concurrent.SignallingRef
 import grpc.health.v1.{HealthCheckRequest, HealthCheckResponse}
-import grpc.health.v1.HealthCheckResponse.ServingStatus
+import grpc.health.v1.ServingStatus
 import io.grpc.{Status, StatusException}
 
 object HealthChecker {
-  def make[F[_] : Async](
-                          checkRef: Ref[F, Map[String, ServingStatus]],
-                          watchSignal: SignallingRef[F, Option[ServiceStatus]]
-                        ): Resource[F, HealthCheckAlgebra[F, Stream[F, *]]] =
+
+  def make[F[_]: Async](
+    checkRef:    Ref[F, Map[String, ServingStatus]],
+    watchSignal: SignallingRef[F, Option[ServiceStatus]]
+  ): Resource[F, HealthCheckAlgebra[F, Stream[F, *]]] =
     Resource.pure {
       new HealthCheckAlgebra[F, Stream[F, *]] {
 
-        private implicit val eqServingStatus: Eq[ServingStatus] =
+        implicit private val eqServingStatus: Eq[ServingStatus] =
           Eq.fromUniversalEquals
 
         private def getStatus(service: String): F[Option[ServingStatus]] =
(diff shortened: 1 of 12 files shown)


admiraladmirable:
I'll resolve the dependency conflicts tomorrow.

node/src/main/scala/co/topl/healthcheck/HealthChecker.scala
```diff
@@ -0,0 +1,49 @@
...
+
+        private def getStatus(service: String): F[Option[ServingStatus]] =
+          checkRef.get.map(_.get(service))
+
+        override def check(request: HealthCheckRequest): F[HealthCheckResponse] =
+          getStatus(request.service).flatMap {
+            case Some(status) =>
+              HealthCheckResponse(status).pure[F]
+            case None =>
+              MonadThrow[F].raiseError(new StatusException(Status.NOT_FOUND))
```
nandotorterolo, line 32: question, I am confused about the response,  if we watch a service, and it is not in the map, you return ServingStatus.SERVICE_UNKNOWN, but if we check, and it is not there, it will return a Not found exception. Maybe I am reading wrong the code.

nandotorterolo (changes requested):
Great PR, some comments, we can pair to resolve them

node/src/main/scala/co/topl/healthcheck/HealthChecker.scala
admiraladmirable: That part was a bit confusing to me too. From following the [gRPC health check guide](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) they indicate that `SERVICE_UNKNOWN` is meant to be used just during streaming from the `watch` method. 

I'm not sure why they couldn't also use the `UNKOWN` status, but for now I'll just follow the recommendation (even if it is only elaborated on from a single comment 😵‍💫 )

node/src/main/scala/co/topl/healthcheck/HealthChecker.scala
```diff
@@ -0,0 +1,51 @@
...
+object HealthChecker {
+
+  def make[F[_]: Async](
+    checkRef:    Ref[F, Map[String, ServingStatus]],
+    watchSignal: SignallingRef[F, Option[ServiceStatus]]
+  ): Resource[F, HealthCheckAlgebra[F, Stream[F, *]]] =
+    Resource.pure {
+      new HealthCheckAlgebra[F, Stream[F, *]] {
+
+        implicit private val eqServingStatus: Eq[ServingStatus] =
```
nandotorterolo, line 23: I think you can remove this section, becuase is imported on import cats.implicits....

nandotorterolo (approved):

topl-grpc/src/main/scala/co/topl/grpc/HealthCheckGrpc.scala
```diff
@@ -0,0 +1,21 @@
...
+import grpc.health.v1._
+import io.grpc.ServerServiceDefinition
+
+object HealthCheckGrpc {
+  object Server {
+
+    def services[F[_] : Async](
+      healthCheck: HealthCheckAlgebra[F, Stream[F, *]]
+    ): Resource[F, List[ServerServiceDefinition]] =
+      List(
```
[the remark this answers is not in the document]
admiraladmirable: I decided to leave this a list for now since it makes merging lists cleaner down the line and the performance/memory impact will be minimal and only during initialization.

PR #3043, opened 2023-09-14
Use CatsUnsafeResource from BramblSc

SeanCheatham:
## Purpose
- BramblSc recently adopted the CatsUnsafeResource [pattern](https://github.com/Topl/BramblSc/pull/98)
- This PR uses BramblSc's implementation
## Approach
- Remove UnsafeResource and CatsUnsafeResource
- Update references of UnsafeResource to cats.effect.Resource
- Update Bifrost references of CatsUnsafeResource to BramblSc
## Testing
- preparePR
## Tickets
- #BN-1144


consensus/src/test/scala/co/topl/consensus/interpreters/ChainSelectionSpec.scala
```diff
@@ -1,283 +1,277 @@
...
 
-class ChainSelectionSpec
-    extends AnyFlatSpec
-    with ScalaCheckDrivenPropertyChecks
-    with Matchers
-    with MockFactory
-    with EitherValues {
-
-  behavior of "ChainSelection"
+class ChainSelectionSpec extends CatsEffectSuite with ScalaCheckEffectSuite with AsyncMockFactory {
```
nandotorterolo, line 19: thanks for this change, later if you can do the same on 'EtaCalculationSpec', we could remove scala.test framework dependency.

nandotorterolo (approved):

consensus/src/test/scala/co/topl/consensus/interpreters/ChainSelectionSpec.scala
SeanCheatham: I considered trying to remove ScalaTest in this PR, but we still use it in a few other places.  I think we should still do that in a future ticket though.

PR #3044, opened 2023-09-14
BN_Genus_GroupAndSeries_BlockUpdater

nandotorterolo:
## Purpose
- Genus. insert group and series policies

## Testing
- byzantine test ok


genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphBlockUpdater.scala
```diff
@@ -110,9 +113,15 @@ object GraphBlockUpdater {
                   )
                   graph.addEdge(s"class:${addressTxoEdge.name}", lockAddressVertex, txoVertex, addressTxoEdge.label)
 
-                // TODO once we are able to create group-series constructor token, for each output, add here group and series vertices
-                // We are not able at this point, because it requires a specif branch on brambl, which is not merged. Related to Validator discussion
-                // see: https://topl.atlassian.net/wiki/spaces/LABS/pages/678658061/Validation+TAMv2+Transactions+-+Meeting+Fernando+Edmundo+2023-09-08
+                }
+                // Group and Series policies, removing them requires manual intervention
```
SeanCheatham, line 117: Do you have future plans to handle this case in the `remove` step down below?  Otherwise, a rollback in the blockchain could result in an inconsistent state for Genus.

SeanCheatham (approved):

genus-library/src/main/scala/co/topl/genusLibrary/interpreter/GraphBlockUpdater.scala
nandotorterolo: Thanks for reviewing it, Correct, we will have an inconsistent state, Let me discuss with @mundacho about the implication of removing policies when a rollback succeeds, but is correct, this step is missing

nandotorterolo: I included the steps to remove police on rollback, but the existing unit testing of the graph block updater is not the best option to implement it (my bad).  In other PR I will create some integration tests that create some coverage  Genus state on rollback situations.

PR #3047, opened 2023-09-19
Fix: Run Genus Replicator

SeanCheatham:
## Purpose
- A previous PR accidentally removed the line to run the Genus Replicator which causes tests to fail
## Approach
- Re-enable Genus Replicator
- Update Genus Replicator initialization to poll Node RPC instead of sleeping
## Testing
- Re-ran failed tests locally
## Tickets
N/A


nandotorterolo (approved):
thanks for the fix

PR #3048, opened 2023-09-19
BlockPacker Invoke Mempool.remove on Failed Auth

SeanCheatham:
## Purpose
- When a transaction fails authorization validation, it will be retried in the next block until validation succeeds, or until the transaction times out
- This allows an adversary to flood the mempool with transactions that will be constantly re-attempted
## Approach
- In BlockPacker, when authorization fails on a transaction, remove its sub-graph from the mempool
## Testing
- Modify unit test to verify mempool.remove expectations
## Tickets
- #BN-1122
## Note
- This change only benefits block producing nodes.  Other types of nodes may still suffer an attack.

### DIFF
--- a/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
+++ b/minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
@@ -1,5 +1,6 @@
 package co.topl.minting.interpreters
 
+import cats.Show
 import cats.effect._
 import cats.implicits._
 import co.topl.brambl.models.TransactionId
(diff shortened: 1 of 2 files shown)


minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
```diff
@@ -323,11 +328,17 @@ object BlockPacker {
...
               .leftSemiflatTap(error =>
-                Logger[F].warn(s"Transaction ${transaction.id} failed authorization validation: $error")
+                Logger[F].warn(show"Transaction id=${transaction.id} failed authorization validation. reason=$error")
               )
               .isRight
         }
       } yield iterative
   }
 
+  implicit val showAuthorizationError: Show[TransactionAuthorizationError] = {
```
nandotorterolo, line 338: maybe we can group all show instances of Validation Errors, in a common place, Auth errors, and Syntax errors defined on ToplRpcServer

nandotorterolo (approved):

minting/src/main/scala/co/topl/minting/interpreters/BlockPacker.scala
SeanCheatham: I'll create a ticket to explore that.  Centralizing the `Show` instances has benefits for general-purpose usage, but one of the main benefits of `Show` is that it can be defined in an ad-hoc manner in different places, depending on the information that needs to be printed.

PR #3057, opened 2023-09-28
BN-1205 Incoming connection is considered as Cold instead of Warm, PeerActor error handling added

o-aleksandrov-topl:
## Purpose
1. Incoming connections from remote peers shall be treated as COLD initially, otherwise, remote peer could try to manipulate our node connections because HOT connection is selected from the pool of WARM connections
2. Added Peer actor error handling, incorrect genesis leads to COLD state, not BAN
3. Do not request a new connection while going to WARM state from COLD if the connection already existed

## Approach
Do not set WARM state for peer during new incoming connection; Add error handling for PeerActor

## Testing
Unit tests + Integration tests

## Tickets
#BN-1205

### DIFF
--- a/networking/src/main/scala/co/topl/networking/TypedProtocolSetFactory.scala
+++ b/networking/src/main/scala/co/topl/networking/TypedProtocolSetFactory.scala
@@ -322,22 +322,25 @@ object TypedProtocolSetFactory {
         queue <- Queue.bounded[F, OutboundMessage](1)
         stream = Stream.fromQueueUnterminated(queue)
         clientCallback = (query: Query) =>
-          requestPermit.use(_ =>
-            for {
-              deferred <- Deferred[F, Option[T]]
-              _        <- responsePromisesQueue.offer(deferred)
-              _        <- queue.offer(OutboundMessage(TypedProtocol.CommonMessages.Get(query)))
-              result <- EitherT(
-                Async[F].race(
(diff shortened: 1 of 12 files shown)


networking/src/main/scala/co/topl/networking/TypedProtocolSetFactory.scala
```diff
@@ -322,22 +322,25 @@ object TypedProtocolSetFactory {
...
+                deferred <- Deferred[F, Option[T]]
+                _        <- responsePromisesQueue.offer(deferred)
+                _        <- queue.offer(OutboundMessage(TypedProtocol.CommonMessages.Get(query)))
+                result   <- deferred.get
+              } yield result
+
+            EitherT(
+              Async[F].race(
+                Async[F].delayBy(
+                  Async[F].delay(new TimeoutException(s"RequestResponse failed 2 for query=$query")),
```
SeanCheatham, line 337: Maybe a typo in this message?

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/TypedProtocolSetFactory.scala
o-aleksandrov-topl: @SeanCheatham yes, thanks. forgot to remove after debuging

PR #3058, opened 2023-10-03
Load Protocol Settings from Genesis Block

SeanCheatham:
## Purpose
- Protocol settings are currently provided via configuration, but our eventual goal for fork-management is to make those settings controlled by the chain itself
- This PR loads the protocol settings from the genesis block
  - The genesis block must contain exactly one UTxO of type UpdateProposal
  - The UpdateProposal should have all fields provided
  - The UpdateProposal is converted into a Protocol object
- The `init-testnet` command also generates a protocol UTxO, using a set of default settings for now
## Approach
- Add helpers to BigBang for converting to-and-from Protocols and UpdateProposals
- Update init-testnet to output an UpdateProposal in the genesis block
## Testing
- Integration tests and byzantine tests
## Tickets
- #BN-1223

### DIFF
--- a/blockchain/src/main/scala/co/topl/blockchain/BigBang.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/BigBang.scala
@@ -5,10 +5,12 @@ import cats.data.{EitherT, ReaderT}
 import cats.effect.Sync
 import cats.implicits._
 import co.topl.brambl.models._
+import co.topl.brambl.models.box.Value
 import co.topl.brambl.models.transaction._
 import co.topl.brambl.syntax._
 import co.topl.codecs.bytes.tetra.instances._
 import co.topl.codecs.bytes.typeclasses.Transmittable
+import co.topl.config.ApplicationConfig
 import co.topl.consensus.algebras.BlockHeaderToBodyValidationAlgebra
 import co.topl.consensus.models._
 import co.topl.crypto.hash.Blake2b256
@@ -16,6 +18,7 @@ import co.topl.models._
 import co.topl.models.utility.HasLength.instances.byteStringLength
 import co.topl.models.utility._
 import co.topl.node.models._
+import co.topl.numerics.implicits._
 import co.topl.typeclasses.implicits._
 import com.google.common.primitives.Longs
 import com.google.protobuf.ByteString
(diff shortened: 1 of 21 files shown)


blockchain/src/main/scala/co/topl/blockchain/PrivateTestnet.scala
```diff
@@ -144,4 +149,22 @@ object PrivateTestnet {
     NetworkConstants.MAIN_LEDGER_ID
   )
 
+  val DefaultProtocolVersion: ProtocolVersion = ProtocolVersion(2, 0, 0)
+
+  val DefaultProtocol: ApplicationConfig.Bifrost.Protocol =
+    ApplicationConfig.Bifrost.Protocol(
+      minAppVersion = "2.0.0",
```
o-aleksandrov-topl, line 156: should default version be not "2.0.0"?

o-aleksandrov-topl (approved):

blockchain/src/main/scala/co/topl/blockchain/BigBang.scala
```diff
@@ -158,4 +161,59 @@ object BigBang {
...
+      chainSelectionKLookback <- proposal.chainSelectionKLookback.toRight("Missing chainSelectionKLookback")
+      slotDuration            <- proposal.slotDuration.toRight("Missing slotDuration")
+      forwardBiasedSlotWindow <- proposal.forwardBiasedSlotWindow.toRight("Missing forwardBiasedSlotWindow")
+      operationalPeriodsPerEpoch <- proposal.operationalPeriodsPerEpoch.toRight(
+        "Missing operationalPeriodsPerEpoch"
+      )
+      kesKeyHours   <- proposal.kesKeyHours.toRight("Missing kesKeyHours")
+      kesKeyMinutes <- proposal.kesKeyMinutes.toRight("Missing kesKeyMinutes")
+    } yield ApplicationConfig.Bifrost.Protocol(
+      "2.0.0",
```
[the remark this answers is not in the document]
SeanCheatham: This settings is the `minAppVersion`.  We still haven't technically released 2.0.0 since we're still in alpha, but I figure this is what it would start with once we release.

blockchain/src/main/scala/co/topl/blockchain/PrivateTestnet.scala
SeanCheatham: I'm not sure what else we could put here while we're still in alpha.  This value doesn't really mean much right now anyway.

PR #3061, opened 2023-10-05
Replace ScalaTest with mUnit

SeanCheatham:
## Purpose
- Most of our unit tests use mUnit as the underlying framework, but some of our older tests still use ScalaTest
- Test reports aren't being fully reported to GitHub
## Approach
- Update ScalaTest tests to use mUnit
- Semi-related: EtaCalculationSpec used a singleton instance of Ed25519VRF which occasionally caused test failures.  Updated to not use a shared singleton where applicable.
- Update the test reports upload directory
## Testing
- `preparePR`
## Tickets
- #BN-1222


node-crypto/src/main/scala/co/topl/crypto/models/KesBinaryTree.scala
```diff
@@ -13,9 +13,47 @@ object KesBinaryTree {
...
+    }
+  }
+
+  final case class SigningLeaf(sk: Array[Byte], vk: Array[Byte]) extends KesBinaryTree {
+
+    override def equals(obj: Any): Boolean =
+      obj match {
+        case s: SigningLeaf =>
+          java.util.Arrays.equals(sk, s.sk) &&
+          java.util.Arrays.equals(vk, s.vk)
```
nandotorterolo, line 47: case _ =>
          false

node-crypto/src/main/scala/co/topl/crypto/models/KesBinaryTree.scala
```diff
@@ -13,9 +13,47 @@ object KesBinaryTree {
...
+          java.util.Arrays.equals(witnessRight, m.witnessRight) &&
+          left == m.left &&
+          right == m.right
+        case _ =>
+          false
+      }
+
+    override def hashCode(): Int = {
+      var r = 1
+      r = 31 * r + java.util.Arrays.hashCode(seed)
```
nandotorterolo, line 32: question: where does '31' come from? it comes from the length of (32 bytes -1)

nandotorterolo (approved):
Great refactor, I didn't be aware that removing ScalaTest would require so many changes places.

node-crypto/src/main/scala/co/topl/crypto/models/KesBinaryTree.scala
SeanCheatham: :+1: Good catch

PR #3062, opened 2023-10-08
BN-1167 Save reputation for peer during node shutdown, neighbours peer have initial reputation equal to their source. Use that reputation to move peer to warm and hot state

o-aleksandrov-topl:
## Purpose
1. Save remote reputation during node shutdown
2. New received peer have an initial reputation equal to source of those peers
3. Use reputation for selecting warm and hot peers

## Approach
Save peers reputation to data store, update reputation based for source for new peers

## Testing

## Tickets
closes #BN-1167


networking/src/main/scala/co/topl/networking/fsnetwork/NetworkManager.scala
```diff
@@ -36,10 +35,10 @@ object NetworkManager {
     headerStore:                 Store[F, BlockId, BlockHeader],
     bodyStore:                   Store[F, BlockId, BlockBody],
     transactionStore:            Store[F, TransactionId, IoTransaction],
-    knownHostsStore:             Store[F, Unit, Seq[KnownHost]],
+    knownHostsStore:             Store[F, Unit, Seq[RemotePeer]],
```
nandotorterolo, line 38: suggestion, DataStoresInit, defines this store with the name "knownRemotePeersStore", maybe we can align names. Maybe knowHostStore, is better for business logic?

nandotorterolo (approved):

PR #3066, opened 2023-10-10
Support Remote Configuration File

SeanCheatham:
## Purpose
- Loading a configuration file from a remote location may help users with onboarding
## Approach
- Update IOBaseApp to support loading configuration via HTTP
  - Also updated initialization functions to use `F` context
## Testing
- Update NodeAppTest such that one of the nodes loads its configuration from HTTP
- AdHoc Test [Success](https://github.com/Topl/Bifrost/actions/runs/6473003420)
## Tickets
- #BN-1220

### DIFF
--- a/common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
+++ b/common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
@@ -1,13 +1,15 @@
 package co.topl.common.application
 
-import cats.data.Chain
+import cats.data.{Chain, Nested}
 import cats.effect._
 import cats.implicits._
 import cats.kernel.Monoid
 import com.typesafe.config.{Config, ConfigFactory}
+import fs2.io.file.Files
+import fs2.io.net.Network
+import org.http4s.client.middleware.FollowRedirect
+import org.http4s.ember.client.EmberClientBuilder
 import pureconfig.{ConfigObjectSource, ConfigSource}
-
-import java.nio.file.Paths
 
 /**
  * Assists with constructing applications which use cats-effect  Initializes the runtime and configurations.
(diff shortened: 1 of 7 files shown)


common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
```diff
@@ -19,93 +21,94 @@ import java.nio.file.Paths
...
-        argsToUserConfigs(cmdArgs).map(_.config()) ++
-        Chain(ConfigSource.default).map(_.config())
-    ).combineAll match {
-      case Right(value) => value.resolve()
-      case Left(e)      => throw new IllegalStateException(e.toString)
-    }
+  def createTypesafeConfig[Args: ContainsUserConfigs: ContainsDebugFlag](cmdArgs: Args): IO[Config] =
+    for {
+      debugConfigs <- Nested(argsToDebugConfigs(cmdArgs)).map(_.config()).value
+      environmentConfigs = Chain.fromOption(ConfigSource.resources("environment.conf").config().toOption).map(_.asRight)
```
nandotorterolo, line 60: question, if there is no environment.conf file, we won't fail right? just asking because the behavior changed.

common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
```diff
@@ -19,93 +21,94 @@ import java.nio.file.Paths
...
+  preInitFunction: AppConfig => IO[Unit] = (_: AppConfig) => ().pure[IO]
 ) {
 
   type F[A] = IO[A]
 
-  protected[application] var _args: CmdArgs = _
-  protected[application] var _config: Config = _
-  protected[application] var _appConfig: AppConfig = _
-  protected[application] var _ioApp: IOApp = _
+  def run(cmdArgs: CmdArgs, config: Config, appConfig: AppConfig): IO[Unit]
```
nandotorterolo, line 32: question: could config be removed from run signature? maybe not, but I don't see how we are using it. Maybe I am not following the code

nandotorterolo (approved):

common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
SeanCheatham: Correct, it will still succeed even if environment.conf does not exist.  The implementation changed, but the result should be the same.  Previously, there would be a `Chain` with a single empty `Config` object, but now it will just be an empty `Chain` because of the `.fromOption`

common-application/src/main/scala/co/topl/common/application/IOBaseApp.scala
SeanCheatham: It _could_, but I figured there's not much harm in including it.  `def config` used to be available as a method so I thought it best to still make it available even if we don't currently use it.

PR #3072, opened 2023-10-18
BN-1238 Move common ancestor tracking to Actor P2P, remove legacy P2P

o-aleksandrov-topl:
## Purpose
Move common ancestor tracking to Actor P2P, remove legacy P2P

## Approach
Adding special request to peer actor

## Testing

## Tickets
closes #BN-1238


config/src/main/scala/co/topl/config/ApplicationConfig.scala
```diff
@@ -52,7 +52,6 @@ object ApplicationConfig {
     )
 
     case class NetworkProperties(
-      legacyNetwork:                        Boolean = false,
```
SeanCheatham, line 55: I think this can be removed from application.conf too.

SeanCheatham (approved):

config/src/main/scala/co/topl/config/ApplicationConfig.scala
o-aleksandrov-topl: @SeanCheatham yeap done

PR #3075, opened 2023-10-24
[BN-1245]: Support Proxy Hosted P2P Connections

admiraladmirable:
## Purpose
In our current cloud hosting setup, we are using a L7 load balancer which is behind a proxy. Normally this can be accessed via hostname (i.e. testnet.topl.tech), but our p2p code only supported direct IP based connections because we were resolving hosts to an IP before storing/connecting peers.

This caused an issue where we would connect to the IP of the proxy which would not forward traffic to the application.

## Approach
* Add `--knownPeersUnresolvedDns` as an argument which accepts a similar list of comma separated peers that we skip resolving DNS for.
* Update Peer classes to add `resolveDns` option to control which peers resolve.

## Testing
* Successfully connected to `testnet.topl.tech:9085`.

## Tickets
* BN-1245

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
@@ -1,7 +1,7 @@
 package co.topl.networking
 
 import cats.data.{NonEmptyChain, OptionT}
-import cats.effect.Async
+import cats.effect._
 import cats.implicits._
 import cats.{Applicative, Monad, MonadThrow, Show}
 import co.topl.algebras.Store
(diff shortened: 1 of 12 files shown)


admiraladmirable:
I realize that adding a different `knownPeers` argument can cause confusion and is not the most user friendly.

I see the value in resolving DNS for direct peers so I didn't want to remove that feature entirely.

Open to feedback!

networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
```diff
@@ -265,13 +265,17 @@ package object fsnetwork {
...
+        if (unresolvedHost.resolveDns)
+          resolver.resolving(unresolvedHost.host).map(_.map(resolved => unresolvedHost.copy(host = resolved)))
+        else Applicative[F].pure(Option(unresolvedHost))
       }
 
     implicit def peerToAddResolver[F[_]: Monad: DnsResolver]: DnsResolverHT[RemotePeer, F] =
       (unresolvedHost: RemotePeer) => {
         val resolver = implicitly[DnsResolverHT[RemoteAddress, F]]
-        resolver.resolving(unresolvedHost.address).map(_.map(resolved => unresolvedHost.copy(address = resolved)))
+        if (unresolvedHost.address.resolveDns)
```
o-aleksandrov-topl, line 276: it shall work without adding that code `resolveDns` could work on `RemoteAddress` level only

node/src/main/scala/co/topl/node/ApplicationConfigOps.scala
```diff
@@ -94,19 +97,26 @@ object ApplicationConfigOps {
...
 
   implicit val knownPeersReader: ConfigReader[List[KnownPeer]] =
     ConfigReader[String].emap(str =>
       Try(
-        parseKnownPeers(str)
+        parseKnownPeers(str, false)
       ).toEither.leftMap(e => error.CannotConvert(str, "InetAddressList", e.getMessage))
     )
 
+  // implicit val knownPeersUnresolvedDnsReader: ConfigReader[List[KnownPeer]] =
```
o-aleksandrov-topl, line 113: remove commented code

o-aleksandrov-topl (commented):
Overall how does it resolve the issue at all? Is it only about saving known peers? because before actual initiating connection hostname will be resolved to IP address anyway

admiraladmirable:
@o-aleksandrov-topl I made some assumptions on how Istio proxies TCP traffic and it looks like it won't work how I expected things. 

When connecting via TCP through the Gateway, it will always resolve directly to the load balancer IP and doesn't support any host name proxying. Going to close this PR as the issue needs to be resolved on the deployment side of things.

PR #3076, opened 2023-10-24
Bifrost CLI: init-testnet Generate Config File

SeanCheatham:
## Purpose
- It would be convenient to users if the `init-testnet` command also generated and saved a sample config file to easily launch the node
## Approach
- Generate and save a config file in the root of the output directory
## Testing
- Local/manual testing
## Tickets
- #BN-1221


nandotorterolo (approved):
lgtm

PR #3082, opened 2023-10-31
BodySemanticValidation Check Reward Input Address

SeanCheatham:
## Purpose
- Reward Transactions are able to construct tokens without actually spending any input
- To keep Reward Transactions unique, a single input is expected on the Transaction, and that single input is expected to reference the parent block ID as its value
## Approach
- In BodySemanticValidation, enforce that the reward transaction's input address contains the same bytes as the block's parent header ID
## Testing
- New unit tests
## Tickets
- #BN-1256
## Unrelated
- Fixed a few compiler warnings in unit tests by adding `@unchecked` annotations

### DIFF
--- a/ledger/src/main/scala/co/topl/ledger/interpreters/BodySemanticValidation.scala
+++ b/ledger/src/main/scala/co/topl/ledger/interpreters/BodySemanticValidation.scala
@@ -1,6 +1,6 @@
 package co.topl.ledger.interpreters
 
-import cats.data.{EitherT, NonEmptyChain, Validated, ValidatedNec}
+import cats.data.{EitherT, NonEmptyChain, OptionT, Validated, ValidatedNec}
 import cats.effect.Sync
 import cats.implicits._
 import co.topl.brambl.models.TransactionId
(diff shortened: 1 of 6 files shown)


ledger/src/main/scala/co/topl/ledger/interpreters/BodySemanticValidation.scala
```diff
@@ -76,6 +80,25 @@ object BodySemanticValidation {
                 )
               )
           } yield transaction
+
+        /**
+         * Verify that the reward transaction (if included) attempts to "spend" the parent header ID.
+         */
+        private def validateReward(context: BodyValidationContext)(body: BlockBody) =
```
nandotorterolo, line 87: question, when you said on the comment:  "a single input is expected ", does it mean that this function should validate that rewards.input.size <= 1, or it never be possible by construction?

nandotorterolo (approved):
thanks for fixing warnings!

ledger/src/main/scala/co/topl/ledger/interpreters/BodySemanticValidation.scala
SeanCheatham: `BodySyntaxValidation` validates that `rewards.input.size == 1`, and that step takes place before `BodySemanticValidation`

nandotorterolo: thanks!

PR #3084, opened 2023-11-01
Load Config from Environment

SeanCheatham:
## Purpose
- To support upcoming work to allow the Bifrost CLI to generate a config file, the node needs a way to load a config file from a predictable location
## Approach
- Use an environment variable to reference a file location of a Bifrost configuration file
  - When empty, no-op
  - When provided, load and parse the file as a config
  - When provided but non-existent on disk, log and continue
- Also, updated the Docker settings
  - Volumes collapsed into `/bifrost`
  - Environment variables will save data to `/bifrost/data` and `/bifrost/staking` instead of the default `/tmp`
  - Configuration file will be loaded from `/bifrost/config/user.yaml` if it exists
    - In the future, this is where the CLI will save a user's configuration file
## Testing
- Manual testing with successive launches of the node
## Tickets
- #BN-1258

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
+import com.typesafe.sbt.packager.docker.DockerChmodType
 import sbt.Keys.{organization, test}
 import sbtassembly.MergeStrategy
+import NativePackagerHelper._
 
 val scala213 = "2.13.12"
 
@@ -66,8 +68,11 @@ lazy val dockerSettings = Seq(
 lazy val nodeDockerSettings =
   dockerSettings ++ Seq(
     dockerExposedPorts := Seq(9084, 9085),
-    dockerExposedVolumes ++= Seq("/bifrost-data", "/bifrost-staking", "/bifrost-config"),
-    Docker / packageName := "bifrost-node"
+    dockerExposedVolumes ++= Seq("/opt/docker/.bifrost/data", "/opt/docker/.bifrost/staking", "/opt/docker/conf"),
+    Docker / packageName := "bifrost-node",
+    Docker / mappings += (file("node/docker/conf/docker.yaml"), "/opt/docker/conf/docker.yaml"),
+    dockerEnvVars ++= Map("BIFROST_CONFIG_FILE" -> "conf/docker.yaml"),
+    dockerChmodType := DockerChmodType.UserGroupWriteExecute
   )
 
 lazy val networkDelayerDockerSettings =
--- a/build.sbt
+++ b/build.sbt
@@ -70,9 +70,10 @@ lazy val nodeDockerSettings =
     dockerExposedPorts := Seq(9084, 9085),
     Docker / packageName := "bifrost-node",
     dockerExposedVolumes += "/bifrost",
+    dockerExposedVolumes += "/bifrost-staking",
     dockerEnvVars ++= Map(
       "BIFROST_APPLICATION_DATA_DIR" -> "/bifrost/data/{genesisBlockId}",
-      "BIFROST_APPLICATION_STAKING_DIR" -> "/bifrost/staking/{genesisBlockId}",
+      "BIFROST_APPLICATION_STAKING_DIR" -> "/bifrost-staking/{genesisBlockId}",
       "BIFROST_CONFIG_FILE" -> "/bifrost/config/user.yaml"
     )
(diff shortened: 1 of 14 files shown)


admiraladmirable (approved):
🚀

byzantine-it/src/test/scala/co/topl/byzantine/MultiNodeTest.scala
```diff
@@ -69,16 +70,14 @@ class MultiNodeTest extends IntegrationSuite {
...
-          .use(client =>
-            OptionT(client.blockIdAtHeight(1))
-              .flatMapF(client.fetchBlockBody)
-              .map(_.transactionIds.head)
-              .flatMapF(client.fetchTransaction)
-              .getOrRaise(new IllegalStateException)
-          )
-          .toResource
+        client         <- node0.rpcClient[F](node0.config.rpcPort)
+        genesisBlockId <- OptionT(client.blockIdAtHeight(1)).getOrRaise(new IllegalStateException).toResource
```
o-aleksandrov-topl, line 74: could you replace 1 to constant?

node/src/main/scala/co/topl/node/NodeApp.scala
```diff
@@ -88,18 +92,24 @@ class ConfiguredNodeApp(args: Args, appConfig: ApplicationConfig) {
 
       bigBangBlockId = bigBangBlock.header.id
       bigBangSlotData <- dataStores.slotData.getOrRaise(bigBangBlockId).toResource
-      _               <- Logger[F].info(show"Big Bang Block id=$bigBangBlockId").toResource
+      _ <- Logger[F].info(show"Big Bang Block id=$bigBangBlockId timestamp=${bigBangBlock.header.timestamp}").toResource
 
       stakingDir = Path(interpolateBlockId(bigBangBlockId)(appConfig.bifrost.staking.directory))
-      _ <- Files[F].createDirectories(stakingDir).toResource
```
SeanCheatham, line 94: @scasplte2 This line here is probably what triggered the permission issue you saw when trying  to run a simple relay node.  With it removed, it should no longer try to create the new directories (unless it's a private testnet and the node is configured as a staker).

documentation/docs/reference/Installing Your Node/docker.md
```diff
@@ -11,10 +11,11 @@ tags:
 ## Quickstart
 
 ```sh
-mkdir ./bifrost-data
-sudo chown  1001 ./bifrost-data
-docker run -d --name bifrost-tetra $(pwd)/bifrost-data:/tmp/bifrost -p 9084:9084 -p 9085:9085 toplprotocol/bifrost-node:2.0.0-alpha10
+mkdir .bifrost
```
SeanCheatham, line 14: @scasplte2 Just wanted to call your attention to this section here.  I think the step for `sudo chown` is new, but it's important for allowing the node to write/modify the data and staking directories.  The gist is to create a directory on the host machine (named `.bifrost`), and update the "owner" of that directory to be the same user as the one in the docker container.  In this case, it's `1001` (the default from sbt-native-packager, though we could override it if we want, perhaps to `7091`?)

build.sbt
```diff
@@ -66,8 +68,13 @@ lazy val dockerSettings = Seq(
 lazy val nodeDockerSettings =
   dockerSettings ++ Seq(
     dockerExposedPorts := Seq(9084, 9085),
-    dockerExposedVolumes ++= Seq("/bifrost-data", "/bifrost-staking", "/bifrost-config"),
-    Docker / packageName := "bifrost-node"
+    Docker / packageName := "bifrost-node",
+    dockerExposedVolumes += "/bifrost",
```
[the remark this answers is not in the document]
SeanCheatham: And just to clarify, the reason I opted to keep `/bifrost` and `/bifrost-staking` separate is so that we can recommend to our users that they also keep those files separate on their host machines.

o-aleksandrov-topl (approved):

byzantine-it/src/test/scala/co/topl/byzantine/MultiNodeTest.scala
SeanCheatham: :+1: Updated

PR #3087, opened 2023-11-03
BN-1243 Add option to expose server port to P2P network or not

o-aleksandrov-topl:
## Purpose
Some nodes have no accessible  server port, thus incoming connections are not available, added an option to not propagate such inaccessible node addresses to P2P network

## Approach
Clear known server prot after established connection, request port from the remote peer. Remote peer send their own server port only if the appropriate option is set to true

## Testing
Unit test + integration test

## Tickets
closes BN-1243

### DIFF
--- a/config/src/main/scala/co/topl/config/ApplicationConfig.scala
+++ b/config/src/main/scala/co/topl/config/ApplicationConfig.scala
@@ -48,6 +48,7 @@ object ApplicationConfig {
       publicHost:        String,
       publicPort:        Int,
       knownPeers:        List[KnownPeer],
+      exposeServerPort:  Boolean,
       networkProperties: NetworkProperties
     )
 
--- a/blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
+++ b/blockchain/src/main/scala/co/topl/blockchain/Blockchain.scala
@@ -59,6 +59,7 @@ object Blockchain {
     nodeProtocolConfiguration: ProtocolConfigurationAlgebra[F, Stream[F, *]],
     additionalGrpcServices:    List[ServerServiceDefinition],
     _epochData:                EpochDataAlgebra[F],
+    exposeServerPort:          Boolean,
     networkProperties:         NetworkProperties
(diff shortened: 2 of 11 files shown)


config/src/main/scala/co/topl/config/ApplicationConfig.scala
```diff
@@ -48,6 +48,7 @@ object ApplicationConfig {
       publicHost:        String,
       publicPort:        Int,
       knownPeers:        List[KnownPeer],
+      exposeServerPort:  Boolean,
```
SeanCheatham, line 51: I think we should instead use `publicPort` (and eventually `publicHost`) for this, but make them optional.  When `publicPort` is set, that's the value that gets used for `co.topl.networking.blockchain.BlockchainPeerServerAlgebra#serverPort`, otherwise None.  That was generally the intention of the publicHost/publicPort settings.

SeanCheatham (approved):

config/src/main/scala/co/topl/config/ApplicationConfig.scala
o-aleksandrov-topl: @SeanCheatham overall good idea, but I'm not sure about "public-host/port" for Blockchain. It could be used for some kind of DDoS attack: we could send wrong/bad information about "public-host:public-port" to other peers so they try to do connection to that other peer. It is more or less safe to send port number and use actual IP address though

SeanCheatham: The port information is already provided voluntarily though, so an adversary can already perform that sort of attack.  Alternatively, an adversary could do something similar when advertising its set of peers.  And generally speaking, an IP address can be spoofed anyway.  By not allowing the user to configure the settings, doesn't that only really harm the honest participants?

o-aleksandrov-topl: @SeanCheatham ok I will create a follow-up ticket about it including public port and public host. Anyway it shall be done together.

PR #3090, opened 2023-11-06
Bifrost CLI: "Configure"

SeanCheatham:
## Purpose
- To assist with user on-boarding, a utility can guide users through the configuration process
## Approach
- New "configure" CLI command
  - Prompt user for common settings, default to None for all
  - Save settings as YAML to disk.  Default location is the value of the `BIFROST_CONFIG_FILE` environment variable.  If file exists, merge-and-overwrite.
## Testing
- User workflow
  - `mkdir .bifrost`
  - `sudo chown 1001 .bifrost`
  - `docker run --rm -it -v $(pwd)/.bifrost:/bifrost toplprotocol/bifrost-node cli`
  - `docker run --rm -it -v $(pwd)/.bifrost:/bifrost toplprotocol/bifrost-node`
## Tickets
- #BN-1295

### DIFF
--- a/node/src/main/resources/logback.xml
+++ b/node/src/main/resources/logback.xml
@@ -1,5 +1,7 @@
 <configuration debug="false">
 
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
+
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
             <pattern>%d{HH:mm:ss.SSS}| %-5level %logger{36} - %msg%n</pattern>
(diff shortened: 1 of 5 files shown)


node/src/main/resources/logback.xml
```diff
@@ -1,5 +1,7 @@
 <configuration debug="false">
 
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
```
SeanCheatham, line 3: This removes the logback preamble at the beginning of every run.

nandotorterolo (approved):

PR #3100, opened 2023-11-11
Added option to use hostnames instead of ip addresses

o-aleksandrov-topl:
## Purpose
Use hostnames instead of ip addresses for saving known peers / propagate known P2P hosts

## Approach
do reverse resolving of ip addresses before saving peers / propagate them to the network

## Testing
Unit tests + integration test

## Tickets
closes #BN-1294

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
@@ -17,11 +17,13 @@ import co.topl.typeclasses.implicits._
 import com.comcast.ip4s.{Dns, Hostname, IpAddress}
 import com.github.benmanes.caffeine.cache.{Cache, Caffeine}
 import org.typelevel.log4cats.Logger
+import scalacache.Entry
+import scalacache.caffeine.CaffeineCache
 import scodec.Codec
 import scodec.codecs.{cstring, double, int32}
 
 import java.time.Duration
-import scala.concurrent.duration.{FiniteDuration, MILLISECONDS}
+import scala.concurrent.duration.{DurationInt, FiniteDuration, MILLISECONDS}
 
 package object fsnetwork {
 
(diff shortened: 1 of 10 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
```diff
@@ -260,18 +265,40 @@ package object fsnetwork {
...
+      val expireAfterWriteDuration: Duration = java.time.Duration.ofMinutes(30)
+
+      val cache: Cache[HostId, HostId] =
+        Caffeine.newBuilder
+          .maximumSize(dnsCacheSize)
+          .expireAfterWrite(expireAfterWriteDuration)
+          .build[String, String]()
+
+      override def resolving(host: HostId): F[Option[HostId]] =
+        if (cache.contains(host)) {
```
SeanCheatham, line 279: I think you could use `cache.cachingF` here

networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
```diff
@@ -293,9 +320,78 @@ package object fsnetwork {
...
+      val expireAfterWriteDuration: Duration = java.time.Duration.ofMinutes(30)
+
+      val cache: Cache[HostId, HostId] =
+        Caffeine.newBuilder
+          .maximumSize(reverseDnsCacheSize)
+          .expireAfterWrite(expireAfterWriteDuration)
+          .build[String, String]()
+
+      override def reverseResolving(hostIdAsIp: HostId): F[HostId] =
+        Option(cache.getIfPresent(hostIdAsIp))
```
SeanCheatham, line 350: Same here: `cache.cachingF`

SeanCheatham (approved):

networking/src/main/scala/co/topl/networking/fsnetwork/package.scala
o-aleksandrov-topl: @SeanCheatham it requires writing to cache, but in that case for cache, the calculation of a new value is not always possible.

PR #3106, opened 2023-11-17
BN-1285 Fixing possible lost request for data downloading; Fixing issue trying verify header too early

o-aleksandrov-topl:
## Purpose
In some rare cases, we could lose the request to download remote data, in that case, the sync is stopped. Added request timeout, thus we will be able to recover.
Block checker could try to apply header which cannot be verified (because bodies for N-2 epoch are missed). Fix it by checking could header be verified first.

## Approach
Added timeout for data download request, thus we could recover if we receive no response. Block checker will pause header applying if we can't verify the header. Header will be requested again as soon as body validation reach the header which can't be verified in past.

## Testing
Unit test + integration test

## Tickets
closes BN-1285

### DIFF
--- a/consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
+++ b/consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
@@ -74,11 +74,11 @@ object BlockHeaderValidation {
     blake2b256Resource:       Resource[F, Blake2b256]
   ) extends BlockHeaderValidationAlgebra[F] {
 
-    def couldBeValidated(header: BlockHeader, currentLocalBestBlock: SlotData): F[Boolean] =
+    def couldBeValidated(header: BlockHeader, lastProcessedBodyInChain: SlotData): F[Boolean] =
       for {
-        headerEpoch    <- clockAlgebra.epochOf(header.slot)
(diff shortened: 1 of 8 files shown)


consensus/src/main/scala/co/topl/consensus/interpreters/BlockHeaderValidation.scala
```diff
@@ -73,6 +74,12 @@ object BlockHeaderValidation {
     blake2b256Resource:       Resource[F, Blake2b256]
   ) extends BlockHeaderValidationAlgebra[F] {
 
+    def couldBeValidated(header: BlockHeader, currentLocalBestBlock: SlotData): F[Boolean] =
+      for {
+        headerEpoch    <- clockAlgebra.epochOf(header.slot)
+        bestBlockEpoch <- clockAlgebra.epochOf(currentLocalBestBlock.slotId.slot)
```
SeanCheatham, line 80: This result can't really be determined from the "current head" since the local head and the remote head might be in the same epoch but have a very distant common ancestor.  It would need to be based on the last saved block body along the remote header's tine.

o-aleksandrov-topl: @SeanCheatham yeap good catch!

SeanCheatham (approved):

PR #3111, opened 2023-11-20
Update Protocol Setting Defaults

SeanCheatham:
## Purpose
- Updating protocol settings to reflect expected simulation results
## Approach
- Modified settings
## Testing
Successful Ad-Hoc Test run: https://github.com/Topl/Bifrost/actions/runs/6936355065
## Tickets
N/A

### DIFF
--- a/node/src/main/scala/co/topl/node/cli/InitTestnetCommand.scala
+++ b/node/src/main/scala/co/topl/node/cli/InitTestnetCommand.scala
@@ -39,13 +39,13 @@ object InitTestnetCommand {
   private[cli] val DefaultProtocol =
     ApplicationConfig.Bifrost.Protocol(
       minAppVersion = "2.0.0",
-      fEffective = Ratio(15, 100),
-      vrfLddCutoff = 50,
+      fEffective = Ratio(12997, 100000),
+      vrfLddCutoff = 15,
       vrfPrecision = 40,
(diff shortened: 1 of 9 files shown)


node/src/main/scala/co/topl/node/cli/InitTestnetCommand.scala
```diff
@@ -39,13 +39,13 @@ object InitTestnetCommand {
...
+      vrfLddCutoff = 15,
       vrfPrecision = 40,
-      vrfBaselineDifficulty = Ratio(1, 20),
-      vrfAmplitude = Ratio(1, 2),
-      // 10x private testnet default, resulting in ~50 minute epochs
-      chainSelectionKLookback = 500,
+      vrfBaselineDifficulty = Ratio(5, 100),
+      vrfAmplitude = Ratio(50, 100),
+      // 20x private testnet default, resulting in ~100 minute epochs
+      chainSelectionKLookback = 1000,
```
SeanCheatham, line 48: @admiraladmirable Just calling out this increase here.  Epochs will be 100 minutes instead of 50

SeanCheatham: This also means the operational periods will be longer, so it may take a couple extra minutes for a block producer to recover after restart.

PR #3112, opened 2023-11-21
Iterative Improvements, Transaction Generator Improvements

SeanCheatham:
## Purpose
- Implementation of Iterative#run might have leaked a fiber (unsure)
## Approach
- Update Iterative to return a Resource, and use `.background`
## Testing
- Ran local node with transaction generator
## Tickets
N/A


cats-utils/src/main/scala/co/topl/catsutils/Iterative.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Continually invoke the given iterative on a "current state" until a signal is received to halt.  Upon receiving
+   * the signal, the current best item is captured, the iterative process is cancelled, and the best item returned.
+   */
+  def run[F[_]: Async, E](init: F[E])(iterative: Iterative[F, E]): Resource[F, F[E]] =
+    for {
+      initialValue <- init.toResource
+      // A Deferred which accepts a single signal from the outside indicating that the Iterative process should stop
+      stopDeferred <- Deferred[F, Either[Throwable, Unit]].toResource
```
SeanCheatham, line 33: Main changes:
- Stop signal is now a Deferred instead of a Queue
- Result is captured in a Ref instead of a Queue
- Iterative process is performed in a Stream and uses an interruption signal.  Also, the fiber is captured via `.background` which returns a Resource, so the fiber should auto-cancel when the Resource closes

nandotorterolo (approved):

PR #3122, opened 2023-11-28
QuickFix for using remote node id instead of IP

o-aleksandrov-topl:
## Purpose
Filter known remote peers by their Id's not ip's

## Approach
Filter known remote peers by their Id's not ip's

## Testing
Unit tests
## Tickets
*


hwbehrens (approved):
Fingers crossed!

Issue #50, opened 2018-05-14
One page for the Whitepaper


Tuxman:
A single page describing the tech stack Topl uses in our platform.

PR #3137, opened 2023-12-01
Level DB Tuning

SeanCheatham:
## Purpose
- I observed that LevelDB is causing a major slowdown after period of block synchronization
## Approach
- Incorporate LevelDBJNI, using the approach from Dion
  - If LevelDBJNI is available, use it.  Otherwise, fallback to Java-LevelDB implementation
## Testing
- Connected to testnet.topl.tech relay node with and without the change; observed faster sync with JNI
## Tickets
N/A

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -610,7 +610,7 @@ lazy val levelDbStore = project
   )
   .dependsOn(
     byteCodecs,
-    algebras,
+    algebras % "compile->compile;test->test",
     catsUtils
   )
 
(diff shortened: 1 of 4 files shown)


nandotorterolo (approved):
- Tested: Uptime: 14 hrs 36 min 03 sec
- cpu < 10 %, heap 500 - 1gb ~
- Linux fernando-um700 5.19.0-50-generic #50-Ubuntu SMP PREEMPT_DYNAMIC Mon Jul 10 18:24:29 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
-  Model name:           AMD Ryzen 7 3750H with Radeon Vega Mobile Gfx
   CPU family:         23,   Model:              24   Thread(s) per core: 2,   Core(s) per socket: 4

PR #3144, opened 2023-12-04
Implement "IdleApp"

SeanCheatham:
## Purpose
- Sometimes it is helpful to perform data backups for a node, but the node must not write data during the backup process
- For Docker deployments, the process/command needs to be running for the container to exist, so the node needs to be active in order for a user to exec/shell into the container
## Approach
- Create a new application mode which runs forever but does nothing
  - Except for launching the healthcheck, which will be necessary for kubernetes deployments
## Testing
image
## Tickets
N/A


o-aleksandrov-topl (approved):

admiraladmirable (approved):
🥇

PR #3146, opened 2023-12-04
Config/Arg to Override LevelDB Implementation

SeanCheatham:
## Purpose
- New versions are incompatible with old versions for both networking and storage, which means it is impossible to bootstrap/synchronize the two versions
## Approach
- Allow overriding LevelDB storage to support levelDb-java when needed
## Testing
- Launch node without flag and verified JNI was loaded.  Launched node with flag `--databaseType levelDb-java` and verified Java was loaded
## Tickets
N/A

### DIFF
--- a/config/src/main/scala/co/topl/config/ApplicationConfig.scala
+++ b/config/src/main/scala/co/topl/config/ApplicationConfig.scala
@@ -37,7 +37,7 @@ object ApplicationConfig {
   object Bifrost {
 
     @Lenses
-    case class Data(directory: String)
+    case class Data(directory: String, databaseType: String)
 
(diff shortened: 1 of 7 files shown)


o-aleksandrov-topl (approved):

SeanCheatham:
> @SeanCheatham shall we raise warning message about slow node performance if we use java version?

@o-aleksandrov-topl We log this message of LevelDB-Java is used.  It doesn't say it'll be slow, but it at least says it's experimental: `13:59:29.594| WARN  Bifrost.Node - Using the pure java LevelDB implementation which is still experimental`

PR #3192, opened 2024-02-08
Added special mode for running nodeApp: pruning data storages by removing all unnecessary data

o-aleksandrov-topl:
## Purpose
Special mode for running nodeApp: pruning data storages by removing all unnecessary data

## Approach
Special node app, launched only if command line argument "pruneDir" is defined and point where pruned data shall be stored.

## Testing
Manual testing

## Tickets
BN-1392

### DIFF
--- a/node/src/main/scala/co/topl/node/PrunedDataStoresApp.scala
+++ b/node/src/main/scala/co/topl/node/PrunedDataStoresApp.scala
@@ -1,0 +1,267 @@
+package co.topl.node
+
+import cats.effect.kernel.Async
+import cats.effect.{IO, Resource}
+import co.topl.config.ApplicationConfig
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import co.topl.algebras.Store
+import co.topl.consensus.models._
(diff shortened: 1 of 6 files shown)


node/src/main/scala/co/topl/node/PrunedDataStoresApp.scala
```diff
@@ -0,0 +1,267 @@
...
+        ().pure[F]
+    } >>
+    (
+      saveData(dataStores.parentChildTree, prunedDataStores.parentChildTree)(blockId),
+      saveSlotData(prunedDataStores.slotData, slotData),
+      saveData(dataStores.headers, prunedDataStores.headers)(blockId),
+      saveData(dataStores.bodies, prunedDataStores.bodies)(blockId),
+      saveHeight(prunedDataStores.blockHeightTreeLocal)(slotData),
+      saveHeight(prunedDataStores.blockHeightTreeP2P)(slotData)
+    ).tupled.void >>
```
SeanCheatham, line 134: Small optimization: `.parTupled` would allow saving these in parallel.

SeanCheatham (approved):

PR #3196, opened 2024-02-13
BN-1408 Do not process SlotData from remote peer if previous slot data is not processed yet

o-aleksandrov-topl:
## Purpose
During sync from scratch, there is no reason to do any actions on the recently adopted block from the remote peer, because we usually have not yet fully processed the previous batch of good SlotData. So we could save CPU time by skipping unnecessary chain comparisons

## Approach
remoteSlotShallBeProcessed only if `lastProcessedHeight >= remoteHeight || currentHeight >= lastProcessedHeight` 

## Testing
Unit test + integration test

## Tickets
BN-1408

### DIFF
--- a/networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockHeaderFetcher.scala
+++ b/networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockHeaderFetcher.scala
@@ -122,28 +122,55 @@ object PeerBlockHeaderFetcher {
   private def slotDataFetcher[F[_]: Async: Logger](
     state:             State[F],
     newBlockIdsStream: Stream[F, BlockId]
-  ): Stream[F, Unit] =
+  ): Stream[F, Option[Long]] =
     // TODO close connection to remote peer in case of error
     newBlockIdsStream
-      .evalMap { newBlockId =>
-        processBlockId(state, newBlockId)
+      .evalScan(Option.empty[Long]) { (lastProcessedHeight, newBlockId) =>
+        processRemoteBlockId(state, newBlockId, lastProcessedHeight)
           .handleErrorWith(
-            Logger[F].error(_)(
-              show"Fetching slot data for $newBlockId from remote host ${state.hostIdString} return error"
-            ) >>
-            state.peersManager.sendNoWait(PeersManager.Message.NonCriticalErrorForHost(state.hostId))
+            Logger[F].error(_)(show"Fetching slot data $newBlockId from remote ${state.hostIdString} returns error") >>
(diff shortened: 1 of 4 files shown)


networking/src/main/scala/co/topl/networking/fsnetwork/PeerBlockHeaderFetcher.scala
```diff
@@ -122,28 +122,55 @@ object PeerBlockHeaderFetcher {
...
+      )
+    } yield res
+
+  private def remoteSlotShallBeProcessed[F[_]: Async](
+    remoteHeight:           Long,
+    currentHeight:          Long,
+    lastProcessedHeightOpt: Option[Long]
+  ): F[Boolean] = {
+    val lastProcessedHeight = lastProcessedHeightOpt.getOrElse(Long.MaxValue)
+    (lastProcessedHeight >= remoteHeight || currentHeight >= lastProcessedHeight).pure[F]
```
SeanCheatham, line 161: What happens in a situation where the remote peer has a better chain using density (but has a lower overall height)?

o-aleksandrov-topl: @SeanCheatham then `lastProcessedHeight > remoteHeight`  and we process new remote slot data.

o-aleksandrov-topl: @SeanCheatham There is one downside for that implementation: because we do not switch to other forks until we process current one then it is possible to download and process non-actual fork. For example we receivied chain a -> b -> c -> d -> f and start to process it. But remote peer update that chain to a -> b1 -> c1 -> d1 -> f1, then we download a , b, c, d, f and b1, c1, d1, f1. But it is a really rare event that could occur only when we almost synced

SeanCheatham: In that circumstance, wouldn't the node process `a, b, c, d, and f`, but then ignore `b1, c1, d1, and f1` because b1 has a lower height than f?

o-aleksandrov-topl: as we process and accept f, then `currentHeight >= lastProcessedHeight` and we will start processing f1 fork

o-aleksandrov-topl: as soon as we receive b1 (if we process b+ block) we will process new slot data

SeanCheatham (approved):

PR #3204, opened 2024-03-06
Standalone Genus

SeanCheatham:
## Purpose
- Genus currently runs in the same JVM as the rest of the node, which may lead to excess pressure on critical consensus mechanisms
## Approach
- Create a "standalone" version of Genus
- Rename `genusLibrary` to `genus`
- Move `genus` code out of the Node module
- Launch Genus gRPC on the same port that is currently expected by clients
- Proxy Node gRPC traffic through Genus gRPC
## Testing
- Node still seems to behave as it did previously
## Tickets
- #BN-1430


admiraladmirable (approved):
Thanks for porting the health check! 🎉 (I'm just reviewing to get more exposure to the code base)

PR #3224, opened 2024-04-04
BN-1434 Added storage for indexing transaction id to block id

o-aleksandrov-topl:
## Purpose
For age based filter information about boxId creation height is required. To get that information we need to have TransactionId -> BlockId information.

## Approach
Added updated storage which stores information about transactionId -> BlockId

## Testing
Unit tests

## Tickets
part of BN-1434


common-interpreters/src/main/scala/co/topl/interpreters/TxIdToBlockIdTree.scala
```diff
@@ -0,0 +1,34 @@
...
+import cats.implicits._
+import co.topl.typeclasses.implicits._
+import org.typelevel.log4cats.Logger
+
+object TxIdToBlockIdTree {
+  type State[F[_]] = Store[F, TransactionId, BlockId]
+
+  def make[F[_]: Async: Logger](
+    currentBlockId: F[BlockId],
+    blockIdToBody:  Store[F, BlockId, BlockBody],
```
SeanCheatham, line 18: I suggest using a function `BlockId => F[BlockBody]` instead of a `Store`

common-interpreters/src/main/scala/co/topl/interpreters/TxIdToBlockIdTree.scala
```diff
@@ -0,0 +1,34 @@
...
+  ): F[EventSourcedState[F, State[F], BlockId]] =
+    EventSourcedState.OfTree.make[F, State[F], BlockId](
+      initialState = initialState.pure[F],
+      initialEventId = currentBlockId,
+      applyEvent = (storage: State[F], id: BlockId) =>
+        blockIdToBody.get(id).flatMap {
+          case Some(body) => body.transactionIds.traverse(storage.put(_, id)) >> storage.pure[F]
+          case None       => storage.pure[F]
+        },
+      unapplyEvent = (storage: State[F], _: BlockId) => storage.pure[F],
```
SeanCheatham, line 30: In the unapply, I think you need to fetch the `BlockBody` and then remove all corresponding `transactionIds` from the storage/state.

common-interpreters/src/main/scala/co/topl/interpreters/TxIdToBlockIdTree.scala
o-aleksandrov-topl: done

common-interpreters/src/main/scala/co/topl/interpreters/TxIdToBlockIdTree.scala
o-aleksandrov-topl: done

SeanCheatham (approved):

PR #3228, opened 2024-04-16
Typed Protocol Replacement

SeanCheatham:
## Purpose
- The TypedProtocol approach is suitable for network communications with complex protocols
- Bifrost's networking uses relatively simple request/response protocols, and in cases where it doesn't, they could be adapted to request/response protocols
## Approach
- Migrate all TypedProtocol and most multiplexer code into a `co.topl.networking.legacy` package
- Added simpler Multiplexer implementation
- Added new BlockchainSocketHandler implementation
- Determine when to use legacy protocol by hooking into the PeerIdentity handshake
## Testing
- Unit tests pass, NodeAppTest passes
- TODO: Test against devnet
## Tickets
- #BN-1439


scasplte2 (commented):
This is a pretty interesting update. It seems much easier to understand and potentially more performant given the reader / writer structure using more effectful structures (Streams, Deferred, etc.). The only thought I had was to potentially choose a more extensible package naming scheme in case there are further updates in the future (which seems plausible) so as to avoid `network.legacy` and `network.legacy.legacy`

networking/src/main/scala/co/topl/networking/blockchain/BlockchainSocketHandler.scala
```diff
@@ -1,216 +1,377 @@
...
+          cache.remoteBlockAdoptions
+            .tryOffer(_)
+            .flatMap(
+              Async[F].raiseUnless(_)(new IllegalStateException("Slow local subscriber of peer block adoptions"))
+            )
+        )
+        .void,
+      Stream
+        .repeatEval(
+          writeRequestNoTimeout(
```
o-aleksandrov-topl, line 179: 1. We use the same mutex during requests, how we could get few requests in the same time?
2. Will that stream normally finished if request is not fulfilled but background stream try to finish it?

networking/src/main/scala/co/topl/networking/blockchain/BlockchainMultiplexerId.scala
```diff
@@ -0,0 +1,42 @@
...
+ * An ADT of blockchain-specific multiplexer ports
+ * @param id the underlying integer value for the multiplexer port
+ */
+sealed abstract class BlockchainMultiplexerId(val id: Int)
+
+object BlockchainMultiplexerId {
+
+  def parse(id: Int): Option[BlockchainMultiplexerId] =
+    id match {
+      case 10 => BlockIdAtHeightRequest.some
```
o-aleksandrov-topl, line 15: any reason to start id from 10?

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraTransmittableCodecs.scala
```diff
@@ -42,6 +41,29 @@ trait TetraTransmittableCodecs {
...
+          .flatMap(bytes =>
+            bytes.byteAt(0) match {
+              case 0 => none[T].asRight[String]
+              case 1 => Transmittable[T].fromTransmittableBytes(bytes.substring(1)).map(_.some)
+              case _ => "Invalid Optional".asLeft[Option[T]]
+            }
+          )
+    }
+
+  val ZeroBS: ByteString = ByteString.copyFrom(Array[Byte](0))
```
o-aleksandrov-topl, line 65: rename to `request / response marker`?

networking/src/main/scala/co/topl/networking/blockchain/BlockchainSocketHandler.scala
```diff
@@ -1,216 +1,377 @@
...
-            peerServerHandlers ++
-            appLevelNotifyHandlers
-      } yield subHandlers -> blockchainProtocolClient
-  }
+object PeerStreamBuffer {
 
+  def make[F[_]: Async]: Resource[F, PeerStreamBuffer[F]] =
+    (
+      Queue.bounded[F, BlockId](256),
+      Queue.bounded[F, TransactionId](256),
```
o-aleksandrov-topl, line 371: maybe increase to 512? We could reach a limit of 256 in case if we receive a lot of transactions from some remote peer. It doesn't solve a problem but could postpone it.

networking/src/main/scala/co/topl/networking/p2p/FS2P2PServer.scala
```diff
@@ -95,6 +102,7 @@ object FS2P2PServer {
                     .void
                     .toResource
                 }
+              _ <- (socket.endOfOutput *> socket.endOfInput).toResource
```
o-aleksandrov-topl, line 105: good addition!

networking/src/main/scala/co/topl/networking/p2p/FS2P2PServer.scala
```diff
@@ -164,6 +172,7 @@ object FS2P2PServer {
                     .void
                     .toResource
                 }
+              _ <- socket.endOfOutput.toResource
```
o-aleksandrov-topl, line 175: why not add `endOfInput` as well?

networking/src/main/scala/co/topl/networking/multiplexer/MultiplexedReaderWriter.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  def writer[F[_]: Async](socket: Socket[F]): (Int, Chunk[Byte]) => F[Unit] =
+    (port, data) =>
+      socket
+        .write(
+          Chunk.array(Ints.toByteArray(port)) ++
+          Chunk.array(Ints.toByteArray(data.size)) ++
+          data
+        )
+        .timeout(3.seconds)
```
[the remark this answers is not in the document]
SeanCheatham: Updated to use a `defaultTimeout` field in NetworkProperties, and use that value in the new network code

networking/src/main/scala/co/topl/networking/blockchain/BlockchainSocketHandler.scala
```diff
@@ -1,216 +1,377 @@
...
+    ).parJoinUnbounded
 
-    val knownHostsRecipF =
-      TypedProtocolSetFactory.CommonProtocols
-        .requestResponseReciprocated[F, CurrentKnownHostsReq, CurrentKnownHostsRes](
-          BlockchainProtocols.KnownHosts,
-          protocolServer.getKnownHosts,
-          17: Byte,
-          18: Byte
+  private def cacheStreams =
```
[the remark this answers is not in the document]
SeanCheatham: Updated to `streamBufferPushers`

networking/src/main/scala/co/topl/networking/blockchain/BlockchainMultiplexerId.scala
SeanCheatham: Only reason I had was to leave room for "imporant" messages later on if needed

tetra-byte-codecs/src/main/scala/co/topl/codecs/bytes/tetra/TetraTransmittableCodecs.scala
SeanCheatham: I kept the names as-is, but added some aliases `RequestMarker` and `ResponseMarker`

networking/src/main/scala/co/topl/networking/blockchain/BlockchainSocketHandler.scala
SeanCheatham: Updated to 512.

Also, just a note: There's a behavior change in the new code.  Instead of dropping notifications for slow subscribers, it'll instead raise an error if these buffers try to overflow.

networking/src/main/scala/co/topl/networking/p2p/FS2P2PServer.scala
SeanCheatham: Updated to include this.  I omitted it previously because I thought it was causing an issue (turned out to be something else though)

SeanCheatham:
@scasplte2 
> This is a pretty interesting update. It seems much easier to understand and potentially more performant given the reader / writer structure using more effectful structures (Streams, Deferred, etc.). The only thought I had was to potentially choose a more extensible package naming scheme in case there are further updates in the future (which seems plausible) so as to avoid `network.legacy` and `network.legacy.legacy`

I think that's a fair point, and there will most certainly be updates to the networking code in the future.

It'll be hard to know how things will change in the future (and which code is applicable to which version of networking).  With that in mind, I think a fair compromise is to have the "current" or "latest" networking code take natural package naming i.e. `co.topl.networking.multiplexer`, but legacy code can live in a versioned namespace.  I opted to go with `co.topl.networking.legacy.v0` for all of the code that we're deprecating.  We intend to remove the `v0` code eventually though anyway.

networking/src/main/scala/co/topl/networking/blockchain/BlockchainSocketHandler.scala
SeanCheatham: 1. I'm not quite sure what you mean here.  There is a single mutex for the whole socket rather than individual mutexes for each multiplexer port.  But you're right that in these cases here, there isn't an expectation of parallel requests because we're converting the port into a stream.
2. These inner-streams will run until the outer stream is canceled.  The outer stream is canceled when the rest of the background stream is canceled, due to the user of `mergeHaltBoth` and/or `concurrently` depending on the circumstance.

PR #3229, opened 2024-04-22
[BN-1409]: Implement Metrics via Kamon

admiraladmirable:
## Purpose
Enable Kamon based metrics and export in Prometheus format to enable Grafana integration.

Currently, we have all of our dashboards in GCP using Log Based Metrics. While this works for getting high level information, we are a bit limited since there is overhead in parsing, storing, and correlating metrics this way. Also, we are not creating metrics based on events, we are using log messages to create the metrics, which in some cases leads to non-instantaneous observability. 

## Approach
* Use existing Kamon integration, and add the Prometheus exporter library.
* Refactor the `Stats` trait and implement a `KamonStatsInterpreter`.
* Add first pass of metrics we'd like to track, suck as:
  * Peer count and state
  * Node adoptions
  * Blocks minted
  * Header/block download time.

This approach allows us to initialize the implicit `Stats` implementation, and then enable stats for our components by simply adding the new implicit and calling `Stats[F].recordGauge(...)`.

For Kamon, since the meters (gauge, histogram, counter, etc) need to be created and stored, we are doing so by using a ref map, which when invoked in other parts of the code will either create the meter if it doesn't exist, or look it up by key and return it.

## Testing
* Ran locally, observed metrics generated in Prometheus/Grafana.
* `sbt preparePR`.

## Tickets
*BN-1409


common-interpreters/src/main/scala/co/topl/interpreters/KamonStatsInterpreter.scala
```diff
@@ -0,0 +1,119 @@
...
+            _ <- Sync[F].delay(attributes.map { case (k, v) =>
+              tagBuilder.add(k, v.noSpaces.stripPrefix("\"").stripSuffix("\""))
+            })
+            tags <- Sync[F].delay(tagBuilder.build())
+          } yield (tags)
+
+          _ <- Sync[F].delay(counter.asInstanceOf[Metric.Counter].withTags(metricTags).increment())
+        } yield ()
+
+      def decrementCounter(statName: String, description: String, attributes: Map[String, Json]): F[Unit] = ???
```
SeanCheatham, line 63: I suggest implementing this _somehow_, either by raising an exception `Sync[F].raiseError(...)` or by no-op `Sync[F].unit`

consensus/src/main/scala/co/topl/consensus/interpreters/LocalChain.scala
```diff
@@ -40,6 +42,14 @@ object LocalChain {
           Sync[F].uncancelable(_ =>
             onAdopted(slotData.slotId.blockId) >>
             headRef.set(slotData) >>
+            Async[F].defer(
```
SeanCheatham, line 45: You can most likely omit the `Async[F].defer`.  If you find that it's behaving weirdly without the call, you could just move it into the implementation in `Stats[F]#recordGauge` instead

SeanCheatham (approved):

PR #3234, opened 2024-05-01
BN-1469 Code cleaning

o-aleksandrov-topl:
## Purpose
Cleaning code

## Testing
Unit tests
## Tickets
*

**Delete everything after and including this line after you are done reading and understand what to put for each section**

>### Purpose
>
>*State in plain words what the scope of the pull request and what problem it is trying to solve*
>
>### Approach
>
>*Enumerate the steps you took to create the code/changes contained in this pull request*
>
>### Testing
>
>*Specify all unit, integration, or end-to-end testing that is contained in this pull request*
>
>### Tickets
>
>*List tickets if any that are connected to this pull request in the following format:*
>
>_* closes #1234_


genus/src/main/scala/co/topl/genus/orientDb/instances/SchemaBlockHeader.scala
```diff
@@ -63,7 +63,7 @@ object SchemaBlockHeader {
       // @formatter:on
     v =>
       BlockHeader(
-        Some(BlockId(ByteString.copyFrom(v(Field.BlockId): Array[Byte]))),
+        Option(BlockId(ByteString.copyFrom(v(Field.BlockId): Array[Byte]))),
```
SeanCheatham, line 66: Any particular reason for this change?  (I think it'll just add an extra null check on something that won't be null)

SeanCheatham (approved):

genus/src/main/scala/co/topl/genus/orientDb/instances/SchemaBlockHeader.scala
o-aleksandrov-topl: yes just for null check and for removing warning in IDEA

PR #3236, opened 2024-05-10
BN-1331 Added option to not expose remote peers to other nodes by id / ip(s)

o-aleksandrov-topl:
## Purpose
BN-1331 Added option to not expose remote peers to other nodes by id / ip(s)

## Approach
Added special class which filter out remote peers before updating the current hot peers list

## Testing
Unit tests

## Tickets
BN-1331


SeanCheatham (approved):

admiraladmirable (approved):
LGTM. Just so I understand, this is meant to be used for public relay nodes so they don't try to broadcast unreachable peers correct?

PR #3244, opened 2024-06-11
BN-1505 Stop trust of SlotData from remote peer P1

o-aleksandrov-topl:
## Purpose
Stop build parent-child tree from slot data; do not use slot data received from remote peer to calculate ETA

## Approach
Associate block parent-child only after receiving header from a remote peer; update slot data based on actual header rather than use slot data from a remote peer.

## Testing

## Tickets
BN-1505


scasplte2 (commented):
nice :fire:

SeanCheatham (approved):

PR #3248, opened 2024-07-11
Block Packer Completion Case

SeanCheatham:
## Purpose
- If the BlockPacker is able to complete processing for a non-empty mempool, it will complete the stream
- If the stream completes, the BlockProducer will immediately make the block, skipping the regtest permit and the slot delay
## Approach
- In BlockProducer, add `Stream.never` to the `BlockPacker#blockImprover` call to force the stream to never complete
## Testing
- Verified locally that broadcasting a TX in regtest mode does not make a new block
## Tickets
- #BN-1540


mundacho (approved):
LGTM!

PR #3249, opened 2024-07-16
BN-1529 Added network control rpc (disabled by default) for controling network.

o-aleksandrov-topl:
## Purpose
For testing purposes, we need to be able to control the network on the node.
Right now, it supports three commands: het host id, forget remote host, add known remote host

## Approach
Add RPC for network controlling

## Testing
Unit tests + integration test

## Tickets
BN-1529


SeanCheatham (approved):
Would it be beneficial to enable this mode automatically when in `regtest` mode? (CC @mundacho )

PR #3250, opened 2024-07-24
BN-1524 Download multiply slot data instead one-by-one if long slot data chain sync is required

o-aleksandrov-topl:
## Purpose
If the node is far behind from remote peer then syncing slot data from the remote peer could take a lot of time because every request will suffer from network latency. To avoid that in some cases we could request many slot data from remote peer at the same time

## Approach
Add `getRemoteSlotDataWithParents(from, to)` request to the server which could be described like "give me as many slot data as possible between `from` and `to`" which could be used for fast building slot data chain. Client will repeat that request until chain `from` to `to` is built.

## Testing
Unit test + integration tests

## Tickets
BN-1524

### DIFF
--- a/networking/src/main/scala/co/topl/networking/legacy/v0/LegacyBlockchainSocketHandler.scala
+++ b/networking/src/main/scala/co/topl/networking/legacy/v0/LegacyBlockchainSocketHandler.scala
@@ -196,7 +196,7 @@ object LegacyBlockchainSocketHandler {
 
           def closeConnection(): F[Unit] = close
 
-          def getRemoteSlotDataWithParents(to: BlockId, from: BlockId): F[Option[List[SlotData]]] =
+          def getRemoteSlotDataWithParents(from: BlockId, to: BlockId): F[Option[List[SlotData]]] =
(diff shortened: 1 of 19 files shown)


scasplte2:
stoked for this 🙌

networking/src/main/scala/co/topl/networking/legacy/v0/LegacyBlockchainSocketHandler.scala
```diff
@@ -195,6 +195,9 @@ object LegacyBlockchainSocketHandler {
             appNotifyCallback(networkLevel).void
 
           def closeConnection(): F[Unit] = close
+
+          def getRemoteSlotDataWithParents(to: BlockId, from: BlockId): F[Option[List[SlotData]]] =
```
admiraladmirable, line 199: Should this be?
```suggestion
          def getRemoteSlotDataWithParents(from: BlockId, to: BlockId): F[Option[List[SlotData]]] =
```

admiraladmirable (approved):

SeanCheatham (approved):

PR #3253, opened 2024-08-01
Socket "readExactly" Helper

SeanCheatham:
## Purpose
- Socket.read may not produce all of the requested bytes
- The consumer assumes all bytes are available, so parsing is likely to fail
- This situation is only likely to happen when crossing physical network boundaries (i.e. local to GCP), but not in local connections
## Approach
- Add a helper which polls the socket (similar to the `readN` implementation but returning an Option instead)
## Testing
- `preparePR`
- This isn't something that is easily reproducible locally
## Tickets
- #BN-1543

### DIFF
--- a/networking/src/main/scala/co/topl/networking/package.scala
+++ b/networking/src/main/scala/co/topl/networking/package.scala
@@ -1,9 +1,14 @@
 package co.topl
 
+import cats.Monad
+import cats.data.OptionT
+import cats.implicits._
 import co.topl.models.p2p._
 import co.topl.networking.fsnetwork.RemotePeer
 import co.topl.node.models.KnownHost
 import com.comcast.ip4s.{IpAddress, SocketAddress}
+import fs2.Chunk
+import fs2.io.net.Socket
 
 import java.net.InetAddress
 import java.nio.ByteBuffer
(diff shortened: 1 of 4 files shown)


networking/src/main/scala/co/topl/networking/package.scala
```diff
@@ -34,4 +39,23 @@ package object networking {
...
+
+  implicit class SocketOps[F[_]](val socket: Socket[F]) extends AnyVal {
+
+    /**
+     * Calling Socket.read(numberOfBytes) may not return all `numberOfBytes`. This situation needs to be checked,
+     * and additional bytes should be requested until the expected number of bytes are received.
+     * @param length the total number of bytes desired
+     * @return a chunk of bytes with the desired length, or None if the socket is closed
+     */
+    def readExactly(length: Int)(implicit monadF: Monad[F]): F[Option[Chunk[Byte]]] =
```
scasplte2, line 51: two quick thoughts:
1. This is probably the case but are sockets "dedicated" in a sense? Just thinking if there is any concern about concurrent processes mixing bytes and invalidating the higher-order message.
2. Could this polling method result in excessive querying of the socket in a way that exacerbates the problem? Perhaps if the read operation is blocking to further writes, then a short delay may be preferable to allow time to ingest further bytes.

Excuse the ignorance if these concerns are unfounded.

SeanCheatham: I think we're probably okay on these notes.

1. A socket is "dedicated" per-peer. While there is some concern about reading and writing from multiple fibers, we should be protected. The "read" side is funneled into a single-consumer stream, so we wouldn't have concurrent reads anyway. The "write" side is protected by a Mutex such that one fiber/thread must finish writing its message before a second fiber can start.
2. Since we use the single-consumer stream, we don't need to worry about excessive querying of the socket. But in either case, the internal implementation of FS2's Socket already adds a Mutex on the read-side anyway. (There's also a separate Mutex on the write-side, internal to FS2's implementation)

o-aleksandrov-topl, line 51: probably better name like `readAtLeast`

o-aleksandrov-topl (approved):

networking/src/main/scala/co/topl/networking/package.scala
SeanCheatham: I think `readExactly` is more accurate in this case since it'll either return Some(exact number of bytes requested) or None. `readAtLeast` implies this method could return more than the requested number of bytes which is not the expected behavior.

PR #3254, opened 2024-08-12
BN-1527 Handling proposals for versioning

o-aleksandrov-topl:
## Purpose
Implementation of soft / hard fork managing. Handling proposal by storing them and convert them to new version if there is enough voting

## Approach
Adding new EventSourcing for versioning

## Testing
Unit tests

## Tickets
BN-1527


consensus/src/main/scala/co/topl/consensus/interpreters/VersionsEventSourceState.scala
```diff
@@ -0,0 +1,364 @@
...
+import co.topl.typeclasses.implicits._
+import org.typelevel.log4cats.Logger
+import co.topl.consensus.models.BlockHeader
+import co.topl.proto.node.EpochData
+import co.topl.algebras.ClockAlgebra.implicits._
+import co.topl.crypto.hash.Blake2b256
+import java.nio.ByteBuffer
+
+case class ProposalConfig(
+  proposalVotingMaxWindow: Int = 5,
```
SeanCheatham, line 23: Could you clarify the meaning of these two parameters `proposalVotingMaxWindow` and `proposalVotingWindow`?

consensus/src/main/scala/co/topl/consensus/interpreters/VersionsEventSourceState.scala
```diff
@@ -0,0 +1,364 @@
...
+      state:         VersionsData[F],
+      previousEpoch: Epoch,
+      currentEpoch:  Epoch,
+      proposalIds:   Set[ProposalId]
+    ): F[Unit] =
+      for {
+        votingMaxWindow <- getVotingRangeForMaxWindow(previousEpoch).pure[F]
+        blocksInEpochs  <- getEpochsBlockCounts(state, votingMaxWindow)
+        // We MUST sort proposals for to have the same proposal processing orders
+        proposalResults <- proposalIds.toList.sorted.traverse(getProposalResult(_, state, blocksInEpochs))
```
SeanCheatham, line 148: Would it be beneficial to use a `ListSet` for the `proposalIds` type? It would preserve the order in which things were proposed. I'm not sure if that actually matters for the purposes of applying votes, or if the only thing that matters is that all nodes do things in the same order.

(Side-question, where is the "Ordering" for a `ProposalId` defined? Having a hard time finding it in the PR)

SeanCheatham (approved):

consensus/src/main/scala/co/topl/consensus/interpreters/VersionsEventSourceState.scala
o-aleksandrov-topl: Proposal voting max window – How many epochs stakers could vote for that update proposal. If proposal is not approved during that window it became inactive. 

Proposal voting window – Number of epoch required for proposal voting to be approved. i.e. how many epoch with required “Update proposal approving percentage” shall be in chain. For example if Proposal voting max window is set to 5 and Proposal voting window is set to 2, and proposal is done on epoch 10 then during epoch 11,12,13,14,15 we need to have two sequential epoch (for example 14 and 15) where stakers approved that proposal.  Proposal voting window shall be equal or less than Proposal voting max window.

consensus/src/main/scala/co/topl/consensus/interpreters/VersionsEventSourceState.scala
o-aleksandrov-topl: We just need the same processing orders among all nodes to have the same versions if two or more proposals are passed. 

`ProposalId` is defined as `  type ProposalId = Int ` so no "Ordering" here

PR #3255, opened 2024-08-22
BN-1556 Implement Version Voting and version defining

o-aleksandrov-topl:
## Purpose
Add handling of version voting and version switching for hard/soft fork

## Testing
Unit tests

## Tickets
BN-1556


algebras/src/main/scala/co/topl/algebras/Store.scala
```diff
@@ -7,17 +7,22 @@ import cats.data.OptionT
 trait StoreReader[F[_], Key, T] {
   outer =>
   def get(id: Key): F[Option[T]]
+  def getAll(): F[Seq[(Key, T)]]
```
SeanCheatham, line 10: It worries me a little bit to have this exposed as a public method and return a Seq of items. If someone accidentally calls `.getAll()` on the wrong Store instance, the JVM would likely run out of memory.

A couple alternatives, neither are particularly great:
1. Have this return a `Stream[F, (Key, T)]` and use FS2 helpers to read from a blocking iterator
2. Use a separate `trait FullStoreReader[F[_], Key, T]`. `LevelDbStore` can still extend all of the traits, but consumers would need to intentionally require a `FullStoreReader` in the paramter if they want such functionality.

SeanCheatham (approved):

algebras/src/main/scala/co/topl/algebras/Store.scala
o-aleksandrov-topl: I think it is ok, since we do not provide any kind of SDK. Otherwise, we could just make function `forEach` where store user could manually save every element to an externally created collection. But in the end we anyway will create `getAll` function with the same functionality.

PR #3263, opened 2024-09-21
BN-1561 Add appropriate header version in minted block

o-aleksandrov-topl:
## Purpose
Minted headers shall contain correct version

## Approach
Add appropriate dynamic version to the header

## Testing
Unit tests

## Tickets
BN-1561


minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
```diff
@@ -133,16 +138,25 @@ object BlockProducer {
...
         blockMaker = prepareUnsignedBlock(parentSlotData, fullBody, timestamp, nextHit)
         eta: Eta = Sized.strictUnsafe[ByteString, Eta.Length](nextHit.cert.eta)
-        _           <- Logger[F].info("Certifying block")
-        maybeHeader <- staker.certifyBlock(parentSlotData.slotId, nextHit.slot, blockMaker, eta)
+        _ <- Logger[F].info("Certifying block")
+        maybeHeader <-
+          staker
+            .certifyBlock(parentSlotData.slotId, nextHit.slot, blockMaker, eta)
+            .map(
+              _.map(_.copy(version = protocolVersion))
```
SeanCheatham, line 158: I think it may be better to update the `co.topl.models.UnsignedBlockHeader` case class to include the `version` field (since it should have been included from the beginning like you observed), and then change the `prepareUnsignedBlock` method in this class.

We can still maintain backwards compatibility by not including the field in the signable bytes for now.

SeanCheatham (approved):

minting/src/main/scala/co/topl/minting/interpreters/BlockProducer.scala
o-aleksandrov-topl: yeap I initially think about it, but there is an issue with codec for UnsignedBlockHeader. So I prefer to use that workaround and do correctly later.

